Issue #38, opened 2025-12-26
System Information: Hard drive info not detected due to volume mount filename mismatch


chriscrosstalk:
## Summary

The System Information page shows "No storage devices detected" because there's a filename mismatch between the Docker Compose volume mount and the path the application expects.

## Root Cause

Three files are involved:

1. **`install/collect_disk_info.sh`** - Writes disk info to `/tmp/nomad-disk-info.json`
2. **`install/management_compose.yaml`** - Mounts that file into the container as `/app/storage/disk-info.json`
3. **`admin/app/services/system_service.ts`** (line 15) - Expects the file at `/storage/nomad-disk-info.json`

The mismatch:
- Mount destination: `/app/storage/disk-info.json`
- Code expects: `/app/storage/nomad-disk-info.json`

## Fix

In `install/management_compose.yaml`, change line 10:

```yaml
# Before
- /tmp/nomad-disk-info.json:/app/storage/disk-info.json

# After
- /tmp/nomad-disk-info.json:/app/storage/nomad-disk-info.json
```

## Impact

- Storage Devices section shows "No storage devices detected"
- System Status shows "Storage Devices: 0"
- Users cannot see disk usage information

## Additional Notes

The background script (`collect_disk_info.sh`) only runs if it was started during installation. After fixing the mount path, users may also need to verify the background script is running on their host.

Issue #37, opened 2025-12-26
System Info: Memory usage calculation shows inflated values due to Linux buffer/cache


chriscrosstalk:
## Problem

The System Information page shows very high memory usage (e.g., 97%) even when the system is not under actual memory pressure. This can cause unnecessary alarm for users.

## Root Cause

The current implementation in `admin/inertia/pages/settings/system.tsx` (lines 25-26) calculates memory usage using `mem.used`:

```typescript
const memoryUsagePercent = info?.mem.total
  ? ((info.mem.used / info.mem.total) * 100).toFixed(1)
  : 0
```

On Linux, `used` memory includes buffers and cache, which the kernel will release when applications need it. This is normal Linux behavior - it uses free RAM as disk cache to improve performance.

**Key indicator the system is fine:** Swap usage shows 0% - if the system were truly under memory pressure, it would be swapping.

## Proposed Solution

Use `mem.available` from the `systeminformation` library instead, which represents how much memory is actually available for new applications (accounting for reclaimable cache):

```typescript
const memoryUsagePercent = info?.mem.total
  ? (((info.mem.total - info.mem.available) / info.mem.total) * 100).toFixed(1)
  : 0
```

## Additional Considerations

- Update the "RAM in Use" label in the Memory Allocation section to reflect what's actually being measured
- Consider showing both "Active" and "Available" memory for more transparency
- The 90% warning threshold may need adjustment after this change

## Verification

Users can verify actual memory state by running `free -h` on the host system and checking the `available` column.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

chriscrosstalk:
<img width="1920" height="959" alt="Image" src="https://github.com/user-attachments/assets/8decfacd-0fbd-4c5f-a992-e0bf5e13cd9d" />

Issue #36, opened 2025-12-24
Download Base Assets fails - incorrect GitHub raw URL construction


chriscrosstalk:
## Summary

Clicking **Download Base Assets** on the Maps Manager page fails with the error: "An error occurred while downloading the base map assets. Please try again."

## Steps to Reproduce

1. Navigate to **Dashboard → Settings → Maps Manager**
2. Click the **Download Base Assets** button
3. Observe the error toast in the bottom-right corner

## Screenshot

![Maps Manager Error](https://github.com/Crosstalk-Solutions/project-nomad/blob/master/assets/maps-manager-error1.png?raw=true)

## Root Cause

In `admin/app/services/map_service.ts` lines 67-72, the default URL for downloading base assets is constructed incorrectly:

```typescript
const defaultTarFileURL = new URL(
  this.baseAssetsTarFile,
  'https://github.com/Crosstalk-Solutions/project-nomad-maps/blob/master'
)
defaultTarFileURL.searchParams.append('raw', 'true')
```

This produces a URL like:
```
https://github.com/Crosstalk-Solutions/project-nomad-maps/blob/master/base-assets.tar.gz?raw=true
```

The `/blob/` path with `?raw=true` query parameter doesn't reliably serve raw files from GitHub. The correct approach is to use the `/raw/` path:

```
https://github.com/Crosstalk-Solutions/project-nomad-maps/raw/refs/heads/master/base-assets.tar.gz
```

## Suggested Fix

Change the URL construction to use the raw URL format:

```typescript
const defaultTarFileURL = new URL(
  this.baseAssetsTarFile,
  'https://github.com/Crosstalk-Solutions/project-nomad-maps/raw/refs/heads/master/'
)
// Remove the searchParams.append line - no longer needed
```

## Environment

- **Version:** v1.12.0
- **Location:** `admin/app/services/map_service.ts:67-72`

Issue #35, opened 2025-12-24
Missing favicon across the application


chriscrosstalk:
## Summary

The browser tab has no favicon when browsing the Project N.O.M.A.D. Command Center. This affects all pages.

## Expected Behavior

A favicon should appear in:
- Browser tabs
- Bookmarks
- Browser history

## Current Behavior

No favicon is displayed - browser shows default blank/globe icon.

## Design Suggestion

Use the NOMAD hexagon shape with the existing color scheme (olive/army green with gold accents), featuring a simple **"N"** inside. This would:
- Be recognizable at small sizes (16x16, 32x32)
- Match the existing branding
- Be distinct from the full logo while staying on-brand

## Implementation

Add favicon files to the public/static directory and reference in the HTML head:

```html
<link rel="icon" type="image/png" sizes="32x32" href="/favicon-32x32.png">
<link rel="icon" type="image/png" sizes="16x16" href="/favicon-16x16.png">
<link rel="apple-touch-icon" sizes="180x180" href="/apple-touch-icon.png">
```

## Environment

- Project N.O.M.A.D. Command Center v1.12.0

chriscrosstalk:
<img width="180" height="180" alt="Image" src="https://github.com/user-attachments/assets/95126600-005c-4e26-b775-40257560df1b" />
<img width="16" height="16" alt="Image" src="https://github.com/user-attachments/assets/821f254f-a08a-4e06-8050-d18d2e37dfc3" />
<img width="32" height="32" alt="Image" src="https://github.com/user-attachments/assets/9b2d708e-6574-4dff-a9a6-ada68add2255" />

<img width="192" height="192" alt="Image" src="https://github.com/user-attachments/assets/a593f1f7-7657-4503-a03b-8de194113912" />
<img width="512" height="512" alt="Image" src="https://github.com/user-attachments/assets/d2e0ab58-d642-4974-af4b-e12fbf4a15cd" />

chriscrosstalk:
added numerous favicon formats

PR #44, opened 2026-01-11
Dev

### LINKED ISSUE #39: fix(ZIM): Remote Explorer download button does nothing - missing API client method
## Bug Description

When clicking the Download button in Settings → ZIM Remote Explorer, the confirmation modal appears, but after clicking "Download" nothing happens. No download starts and nothing appears under "Active Downloads".

## Root Cause

The frontend component calls `api.downloadRemoteZimFile()` which **does not exist** in the API client class.

**Location of the call:** `admin/inertia/pages/settings/zim/remote-explorer.tsx:164`
```typescript
async function downloadFile(record: RemoteZimFileEntry) {
  try {
    await api.downloadRemoteZimFile(record.download_url)  // ← Method doesn't exist!
    invalidateDownloads()
  } catch (error) {
    console.error('Error downloading file:', error)
  }
}
```

**Missing from:** `admin/inertia/lib/api.ts`

The backend route `/api/zim/download-remote` exists and works. The equivalent map download method `downloadRemoteMapRegion()` exists, but the ZIM version was never added.

## Fix

Add the missing method to `admin/inertia/lib/api.ts`:

```typescript
async downloadRemoteZimFile(url: string) {
  return catchInternal(async () => {
    const response = await this.client.post<{
      message: string
      filename: string
      jobId: string
      url: string
    }>('/zim/download-remote', { url })
    return response.data
  })()
}
```

## Impact

- All individual ZIM file downloads from the Remote Explorer are broken
- Collection downloads still work (they use a different method)
- The error is silently caught and logged to console, so users see no feedback

jakeaturner:
Closes #39 #38 #37 #36 #35 #34 #33 #32 #31 #30 #29

### DIFF
--- a/admin/app/services/system_service.ts
+++ b/admin/app/services/system_service.ts
@@ -58,6 +58,8 @@ export class SystemService {
   }
 
   async getServices({ installedOnly = true }: { installedOnly?: boolean }): Promise<ServiceSlim[]> {
+    await this._syncContainersWithDatabase() // Sync up before fetching to ensure we have the latest status
+
     const query = Service.query()
       .orderBy('friendly_name', 'asc')
       .select('id', 'service_name', 'installed', 'ui_location', 'friendly_name', 'description')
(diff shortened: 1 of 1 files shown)


jakeaturner:
Closes #39

PR #108, opened 2026-01-28
fix(migrations): Fix timestamp ordering for builder_tag migration

chriscrosstalk:
## Summary
- Rename migration file from `1769324448_add_builder_tag_to_benchmark_results.ts` to `1769324448000_add_builder_tag_to_benchmark_results.ts`

## Problem
Fresh installs fail with:
```
[ error ] alter table `benchmark_results` add `builder_tag` varchar(64) null - Table 'nomad.benchmark_results' doesn't exist
```

The migration file used a **10-digit timestamp** (`1769324448`) while all other migrations use **13-digit timestamps**. When AdonisJS sorts migrations numerically:
- `1769324448` (1.7 billion) < `1769097600001` (1.7 trillion)

This caused the `ALTER TABLE` migration to run **before** the `CREATE TABLE` migration.

## Solution
Append `000` to make it a 13-digit timestamp: `1769324448` → `1769324448000`

Now the sort order is correct:
1. `1769097600001_create_benchmark_results_table` (creates table)
2. `1769324448000_add_builder_tag_to_benchmark_results` (alters table) ✓

## Test plan
- [ ] Fresh install completes without migration errors
- [ ] Existing installs with the column already added continue to work


chriscrosstalk:
Merging without waiting for review - this is a critical bug that breaks all fresh installs. The fix is a simple file rename (10-digit timestamp → 13-digit). Verified working on NOMAD3 with manual workaround.

PR #230, opened 2026-03-12
feat: Add Remote ollama

hestela:
Adds ability to set an ollama host URL. Tested using a nomad host running in proxmox LXC and a desktop with a GPU and ollama setup with OLLAMA_HOST=0.0.0.0:11434". Allows for AI assistant to install on a host without a GPU or ollama installed.  

User is responsible for setup of ollama+gpu drivers in other other machine if this is used.

screenshot of new setting:
assisstant-url

### DIFF
--- a/admin/app/controllers/settings_controller.ts
+++ b/admin/app/controllers/settings_controller.ts
@@ -52,7 +52,7 @@ export default class SettingsController {
 
     async models({ inertia }: HttpContext) {
         const availableModels = await this.ollamaService.getAvailableModels({ sort: 'pulls', recommendedOnly: false, query: null, limit: 15 });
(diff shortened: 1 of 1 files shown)


hestela:
Also, an Idea for more customization would be to use the openai api which ollama does support, then you could run any api host such as LM studio. https://docs.ollama.com/api/openai-compatibility

Otherwise I tested this PR on a couple different ollama hosts I have in my network and it worked fine. Running docker in an lxc host requires a hack to the lxc config file but it otherwise works well as a quick testbed.

hestela:
rebased and added (ai-chat) scope. opened issue since the contributing guide calls for it #240

PR #243, opened 2026-03-13
feat(install): add multi-distro Linux support

### LINKED ISSUE #235: feat: Add multi-distro Linux support to install script
## Problem

Hey! I run Arch and wanted to try out NOMAD, but the install script bails out immediately if you're not on Debian. Since everything runs in Docker anyway, there's no real reason the installer needs to be Debian-only — it just needs to install Docker and a couple of dependencies, which every major distro can do.

## What I'd like to change

I've already got a working version of the install script that handles multiple distros. Here's the gist of what's different:

- **Distro detection** — reads `/etc/os-release` instead of checking for `/etc/debian_version`, figures out which package manager to use from there
- **Package manager wrapper** — a simple `pkg_install` function that calls `apt-get`, `pacman`, `dnf`, `zypper`, etc. depending on what distro you're on
- **Docker install** — `get.docker.com` doesn't actually support Arch or openSUSE, so for those it installs docker from the distro's own repos instead
- **NVIDIA toolkit** — the repo setup is different for deb vs rpm vs Arch, so that's handled per-distro now

The README would also need a small update to reflect the broader distro support.

## Testing

I spun up VMs for each of the major distro families and ran the install on all of them:

- Debian 12 — works (same as before)
- Arch Linux — works
- Fedora 42 — works
- openSUSE Leap 15.6 — works

All four serve the Command Center UI after install. No changes to any application code or containers — just the install script and README.

Happy to open a PR if you're interested.

bgauger:
## Summary

The install script currently exits if you're not on Debian, but since NOMAD runs entirely in Docker there's no reason it can't work on other distros. This PR swaps out the Debian-only gate for automatic distro detection and adds support for all the major Linux families.

**What changed:**
- Distro detection via `/etc/os-release` with a `pkg_install` wrapper that calls the right package manager (apt, pacman, dnf, zypper, etc.)
- Docker installs from native repos on Arch and openSUSE since `get.docker.com` doesn't support them
- NVIDIA container toolkit setup handled per-distro (deb repos, rpm repos, Arch AUR)
- Fallback to `ip` command when `hostname` isn't available on minimal installs
- README updated to list supported distributions

**No changes to application code, containers, or compose files** — just the install script and docs.

## Test plan

Spun up cloud-init VMs and ran the full install on each:

- [x] Debian 12 (bookworm) — pass
- [x] Arch Linux — pass
- [x] Fedora 42 (Cloud Edition) — pass
- [x] openSUSE Leap 15.6 — pass

All four serve the Command Center UI after install.

Closes #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,22 @@
 Project N.O.M.A.D. is a self-contained, offline-first knowledge and education server packed with critical tools, knowledge, and AI to keep you informed and empowered—anytime, anywhere.
 
 ## Installation & Quickstart
-Project N.O.M.A.D. can be installed on any Debian-based operating system (we recommend Ubuntu). Installation is completely terminal-based, and all tools and resources are designed to be accessed through the browser, so there's no need for a desktop environment if you'd rather setup N.O.M.A.D. as a "server" and access it through other clients.
+Project N.O.M.A.D. can be installed on most Linux distributions. The install script automatically detects your distro and uses the appropriate package manager. Installation is completely terminal-based, and all tools and resources are designed to be accessed through the browser, so there's no need for a desktop environment if you'd rather setup N.O.M.A.D. as a "server" and access it through other clients.
+
+#### Supported Distributions
+| Family | Distros |
+|--------|---------|
+| Debian/Ubuntu | Debian, Ubuntu, Raspbian, Linux Mint, Pop!_OS, Elementary, Zorin, Kali |
+| Arch | Arch Linux, Manjaro, EndeavourOS, Garuda, CachyOS |
+| Fedora/RHEL | Fedora, RHEL, CentOS, Rocky, AlmaLinux, Nobara |
+| openSUSE | openSUSE Leap, openSUSE Tumbleweed, SLES |
+| Other | Void Linux, Alpine Linux |
 
 *Note: sudo/root privileges are required to run the install script*
 
 ### Quick Install (Debian-based OS Only)
 ```bash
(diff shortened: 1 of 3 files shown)


salt-peter431:
This updated `install/install_nomad.sh` worked for me on CachyOS (Arch). I was able to install the program cleanly. Thanks!

chriscrosstalk:
Reviewed the diff — this looks great. Clean distro detection, good fallback handling, and nice touches like the `ip` command fallback and `systemctl enable --now`. No concerns from my end.

One minor note: `pacman -Sy` (partial sync) is generally discouraged on Arch, but it's fine in a fresh install context.

Leaving the final merge call to @jakeaturner per his comment on #235.

bgauger:
Good catch on the -Sy — it's intentional here since -Syu would trigger a full system upgrade every time someone runs the install script, and we don't want that. Let the Arch folks update their own systems. Happy to add a code comment noting the reasoning if that helps.

hestela:
i could open an issue to discuss further but an idea for the future would maybe be to use something such as ansible to install packages so that distro specific install commands (ie yum install vs apt install and so on) dont have to be maintained. although even then thats not perfect as some distros may have something like pagename-dev but then the other OS you want packagename1 and so on (but ansible supports this well enough).



but otherwise if its only in the [un]install scripts then maybe its fine

chriscrosstalk:
UPDATE:  Keeping on the roadmap for now until we're a bit more stable.

mateowoetam:
I'm wondering if this is still being worked on in this PR or was multi-distro Linux support added already by another PR? 

... and about distros, I think NixOS should be on the list of feats.

Issue #240, opened 2026-03-13
Feat: Add ability to use a remote host for LLM chat


hestela:
I see the new CONTRIBUTING guide suggests opening an issue before adding code which was added before i opened #230 .
But proposal is:

- Add a setting for the Chat feature to set the OLLAMA_HOST
- Allow for the nomad host to install AI Chat without a GPU and/or Ollama service present
- User would be responsible for setting up their own ollama+gpu software stack on the remote machine. Ollama install is a simple curl | bash command.

Related future task I thought about would be to switch from usage of Ollama API to OpenAI API format which ollama natively supports and then users could even use other programs like LM Studio instead of Ollama.

hestela:
TIL renaming the branch causes the PR to close. but new PR is at #244

hestela:
adding openai api would probably solve #248 
i'll see about adding that to my current PR

hestela:
updated the PR to add OpenAI API support

mattbisme:
This is absolutely a must. I already have a server that could run NOMAD that doubles as a NAS. But then my Mac Studio could easily run the AI inference. No extra hardware needed.

hestela:
OK, PR should be good now. Main downside to not using ollama is that openai api doesn't support model download so you have to download model and then refresh the chat page. But model load at least works when I tried with LM studio

chriscrosstalk:
Great proposal — PR #244 covers this. Follow that PR for updates on review and merge status. Thanks for the contribution!

PR #244, opened 2026-03-13
feat(ai-chat): enable remote AI Chat host

hestela:
Adds ability to set an ollama host URL. Tested using a nomad host running in proxmox LXC and a desktop with a GPU and ollama setup with OLLAMA_HOST=0.0.0.0:11434". Allows for AI assistant to install on a host without a GPU or ollama installed.  OpenAI API servers are supported such as LM Studio and llama.cpp

User is responsible for setup of OpenAI API+gpu drivers+model download in other other machine if this is used.

Only ollama supports model download so ollama is still recommended for ease of use.

screenshot of new setting:
Screenshot_20260312_225346

Tested this PR on a couple different ollama hosts and with LM Studio

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,19 +85,15 @@ To run LLM's and other included AI tools:
 - OS: Debian-based (Ubuntu recommended)
 - Stable internet connection (required during install only)
 
-#### Running AI models on a different host
-By default, N.O.M.A.D.'s installer will attempt to setup Ollama on the host when the AI Assistant is installed. However, if you would like to run the AI model on a different host, you can go to the settings of of the AI assistant and input a URL for either an ollama or OpenAI-compatible API server (such as LM Studio, llama.cpp).  
-Note that if you use Ollama on a different host, you must start the server with this option `OLLAMA_HOST=0.0.0.0`.  
-You are responsible for the setup of Ollama/OpenAI server on the other host.
-
-#### Running AI models on a different host
-By default, N.O.M.A.D. will attempt to setup Ollama on the host when the AI Assistant is installed. However, if you would like to run the AI model on a different host, you can go to the settings of of the AI assistant and input a URL for either an ollama or OpenAI-compatible API server (such as LM Studio, llama.cpp).  
-Note that if you use Ollama on a different host, you must start the server with this option `OLLAMA_HOST=0.0.0.0`.
-You are responsible for the setup of Ollama/OpenAI server on the other host.
-
 **For detailed build recommendations at three price points ($150–$1,000+), see the [Hardware Guide](https://www.projectnomad.us/hardware).**
 
 Again, Project N.O.M.A.D. itself is quite lightweight - it's the tools and resources you choose to install with N.O.M.A.D. that will determine the specs required for your unique deployment
+
+#### Running AI models on a different host
+By default, N.O.M.A.D.'s installer will attempt to setup Ollama on the host when the AI Assistant is installed. However, if you would like to run the AI model on a different host, you can go to the settings of of the AI assistant and input a URL for either an ollama or OpenAI-compatible API server (such as LM Studio).  
+Note that if you use Ollama on a different host, you must start the server with this option `OLLAMA_HOST=0.0.0.0`.  
+Ollama is the preferred way to use the AI assistant as it has features such as model download that OpenAI API does not support. So when using LM Studio for example, you will have to use LM Studio to download models.
+You are responsible for the setup of Ollama/OpenAI server on the other host.
 
 ## Frequently Asked Questions (FAQ)
 For answers to common questions about Project N.O.M.A.D., please see our [FAQ](FAQ.md) page.
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ For more control over the installation process, copy and paste the [Docker Compo
 N.O.M.A.D. is a management UI ("Command Center") and API that orchestrates a collection of containerized tools and resources via [Docker](https://www.docker.com/). It handles installation, configuration, and updates for everything — so you don't have to.
 
 **Built-in capabilities include:**
-- **AI Chat with Knowledge Base** — local AI chat powered by [Ollama](https://ollama.com/), with document upload and semantic search (RAG via [Qdrant](https://qdrant.tech/))
+- **AI Chat with Knowledge Base** — local AI chat powered by [Ollama](https://ollama.com/) or you can use OpenAI API compatible software such as LM Studio or llama.cpp, with document upload and semantic search (RAG via [Qdrant](https://qdrant.tech/))
 - **Information Library** — offline Wikipedia, medical references, ebooks, and more via [Kiwix](https://kiwix.org/)
 - **Education Platform** — Khan Academy courses with progress tracking via [Kolibri](https://learningequality.org/kolibri/)
 - **Offline Maps** — downloadable regional maps via [ProtoMaps](https://protomaps.com)
@@ -84,6 +84,16 @@ To run LLM's and other included AI tools:
 - Storage: At least 250 GB free disk space (preferably on SSD)
 - OS: Debian-based (Ubuntu recommended)
 - Stable internet connection (required during install only)
+
+#### Running AI models on a different host
+By default, N.O.M.A.D.'s installer will attempt to setup Ollama on the host when the AI Assistant is installed. However, if you would like to run the AI model on a different host, you can go to the settings of of the AI assistant and input a URL for either an ollama or OpenAI-compatible API server (such as LM Studio, llama.cpp).  
+Note that if you use Ollama on a different host, you must start the server with this option `OLLAMA_HOST=0.0.0.0`.  
+You are responsible for the setup of Ollama/OpenAI server on the other host.
+
+#### Running AI models on a different host
+By default, N.O.M.A.D. will attempt to setup Ollama on the host when the AI Assistant is installed. However, if you would like to run the AI model on a different host, you can go to the settings of of the AI assistant and input a URL for either an ollama or OpenAI-compatible API server (such as LM Studio, llama.cpp).  
+Note that if you use Ollama on a different host, you must start the server with this option `OLLAMA_HOST=0.0.0.0`.
+You are responsible for the setup of Ollama/OpenAI server on the other host.
 
 **For detailed build recommendations at three price points ($150–$1,000+), see the [Hardware Guide](https://www.projectnomad.us/hardware).**
 
--- a/admin/app/controllers/easy_setup_controller.ts
+++ b/admin/app/controllers/easy_setup_controller.ts
@@ -1,6 +1,7 @@
 import { SystemService } from '#services/system_service'
 import { ZimService } from '#services/zim_service'
 import { CollectionManifestService } from '#services/collection_manifest_service'
+import KVStore from '#models/kv_store'
 import { inject } from '@adonisjs/core'
 import type { HttpContext } from '@adonisjs/core/http'
 
@@ -12,10 +13,14 @@ export default class EasySetupController {
   ) {}
 
   async index({ inertia }: HttpContext) {
-    const services = await this.systemService.getServices({ installedOnly: false })
+    const [services, remoteOllamaUrl] = await Promise.all([
+      this.systemService.getServices({ installedOnly: false }),
+      KVStore.getValue('ai.remoteOllamaUrl'),
+    ])
     return inertia.render('easy-setup/index', {
       system: {
         services: services,
+        remoteOllamaUrl: remoteOllamaUrl ?? '',
       },
     })
   }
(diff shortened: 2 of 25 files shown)


hestela:
updated message. this now has support for OpenAI API software like LM Studio and llama.cpp. Fixes #240
 and #248

hestela:
I tested by first installing nomad as normal but on a host that does not have a GPU. and then i edited `/opt/project-nomad/compose.yml` to change the admin image to point to the container i built and published to my local docker registry. And then i stoppped nomad, ran the update script and tried with my LM studio and ollama from my desktop. works pretty well

hestela:
This may have broken "Knowledge Base" document ingestion due to the openai api and ollama api having some differences. I'll look into it tomorrow.

hestela:
ok this should be good now. document ingestion works with ollama and openai

hestela:
I managed to get chat working with llama.cpp. I'm not familiar with it at all but i dont know if document ingestion will work since you have to switch to a different model. but with LM Studio and ollama on remote machines all features seem to work fine

chriscrosstalk:
Nice work on this! One thought — this might be a good candidate for our planned "Advanced Features" toggle in settings. Remote AI host and OpenAI API support are power-user features that could live behind that toggle to keep the default experience simple.

Leaving the code review to @jakeaturner.

hestela:
i force pushed to remove a typo in the commit that said openapi instead of openai

ptahbe:
The AI Assistant won't install. I see this
image

Is this the same issue as described above?

hestela:
> The AI Assistant won't install. I see this image
> 
> Is this the same issue as described above?

I tried setting up a fresh instance and then installing my changes and theres still sometimes some problems with ai assistant install so some general install issues seem to Still exist. This PR should allow you to install when not having a GPU on the nomad host

hestela:
did a rebase via force push, no conflicts

hestela:
i rebased, updated the AI Assistant container and others along with my locally built admin container and it still seems to work. i did notice that certain models (most dont have this issue) dont like to run due to context length issues but i dont think thats related to this patchset.
anyways proof of it working LM Studio (I have a random Unifi Specs Doc in the KB)
image

I asked qwen2.5:3B:

> what wifi standards does the UAP-AC-LR support?

response:

> The UniFi AC LR (UAP-AC-LR) supports the following Wi-Fi standards:
> 
>     802.11ac
>     802.11n
>     802.11a
>     802.11b
> 
> These are the Wi-Fi standards supported by the UAP-AC-LR.

the spec sheet i uploaded says `802.11 a/b/g/n/r/k/v/ac` so pretty good although it could just be a lucky guess as thats a common set of standards these days.

but the phi4 model will mention the knowledge base so i think everything is good
image

hestela:
ok managed to fix the `The number of tokens to keep from the initial prompt is greater than the context length (n_keep: 4705>= n_ctx: 4096)` issue, it was a small patch so I tacked it onto this PR.

hestela:
@jakeaturner , do you want me to take a look at the conflicts or will you take care of it? I'm ok either way

jakeaturner:
@hestela this looks really great so far! Only three small things of note:
1. I see that there is logic for displaying a warning in the Easy Setup Wizard when a remote is configured, but I wasn't able to get it to show up
2. It would be great if we could add a "badge" in the chat interface that says "Remote Connected" just for additional visual confirmation that requests are being sent to the remote - maybe to the left of the model selection dropdown? I think the green alert style you have for "Current configured:" would be perfect for it as well.
3. Could we use the `Alert` component for the "Model downloading is only supported when..." warning in `/settings/models`? I think we should eventually just hide most of the UI meant for local Ollama only when a remote is configured, but that's a bigger lift and not a blocker for this.

Thanks for the work on this, you are going to make a lot of folks happy! Unfortunately this is a little too much for me to fully test before squeezing into tonight's release (3/19) but we will definitely do an RC in the next few days and get this in people's hands asap.

hestela:
What branch should I target? Since you deleted the Dev branch that closes this PR and I'm not allowed to open it now

hestela:
<strike>ill base off of main and so ill have to open a new PR. will post the number when im done with requested changes</strike>

hestela:
Ok, I think we are good now, @jakeaturner i addressed all your comments. Added dark mode support for the new UI element in AI settings. and geez, the differences between openai api and ollama is crazy. 
new things:
- added detection for <think></think> since in openai it will put those in the reply but ollama just doesnt, its in a separate field
- fixes for RAG upload, mainly for openai
- connection status badge (red vs green and disconnected vs connected) status in ai chat
- warning in easy setup when theres a remote ollama/openai setup and it for now at least will refuse to download models in there.
- model download for "remote" ollama works as if its the nomad managed one, its enabled in the ai settings

note: i'm not able to properly test llama.cpp as it does not seem to support multi model usage out of the box, we will want to steer users towards ollama (best experience even in remote mode), or Lm studio (or anything similar that can hot load models). folks using llama.cpp may need to setup https://github.com/mostlygeek/llama-swap to get it properly working with nomad.

otherwise im sure we will find more features that are missing in ollama vs openai, ill stop looking for those gaps for now since i dont want perfect to be the enemy of done.

I've managed to make the LM Studio experience pretty decent, but sill use ollama when possible.

I'm not going to try and fix this here since its not a real use case IMO, but switching between ollama and lm studio (ie changing the API port in settings) is jank, you have to wait and then refresh the chat page a few times until the LLM decides to reply. 

otherwise please go ahead with review. this was rebased off of dev earlier

hestela:
and some screenshots:
ollama-download-only
chat-lm-studio
remote-settings
connected
Screenshot_20260320_133402
remote-ollama

bonus feature in easy setup, if you tried to download a model when previously connected to ollama (and it stalled, failed, etc) and you switch to openai you get a message when you come back to easy setup

ollama-only-download

hestela:
i just noticed for the new openai think detection it is lost on refresh, lmk if you want it fixed or do it later:
during chat: 
image
on refresh:
image

on ollama you just dont see the think stuff

hestela:
i'm not much of a llama.cpp user but it seems to work for only chat but to switch from the knowledge base back to chat you have to relaunch llama.cpp since it only supports 1 model at a time and nomad only supports 1 api URL for the chat and for the RAG processing. also i cant get RAG processing to work, it keeps complaining about being sent too many tokens at once and ive already tried adjusting those sizes to no avail. 

for chat you can do:
```
docker run --gpus all -p 8080:8080 -v /mnt/models/llama/:/models ghcr.io/ggml-org/llama.cpp:server-cuda --port 8080 --host 0.0.0.0 -np 2 -m /models/Qwen_Qwen2.5-3B-Instruct-GGUF_qwen2.5-3b-instruct-q4_k_m.gguf
```
for RAG you have to add --embeddings (also may need to update llama.cpp):
```
docker run --gpus all -p 8080:8080 -v /mnt/fast-intel/llama/:/models ghcr.io/ggml-org/llama.cpp:server-cuda --embeddings --port 8080 --host 0.0.0.0 -m /models/nomic-ai_nomic-embed-text-v1.5-GGUF_nomic-embed-text-v1.5.Q2_K.gguf
```
but again, RAG doesnt work right. --emedding support seems not very good in llama.cpp. lm studio and ollama handle it much better. I initially had a crash from my ~1 week old copy of the llama.cpp:server-cuda container

hestela:
FYI @KeroZelvin I pushed your fix from #451 to the branch for this PR here. https://github.com/hestela/project-nomad/tree/feat/add-remote-ollama
It has some merge conflicts fixed.
This does mostly work as it initially just installs the qdrant image just fine but then the ai assistant never shows up for me.
so you have to go back to settings and click to install ai assistant but then it still goes and pulls ollama/ollama:0.15.2 in the background (same as when i was working on this feature earlier). if i remove the patch i just added then the installer gives no real feedback (ie doesnt say its pulling qdrant) and you come back in a few minutes and ai assistant shows up but thats because it pulled the ollama container.
FWIW, I also tried to hack out the ollama/ollama:0.15.2 image requirement to no avail but your patch is generally better than my attempts that i have pushed anywhere. 

Maybe @chriscrosstalk / jake may know better but it looks like project nomad **_really_** does not want to let you use the AI Assistant unless the nomad_ollama container is running, even if you run ollama yourself or in a different host. with or without the change from KeroZelvin i see this happening, unless you saw something different with just your patch KeroZelvin?

some screenshots:
Image

Image
It's maybe ok to leave it way (especially if the advanced feature toggle comes with this) but I'll leave it to @chriscrosstalk  to decide but it otherwise seems functional (ie you have to go from easy installer to set the remote/local server and then to back to settings to run ai installer). RAG and ai chat still work with my offsite ollama with the patch i just added on top of the one from KeroZelvin. it might be an easy fix, i just ran out of time for today

KeroZelvin:
I dug through this in the PR branch, and your read looks basically right: the app is still only partially “remote Ollama aware,” and a few core paths still assume nomad_ollama exists as a local managed service/container.

The main issue I found is that configureRemote() does save ai.remoteOllamaUrl and marks the Ollama service as installed, but SystemService._syncContainersWithDatabase() later flips that service back to installed = false if there is no local nomad_ollama container. The UI uses that installed-services state to decide whether the AI Assistant is available, which lines up with what you’re seeing: Qdrant installs, but the AI Assistant never really “sticks” unless the local Ollama container also exists.

There are a couple other places where the feature is still coupled to local Ollama too:

Easy Setup skips local Ollama install when remote mode is enabled, but it does not fully guard the rest of the AI flow.
It can still dispatch model download actions, which only really make sense for a real Ollama backend and can end up pulling locally / triggering background behavior that looks like the old path.

So right now remote AI support is additive, not a full decoupling from the assumption that nomad_ollama is the thing that makes the AI Assistant “installed.”
I don’t think this is just your merge-conflict resolution or just the patch from #451. The underlying problem seems to be that NOMAD still treats “AI Assistant installed” as “local nomad_ollama container exists,” even when a remote Ollama/OpenAI-compatible endpoint is configured.

I think a workable approach might be:

Treat ai.remoteOllamaUrl as satisfying the AI Assistant install state.
Exempt nomad_ollama from the container-sync “mark not installed if container missing” logic when remote mode is configured.
Disable or reroute model download/install flows when the backend is remote and not native Ollama.
That would explain why your branch “mostly works” but still falls back to pulling ollama/ollama:0.15.2 in the background.

hestela:
that makes sense, I'm kind of thinking to leave it to Chris if one of us should continue and try to fully de-couple ollama like you mentioned or if we just make this "LLM via API" feature good enough for now to let people get going on  dev or maybe a RC build. I would maybe lean towards seeing if we can merge this to dev and/or get more testers first and then make a follow on issue/PR to fully break the ollama container requirement since that would also save ~5GB of space. 

Related, for fun i tried running nomad on a RPI4 and the PI really did not want to download/unpack the ollama container so that would help unblock lower powered platforms at some point.

hestela:
force push was to rebase and solve a conflict from dev

hestela:
ok great news, i managed to make it so that the nomad_ollama container does not get setup when you go through the easy setup wizard on a fresh install (old installs may still run it but its harmless, lets users switch between the two if theyd like) check the box for remote ollama and put in your API endpoint and it will only setup the nomad_qdrant container and the ai assistant shows up! so this should be good to go.

RAG still working fine.

FYI @chriscrosstalk 
user experience issues should be mostly solved now with my force push i just did and the new commit.

hestela:
tested switching from a previously setup local ollama to a "remote" that was my machine. it seems docker doesnt like localhost, i had to either use my nomad host IP or the hostname but it works. so then, this patch technically enables any ollama/lm-studio supported GPU. so if someone wants to do AMD or one of those pci-e AI inference cards then they should be good if they can get the openai api server setup

kane9287:
> tested switching from a previously setup local ollama to a "remote" that was my machine. it seems docker doesnt like localhost, i had to either use my nomad host IP or the hostname but it works. so then, this patch technically enables any ollama/lm-studio supported GPU. so if someone wants to do AMD or one of those pci-e AI inference cards then they should be good if they can get the openai api server setup

This is because docker doesn't use the host network by default. When a docker container uses localhost it's trying that on whatever docker network it's connected to. You have to put it on Network host to use localhost without a proxy layer or host network access defined in the compose or docker stack.

Personally I think this is cool for a stand in place server type env, but it would largely be wasteful in a true catastrophic crisis scenario. The energy that a GPU uses can be massive and that's usually not how you want to be soaking your energy structure. Ultimately I'm finding with a low end APU on a Ryzen 5 3400ge with 16gb of RAM I'm able to quickly run <=1GB models with conversational token response. I only use higher models for RAG.

hestela:
rebased. RAG and chat on remote instance still working fine.

chriscrosstalk:
UPDATE:  Staging for v1.31.0

systemofapwne:
Is there any chance, an auth-token feature could be added?
I am using ollama behind a litellm gateway-proxy to bundle different openai apis to one single endpoint. Litellm is configured to require auth-tokens. Right now, I cannot specify any auth-token so naturally, nomad cannot access my remote AI endpoint.

hestela:
I'll leave that to @chriscrosstalk to decide. It should be an easy feature to add but then adding token support would also let people use cloud based openai servers and i don't believe theres been any token/cost optimization here so we can't really do anything if someone generates a ton of cost and gets unhappy. Thats really my only concern

PR #273, opened 2026-03-13
fix(collections): update stale React devdocs ZIM URL

### LINKED ISSUE #269: Stale URL: devdocs_en_react_2026-01.zim returns 404, blocks Computing tier status
## Problem

The `devdocs_en_react` resource in `collections/kiwix-categories.json` references a URL that returns 404:

```
https://download.kiwix.org/zim/devdocs/devdocs_en_react_2026-01.zim
```

This file does not exist on Kiwix's servers. The download job retries 3 times and lands in the failed state in Redis.

### Impact

1. **Tier status is wrong** — Computing & Technology shows as "Essential" installed even though all other Standard and Comprehensive resources are present. The `getInstalledTierForCategory` check requires *all* resources in a tier to be installed, so one missing resource blocks the entire chain.

2. **Failed job blocks re-queue** — The failed job remains in Redis, and `RunDownloadJob.getByUrl` finds it on subsequent attempts, so clicking "Comprehensive" again does nothing (returns null with no user feedback).

3. **No UI indication** — The user sees "Essential" with no explanation of why Standard/Comprehensive won't download. The failed 404 is only visible in container logs.

### Steps to Reproduce

1. Fresh install or clean state
2. Select Computing & Technology → Comprehensive
3. All resources download except `devdocs_en_react` (404)
4. UI shows tier as "Essential" instead of "Comprehensive"
5. Clicking Comprehensive again does nothing

### Suggested Fix

- Remove or replace `devdocs_en_react` in `kiwix-categories.json` — this file doesn't exist on Kiwix
- Related to #127 — the graceful error handling doesn't prevent the failed job from blocking tier status advancement

### Environment

- Project N.O.M.A.D. v1.29.1
- Ubuntu Server 24.04.2 LTS

chriscrosstalk:
## Summary
- Kiwix skipped the January 2026 build of `devdocs_en_react` — the `2026-01` URL returns 404
- Updated to `2026-02` which exists and resolves correctly
- Same class of bug as PR #166 (devdocs URL renames)

Closes #269

## Test plan
- [ ] Verified `devdocs_en_react_2026-02.zim` returns HTTP 302 from Kiwix servers
- [ ] Select Computing & Technology → Standard or Comprehensive and confirm React docs download successfully

### DIFF
--- a/collections/kiwix-categories.json
+++ b/collections/kiwix-categories.json
@@ -557,10 +557,10 @@
             },
             {
               "id": "devdocs_en_react",
-              "version": "2026-01",
+              "version": "2026-02",
               "title": "React Documentation",
               "description": "React library reference and tutorials",
-              "url": "https://download.kiwix.org/zim/devdocs/devdocs_en_react_2026-01.zim",
+              "url": "https://download.kiwix.org/zim/devdocs/devdocs_en_react_2026-02.zim",
               "size_mb": 3
             },
             {


collections/kiwix-categories.json
```diff
@@ -557,10 +557,10 @@
             },
             {
               "id": "devdocs_en_react",
-              "version": "2026-01",
+              "version": "2026-02",
               "title": "React Documentation",
               "description": "React library reference and tutorials",
-              "url": "https://download.kiwix.org/zim/devdocs/devdocs_en_react_2026-01.zim",
```
hestela: I know this is a json file so it cant be done here but it might be good to use this file as a fallback maybe since one could either scan https://download.kiwix.org/zim/devdocs/ for the latest file (as at least at the moment directory listing is enabled) or since its kind of a value you can guess (ie, current_year-[01-12]) so that this file doesnt have to be maintained constantly. 

alternative would be to create an action that goes and tries to update the URLs and opens a PR on some regular cadence, maybe even 1/month works

PR #309, opened 2026-03-14
Feature/port conflict recovery

### LINKED ISSUE #308: [Feature]: Installation Script can detect port conflicts and allow user to select a new port
### Feature Category

UI/UX Improvement

### Problem Statement

when installing, there could be existing docker containers on a system that use the same ports, right now the install script just fails when that happens.   This script modification creates a method to detect the collision, and offer alternatives or let the user enter an alternative.   user can either accept recommendation, or provide a new port they want to use.  This propagates into the application as it is installed to use the new port, vs the hardcoded options.

### Proposed Solution

update script that tests the port, detects collision, and offers alternatives.

modify codebase to accept environment variables so that when a port is changed externally, instead of looking for the hardcoded port it accepts the environment variable that was set as part of the installation, listening on the alternate port.

I'm currently working on this and testing it out

### Alternative Solutions

with Dozzle (my particular conflict) being hard coded just changing the docker yaml after the fact doesn't help.   needed to expose an environmental variable that can be changed.

### Use Case

someone with lots of docker services running on a high end machine with 64gb of ram, earlier applications that have been running for years already use those ports, and since 8080 is very common as well in many management apps just makes it more customizable for someone not dedicating a hardware platform for it.

All Users

Critical - Blocking my use of N.O.M.A.D.

make environment variables instead of hard coding.   I'm already working on this and will submit a PR when I get it working.

### Examples or References

_No response_

Yes - I can write the code

### Additional Context

I've already gotten first pass done, just testing it

 1. feat(install): detect port conflicts and suggest available alternatives — the core port conflict recovery logic                                                      
 2. fix: resolve hardcoded port references for Dozzle and update script — the downstream references:                                                                     
     - SettingsLayout.tsx — Dozzle link now reads dozzlePort from Inertia shared props instead of hardcoded 9999                                                         
     - management_compose.yaml — passes DOZZLE_PORT=9999 env var to the admin container                                                                                  
     - env.ts / inertia.ts — wires the env var through to the frontend                                                                                                   
     - install_nomad.sh — seds DOZZLE_PORT in compose when port is remapped                                                                                              
     - update_nomad.sh — reads the actual admin port from the compose file instead of hardcoding 8080 (also fixed the message saying "installation" instead of "update") 

### Pre-submission Checklist

trek-e:
## Summary

Resolves #308

The installer now detects port conflicts before starting containers and offers recovery instead of aborting. When a required port (8080, 9999, 3306, 6379) is already in use, the installer identifies what process is using it, suggests up to 3 available alternatives, and lets the user choose.

### Install script (`install/install_nomad.sh`)
- **`check_port_available()`** — checks if a port is free using `ss`, `netstat`, or `/dev/tcp` fallback
- **`find_alternate_ports()`** — finds up to 3 available alternatives by trying offsets of +10000, +20000, +30000
- **`get_port_process_info()`** — identifies what process is occupying a conflicted port
- **`resolve_port_conflicts()`** — checks all 4 host ports, prompts the user for each conflict with a recommended alternative, validates input
- Remapped ports are applied to the downloaded compose file via `sed` (host-side port bindings only — container-internal ports stay the same)
- Success message shows actual URLs with remapped ports, plus a port map summary if anything was changed

### Hardcoded port references fixed
- **`SettingsLayout.tsx`** — Dozzle "Service Logs & Metrics" link was hardcoded to port `9999`. Now reads `dozzlePort` from Inertia shared props
- **`management_compose.yaml`** — passes `DOZZLE_PORT` env var to the admin container
- **`env.ts`** / **`inertia.ts`** — wires the env var through AdonisJS to the frontend
- **`update_nomad.sh`** — success message was hardcoded to port `8080`. Now reads the actual admin port from the existing compose file. Also fixed message saying "installation completed" instead of "update completed"

## User experience

When a port conflict is detected during install:
```
# Port 9999 is already in use by dozzle (needed for Dozzle (Log Viewer))
# Available alternatives: 19999 (recommended), 29999, 39999

# Enter port for Dozzle (Log Viewer) [19999]:
```

Post-install summary with custom ports:
```
# Custom port assignments:
#   Command Center:           http://192.168.1.100:18080
#   Dozzle (Log Viewer):      http://192.168.1.100:19999
#   MySQL:                    port 3306
#   Redis:                    port 6379
```

## Deployment note

This PR includes both **install-time** and **app-level** changes. For the full fix to work end-to-end:

1. **Install script changes work immediately** — port conflict detection, alternate port suggestion, and compose file rewriting all take effect as soon as the updated install script is used.
2. **UI link fix requires a new Docker image build** — the Dozzle "Service Logs & Metrics" link in `SettingsLayout.tsx` is compiled into the React bundle inside the Docker image. Until a new image is built and published from this branch, the Command Center UI will still link to `:9999` regardless of the actual Dozzle port. The `DOZZLE_PORT` env var is correctly passed through the compose file, but the running app code needs the updated `SettingsLayout.tsx` + `inertia.ts` + `env.ts` to read it.

**Merge order:** Merge → build new Docker image → new installs get both the updated install script AND the updated app code with the dynamic Dozzle port link.

## Testing
- Verified bash syntax (`bash -n`) passes on both install and update scripts
- No breaking changes to default behavior — if no ports conflict, the script behaves identically to before

### DIFF
--- a/admin/config/inertia.ts
+++ b/admin/config/inertia.ts
@@ -2,6 +2,7 @@ import KVStore from '#models/kv_store'
 import { SystemService } from '#services/system_service'
 import { defineConfig } from '@adonisjs/inertia'
 import type { InferSharedProps } from '@adonisjs/inertia/types'
+import env from '#start/env'
 
 const inertiaConfig = defineConfig({
   /**
@@ -19,6 +20,7 @@ const inertiaConfig = defineConfig({
       const customName = await KVStore.getValue('ai.assistantCustomName')
       return (customName && customName.trim()) ? customName : 'AI Assistant'
     },
+    dozzlePort: () => env.get('DOZZLE_PORT', 9999),
   },
 
   /**
--- a/admin/inertia/layouts/SettingsLayout.tsx
+++ b/admin/inertia/layouts/SettingsLayout.tsx
@@ -17,7 +17,7 @@ import useServiceInstalledStatus from '~/hooks/useServiceInstalledStatus'
 import { SERVICE_NAMES } from '../../constants/service_names'
 
 export default function SettingsLayout({ children }: { children: React.ReactNode }) {
-  const { aiAssistantName } = usePage<{ aiAssistantName: string }>().props
+  const { aiAssistantName, dozzlePort } = usePage<{ aiAssistantName: string; dozzlePort: number }>().props
   const aiAssistantInstallStatus = useServiceInstalledStatus(SERVICE_NAMES.OLLAMA)
 
   const navigation = [
@@ -29,7 +29,7 @@ export default function SettingsLayout({ children }: { children: React.ReactNode
     { name: 'Maps Manager', href: '/settings/maps', icon: IconMapRoute, current: false },
     {
       name: 'Service Logs & Metrics',
-      href: getServiceLink('9999'),
+      href: getServiceLink(String(dozzlePort ?? 9999)),
       icon: IconDashboard,
       current: false,
       target: '_blank',
--- a/admin/start/env.ts
+++ b/admin/start/env.ts
@@ -60,4 +60,11 @@ export default await Env.create(new URL('../', import.meta.url), {
   |----------------------------------------------------------
   */
   NOMAD_API_URL: Env.schema.string.optional(),
+
+  /*
+  |----------------------------------------------------------
+  | Variables for configuring companion service ports
+  |----------------------------------------------------------
+  */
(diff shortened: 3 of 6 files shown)


238SAMIxD:
I support the idea of custom ports as I would like to deploy it in my multi container cluster

chriscrosstalk:
We're adding the ability to do custom Docker Compose files for NOMAD which addresses this one as a side effect.  We will be going that route rather than making the install script more complex.  Thanks for the suggestion!

Issue #307, opened 2026-03-14
[Bug]: AI install failing


Talaraine:
### Issue Category

AI Assistant (Ollama)

### Bug Description

Hi, I'm running into a similar issue as 266 but with a twist — I actually HAVE a GPU (NVIDIA GeForce RTX 3060 Ti, 8GB VRAM) and the UI is still showing "Not detected."

Docker can see the GPU fine — nvidia-smi runs successfully inside the nomad_admin container and I've added the nvidia runtime to the compose file with the deploy/resources/reservations block. The container is definitely GPU-passthrough enabled.

However the dashboard still shows GPU: Not detected, and attempting to install the AI assistant returns a 400 error: "Failed to install service: An internal error occurred while trying to install the service."

I'm running Zorin 18 Core

### Steps to Reproduce

Install AI from GUI

### Expected Behavior

Move to Model Selection Screen

### Actual Behavior

Attempting to install the AI assistant returns a 400 error: "Failed to install service: An internal error occurred while trying to install the service."

### N.O.M.A.D. Version

1.29.0

### Operating System

Ubuntu 24.04

### Do you have a dedicated GPU?

Yes

### Service Status (if relevant)

nomad@nomad-MS-7D09:/opt/project-nomad$ docker ps
CONTAINER ID   IMAGE                                              COMMAND                  CREATED          STATUS                          PORTS                                                             NAMES
414237adfdf3   ghcr.io/crosstalk-solutions/project-nomad:latest   "/usr/local/bin/entr…"   37 minutes ago   Up 37 minutes (healthy)         0.0.0.0:8080->8080/tcp, [::]:8080->8080/tcp                       nomad_admin
ee717c5f88cd   ghcr.io/gchq/cyberchef:10.22.1                     "/docker-entrypoint.…"   2 hours ago      Up 38 minutes                   0.0.0.0:8100->80/tcp, [::]:8100->80/tcp                           nomad_cyberchef
3a3787372e21   ghcr.io/kiwix/kiwix-serve:3.8.2                    "/usr/bin/dumb-init …"   2 hours ago      Restarting (0) 56 seconds ago                                                                     nomad_kiwix_server
08fd1e3b9414   qdrant/qdrant:v1.16                                "./entrypoint.sh"        2 hours ago      Up 38 minutes                   0.0.0.0:6333-6334->6333-6334/tcp, [::]:6333-6334->6333-6334/tcp   nomad_qdrant
[...]

PR #315, opened 2026-03-15
fix: surface actual error message when service installation fails

trek-e:
## Problem

Attempting to install a service that fails preflight checks shows a generic error:

> "Failed to install service: An internal error occurred while trying to install the service."

The actual reason — already installed, installation already in progress, service not found — is never shown to the user. This makes it impossible to diagnose the root cause of failures like #307.

## Root Cause

Two issues in the error handling chain:

1. **Backend** (`system_controller.ts`): The 400 error response sent `{ error: result.message }` but the frontend expected the `message` key. The response shape was inconsistent with the success path (`{ success, message }`).

2. **Frontend** (`api.ts`): The `catchInternal` wrapper caught Axios errors from 400 responses and returned `undefined`, completely discarding the server's error message. The caller in `apps.tsx` then saw a falsy response and threw a generic "An internal error occurred" — hiding the real reason.

```
Server returns 400 { error: "Service already installed" }
  → Axios throws (400 is error status)
    → catchInternal catches, logs, returns undefined
      → installService() sees !response, throws generic error
        → User sees "An internal error occurred"
```

## Fix

**Backend**: Changed 400 response shape from `{ error: message }` to `{ success: false, message }` for consistency.

**Frontend**: Replaced `catchInternal` wrapper with direct error handling in `installService`, `affectService`, and `forceReinstallService`. On Axios errors with a response body, extracts `error.response.data.message` and returns `{ success: false, message }` so callers display the real error.

```
Server returns 400 { success: false, message: "Service already installed" }
  → Axios throws
    → API method catches, extracts message, returns { success: false, message }
      → installService() sees !response.success, shows real message
        → User sees "Failed to install service: Service already installed"
```

## Files Changed

| File | Change |
|------|--------|
| `admin/app/controllers/system_controller.ts` | Fix 400 response shape |
| `admin/inertia/lib/api.ts` | Replace `catchInternal` with direct Axios error extraction in 3 methods |

## Related

Ref #307 — this fix will surface the actual error message for that user's failed AI Assistant install, helping isolate the underlying cause.

### DIFF
--- a/admin/app/controllers/system_controller.ts
+++ b/admin/app/controllers/system_controller.ts
@@ -35,7 +35,7 @@ export default class SystemController {
         if (result.success) {
             response.send({ success: true, message: result.message });
         } else {
-            response.status(400).send({ error: result.message });
+            response.status(400).send({ success: false, message: result.message });
         }
     }
 
--- a/admin/inertia/lib/api.ts
+++ b/admin/inertia/lib/api.ts
@@ -1,4 +1,4 @@
-import axios, { AxiosInstance } from 'axios'
+import axios, { AxiosError, AxiosInstance } from 'axios'
 import { ListRemoteZimFilesResponse, ListZimFilesResponse } from '../../types/zim'
 import { ServiceSlim } from '../../types/services'
 import { FileEntry } from '../../types/files'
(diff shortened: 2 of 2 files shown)


chriscrosstalk (approved):
Great catch and excellent writeup on the error chain. The `catchInternal` wrapper was silently eating the actual error message, which made #307 impossible to diagnose remotely. Consistent `{ success, message }` shape from the backend + direct extraction on the frontend is the right fix.

jakeaturner:
Great stuff, thanks! I'd gladly welcome a future PR for a reusable error extractor for `api` but that's just a nice-to-have. This fix will make a lot of users happy!

PR #346, opened 2026-03-15
add install_nomad-cross.sh

wilsonmar:
Add installation of macOS components. 
For manual review only. I haven't looked at other scripts yet.

### DIFF
--- a/install/install_nomad-cross.sh
+++ b/install/install_nomad-cross.sh
@@ -1,0 +1,769 @@
+#!/bin/bash
+
+# Project N.O.M.A.D. Installation Script described at https://www.projectnomad.us/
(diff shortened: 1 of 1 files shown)


chriscrosstalk:
Thanks for the effort! A couple of concerns:

1. macOS isn't a supported deployment target — Docker Desktop on Mac doesn't support GPU passthrough, which limits the AI features significantly.
2. PR #243 handles multi-distro support by extending the existing install script, which is the pattern we're following.

Closing for now — we're focused on getting Linux support solid across more distros first.

wilsonmar:
Last April 2025, Docker went GA with their https://docs.docker.com/model-runner/ (DMR) open sourced at https://github.com/docker/model-runner
to replace olamma app
See https://www.docker.com/blog/announcing-docker-model-runner-ga/
See https://www.docker.com/blog/run-llms-locally/

PR #354, opened 2026-03-16
fix: add stall detection to resumable downloader

markjoldham:
Axios stream timeout only covers initial connection, not the body. Once streaming starts, a stalled server would hang indefinitely.

- Added stallTimeout param (default 60s) to DoResumableDownloadParams
- Reset a timer on every received chunk; abort if no data arrives within the window
- Treat stall errors as retriable network errors in doResumableDownloadWithRetry so partial downloads can resume from where they left off

### DIFF
--- a/admin/app/utils/downloads.ts
+++ b/admin/app/utils/downloads.ts
@@ -18,6 +18,7 @@ export async function doResumableDownload({
   url,
   filepath,
   timeout = 30000,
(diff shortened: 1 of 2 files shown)


markjoldham:
I would like to retract this.   It does not fully fix the problem, and I see I did not follow the guidelines in submitting it.   More work needs to be done on this problem.

PR #438, opened 2026-03-21
fix(ai): surface model download errors and prevent silent retry loops

chriscrosstalk:
## Summary
- Model downloads that fail (e.g., Ollama version too old for a model) now show the actual error in the Active Model Downloads UI instead of silently failing
- Version mismatch errors (Ollama 412) use `UnrecoverableError` to fail immediately instead of retrying 40 times over 40 minutes
- Stale failed jobs are cleared from Redis on retry so users aren't permanently blocked from re-attempting a download

## Context
When a user tries to download a model that requires a newer Ollama version, Ollama returns a 412 with a clear error message. Previously, `ollama_service.ts` swallowed this and returned a generic "Failed to download model" message. The BullMQ job would then retry 40 times (once per minute) before giving up, and nothing was ever shown to the user.

## Changes
| File | Change |
|------|--------|
| `ollama_service.ts` | Pass through actual error message, detect version mismatch, broadcast errors via SSE |
| `download_model_job.ts` | `UnrecoverableError` for non-retryable failures, clear stale failed jobs before dispatch |
| `useOllamaModelDownloads.ts` | Handle `percent: -1` as error state, auto-dismiss after 15 seconds |
| `ActiveModelDownloads.tsx` | Show failed downloads with red border, warning icon, and error message |

## Test plan
- [ ] Try downloading a model that requires a newer Ollama version — error should appear in Active Model Downloads with red border
- [ ] Error auto-dismisses after ~15 seconds
- [ ] Retry the same model — should dispatch a fresh job (not get stuck on stale failed job)
- [ ] Normal model downloads still work and show progress bar as before
- [ ] Update Ollama from Apps page, then retry — download should succeed

Tested on NOMAD1 (192.168.200.199) with `qwen3.5:latest` on Ollama 0.15.2.

### DIFF
--- a/admin/app/jobs/download_model_job.ts
+++ b/admin/app/jobs/download_model_job.ts
@@ -1,4 +1,4 @@
-import { Job } from 'bullmq'
+import { Job, UnrecoverableError } from 'bullmq'
 import { QueueService } from '#services/queue_service'
 import { createHash } from 'crypto'
 import logger from '@adonisjs/core/services/logger'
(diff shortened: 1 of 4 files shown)


hestela:
subscribing since changes here with ollama_service.ts will conflict with #244 . other files changed here are not touched in 244.

Issue #429, opened 2026-03-20
[Bug]: port 8090 not working after install


chrislebeck:
### Issue Category

Installation/Setup

### Bug Description

after install port 8090 to view the information just shows that the site cannot be reached.

### Steps to Reproduce

unsure

### Expected Behavior

to view the information

### Actual Behavior

getting site cannot be reached

### N.O.M.A.D. Version

1.29.1

### Operating System

Ubuntu 24.04

### Do you have a dedicated GPU?

No

### Relevant Logs

```shell

```

### Browser Console Errors (if UI issue)

```javascript

```

### Pre-submission Checklist

- [x] I have searched for existing issues that might be related to this bug
- [x] I am running the latest version of Project N.O.M.A.D. (or have noted my version above)
- [x] I have redacted any personal or sensitive information from logs and screenshots
- [x] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

avieliky:
I am reinstalling my NOMAD because of the same thing. It looks like it may be missing the kiwix container.

rhoubergdb:
I'm new to NOMAD, but I've been fighting this for (2) days. As near as I can tell, 8090 is being routed to 8080 inside the kiwix container, but that container is listening on 80 and not 8080. No amount of reinstalling (stripped it all the way down to a new Ubuntu 24.04 install) seems to help.

avieliky:
Doing a lot of testing, removing the following ZIM have made it load. gutenberg_en_lcc-s_2026-03.zim, ifixit_en_all_2025-12.zim, libretexts.org_en_bio_2025-01.zim, wikibooks_en_all_maxi_2025-10.zim Seems some either got corrupt or just failed at least for me.

artbird309:
I was having the same issue with the kiwix container constantly restarting. Tailing the logs revealed the issue `sudo docker logs -f --tail 70 nomad_kiwix_server` I found the issue `Unable to add the ZIM file 'coreyms_en_flask-tutorials_2025-10.zim' to the internal library.`

The file was actively being downloaded at the time. Kiwix was picking up the incomplete file from the zim directory and failing to load it, causing the container to crash on startup.

This may also explain the reports above of corrupt ZIMs. There are two scenarios where this can cause the container to crash:

1. A ZIM file is actively being downloaded and kiwix attempts to load the incomplete file
[...]

Issue #396, opened 2026-03-18
[Bug]: When uploading invalid pmtiles file, delete button does nothing


travelingnomad1:
### Issue Category

Installation/Setup

### Bug Description

I uploaded a malformed pmtiles file. Upload went fine, but obviously nothing showed on the map. When scrolling down in the maps manager section pressing delete, I press the button, click confirm, but the file does not get deleted. Had to delete it manually via shell

### Steps to Reproduce

1. Go to http://127.0.0.1:8080/settings/maps
2. upload a map file
3. refresh page
4. when file shows up, click delete
5. confirm deletion
6. refresh page
7. map file is still there

### Expected Behavior

clicking delete actually deletes the file from list and from disk

### Actual Behavior

clicking delete and confirm does not delete the file, but does not give an error message

### N.O.M.A.D. Version

1.29.1

### Operating System

Debian 12 (Bookworm)

### Docker Version

29.3.0

### Do you have a dedicated GPU?

No

### Relevant Logs

```shell

```

### Browser Console Errors (if UI issue)

```javascript

```

### Pre-submission Checklist

- [x] I have searched for existing issues that might be related to this bug
- [x] I am running the latest version of Project N.O.M.A.D. (or have noted my version above)
- [x] I have redacted any personal or sensitive information from logs and screenshots
- [x] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

chriscrosstalk:
Same class of bug as #372 -- corrupt or malformed files can't be deleted via the UI. Tracking the fix there. Manual workaround:
```bash
sudo rm /opt/project-nomad/storage/maps/<filename>.pmtiles
```

Issue #372, opened 2026-03-16
[Bug]: ZIM files not deleting properly


PattheRat22:
### Issue Category
Content Management (ZIM, Maps, Collections)

### Bug Description
I was downloading Wikipedia, and somehow the file got corrupted while it was downloading. When I went to the Content Manager and attempted to delete the file, it did not delete the file. Instead, it gave me an error message in the web console and it kept giving me errors in the 'nomad_admin' log.

### Steps to Reproduce
1. Click on 'Content Manager'
2. Delete any file
### Expected Behavior

The corrupt file would be deleted, and it would disappear from the list of .zim files.

### Actual Behavior
The file was not deleted, and it spat out an error in the web console, and it kept spitting out errors in nomad_admin around every 10 seconds.

### N.O.M.A.D. Version
1.29.1

### Operating System
Ubuntu 24.04

### Docker Version
Docker version 29.3.0

### Do you have a dedicated GPU?
No

### System Specifications
AMD Ryzen™ 7 4800U with Radeon™ Graphics × 16
32.0 GiB Memory
176.6 GB/308.3 GB storage available

### Service Status (if relevant)
c349be4cdcc7 ghcr.io/kiwix/kiwix-serve:3.8.2 "/usr/bin/dumb-init …" 22 hours ago Restarting (0) 16 seconds ago nomad_kiwix_server

### Relevant Logs
nomad_admin: 
```shell
Error: read ECONNRESET
at TCP.onStreamRead (node:internal/stream_base_commons:216:20) {
errno: -104,
code: 'ECONNRESET',
syscall: 'read'
}
```
The next error kept repeating every 10 seconds after the first error appeared:
```
Error: write EPIPE
at afterWriteDispatched (node:internal/stream_base_commons:159:15)
at writeGeneric (node:internal/stream_base_commons:150:3)
at Socket._writeGeneric (node:net:966:11)
at Socket._write (node:net:978:8)
at writeOrBuffer (node:internal/streams/writable:572:12)
at _write (node:internal/streams/writable:501:10)
at Writable.write (node:internal/streams/writable:510:10)
at EventEmitter.sendCommand (/app/node_modules/ioredis/built/Redis.js:406:29)
at EventEmitter.client (/app/node_modules/ioredis/built/utils/Commander.js:90:25)
at /app/node_modules/ioredis/built/redis/event_handler.js:65:22 {
errno: -32,
code: 'EPIPE',
syscall: 'write'
}
```
[...]

PR #448, opened 2026-03-21
fix(downloads): stage downloads to .tmp to prevent Kiwix loading partial files

artbird309:
> **Note:** This PR was developed with the assistance of [Claude Code](https://claude.ai/claude-code) (Anthropic's AI coding tool). It is intended as a proposed starting point to address the issue — I'd welcome review and feedback from the maintainers before it's considered merge-ready.

## Problem

Kiwix crashes on startup when it encounters a partial or corrupt ZIM file on disk. This happens when a download is in progress (or failed mid-way) and Kiwix is restarted — it picks up the incomplete `.zim` file and fails with `Unable to add the ZIM file to the internal library`.

This was confirmed on a live installation: `gutenberg_en_lcc-t_2026-03.zim` (6.2GB, appeared complete by size but had invalid internal structure) crashed Kiwix on every restart until manually removed.

Ref #372 (also related: #429, #396 — same root cause affects `.pmtiles` map files)

## Fix

`doResumableDownload()` now writes to `filepath + '.tmp'` and only atomically renames to the final path on successful completion.

- Kiwix globs for `*.zim` — `.tmp` files are invisible to it
- `ZimService.list()` already filters `.endsWith('.zim')` — `.tmp` files are invisible here too
- `fs.rename()` is a single `rename(2)` syscall — atomic on the same filesystem
- The same fix covers `.pmtiles` map downloads, which go through the same code path

If a download is interrupted, the `.tmp` file remains on disk as a resume checkpoint (existing resume logic unchanged), but Kiwix will never see it.

## Changes

Single file modified: `admin/app/utils/downloads.ts`

- Add `import { rename } from 'fs/promises'`
- Compute `tempPath = filepath + '.tmp'` at the start of `doResumableDownload`
- Resume check now reads `getFileStatsIfExists(tempPath)` instead of `filepath`
- Add idempotency check: if final file already exists at correct size, return early
- Add `.tmp`-complete shortcut: if `.tmp` is already full size (complete but never renamed), just rename and return
- Non-range-request cleanup now deletes `tempPath` instead of `filepath`
- Write stream now targets `tempPath`
- On `finish`: atomically `rename(tempPath, filepath)` before calling `onComplete` / resolving

## Call sites audited

All 4 callers of `doResumableDownload` / `doResumableDownloadWithRetry` were reviewed:

- **`RunDownloadJob`** (ZIM + map queue downloads) — `onComplete` receives final `filepath` after rename ✅
- **`collection_update_service.ts`** (resource updates, `forceNew: true`) — bypasses early returns, downloads fresh to `.tmp`, renames correctly ✅
- **`map_service.ts` `downloadBaseAssets()`** — passes `filepath: tempTarPath`; final file lands at `tempTarPath` after rename; subsequent `getFileStatsIfExists(tempTarPath)` and `extract({ file: tempTarPath })` both work ✅
- **`docker_service.ts` Kiwix pre-install** — downloads Wikipedia ZIM to final `.zim` path; staged via `.tmp`, renamed on completion ✅

## Testing

Tested locally using a minimal `docker-compose` environment (admin built from this branch, MySQL, Redis, Kiwix 3.8.2):

- **`.tmp` staging (ZIM)** — file appears as `.zim.tmp` throughout, renamed to `.zim` on completion ✅
- **`.tmp` staging (maps)** — multiple `.pmtiles.tmp` files staged simultaneously, all renamed correctly ✅
- **Kiwix invisible to `.tmp`** — restarted Kiwix mid-download (584MB into a 4.7GB file), started cleanly with zero errors ✅
- **Content serving while downloading** — existing ZIM content accessible via Kiwix during an active download ✅
- **Multiple concurrent downloads** — two files staged simultaneously, both renamed correctly ✅
- **TypeScript compilation** — `tsc --noEmit` exit code 0 ✅

🤖 Developed with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/admin/app/utils/downloads.ts
+++ b/admin/app/utils/downloads.ts
@@ -86,15 +86,27 @@ export async function doResumableDownload({
     headers.Range = `bytes=${startByte}-`
   }
 
-  const response = await axios.get(url, {
-    responseType: 'stream',
-    headers,
-    signal,
-    timeout,
-  })
+  const fetchStream = (hdrs: Record<string, string>) =>
+    axios.get(url, { responseType: 'stream', headers: hdrs, signal, timeout })
+
+  let response = await fetchStream(headers)
 
   if (response.status !== 200 && response.status !== 206) {
     throw new Error(`Failed to download: HTTP ${response.status}`)
+  }
+
+  // If we requested a range but the server returned 200 (ignored the Range header),
+  // appending would corrupt the .tmp file — delete it and restart from byte 0.
+  if (headers.Range && response.status === 200) {
+    response.data.destroy()
+    await deleteFileIfExists(tempPath)
+    startByte = 0
+    appendMode = false
+    delete headers.Range
+    response = await fetchStream(headers)
+    if (response.status !== 200 && response.status !== 206) {
+      throw new Error(`Failed to download: HTTP ${response.status}`)
+    }
   }
 
   return new Promise((resolve, reject) => {
@@ -149,7 +161,6 @@ export async function doResumableDownload({
       flags: appendMode ? 'a' : 'w',
     })
 
-    // Handle errors and cleanup
     const cleanup = (error?: Error) => {
       clearStallTimer()
       progressStream.destroy()
@@ -163,7 +174,6 @@ export async function doResumableDownload({
     response.data.on('error', cleanup)
     progressStream.on('error', cleanup)
     writeStream.on('error', cleanup)
-    writeStream.on('error', cleanup)
 
     signal?.addEventListener('abort', () => {
       cleanup(new Error('Download aborted'))
@@ -175,8 +185,15 @@ export async function doResumableDownload({
         // Atomically move the completed .tmp file to the final path
         await rename(tempPath, filepath)
       } catch (renameError) {
-        reject(renameError)
-        return
+        // A parallel job may have completed the same file first — treat as success
+        // if the destination already exists at the expected size.
+        const existing = await getFileStatsIfExists(filepath)
+        if (existing && Number(existing.size) === totalBytes && totalBytes > 0) {
+          // fall through to resolve
+        } else {
+          reject(renameError)
+          return
+        }
       }
       if (onProgress) {
         onProgress({
@@ -228,7 +245,7 @@ export async function doResumableDownloadWithRetry({
       })
 
       return result // return on success
-    } catch (error) {
+    } catch (error: any) {
       attempt++
       lastError = error as Error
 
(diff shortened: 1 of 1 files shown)


chriscrosstalk (dismissed):
Clean fix for a real problem. Atomic rename on completion is the right approach, and the call site audit is thorough. This directly addresses the root cause behind #372 and the related class of corrupt-file issues (#429, #396).

Approved from my end.

chriscrosstalk:
This is the fix for the #1 pain point right now — Kiwix crash-looping on partial downloads is hitting a ton of fresh installs (#539, #502, #573). Small change, big impact. Leaving for Jake to merge. Thanks for the contribution.

LHammonds:
FYI - There is a zimcheck tool that can validate a .zim file before you transition it from temporary download to the final location for the service to read it.

artbird309:
Update (2026-03-30): PR #562 has been opened to implement post-download ZIM validation. The testing data below may be relevant context for that work.

I installed zimcheck and tested it against 71 ZIM files on a test server (Proxmox LXC, 40 vCPUs, Debian Trixie, zimcheck 3.5.0, storage on RAIDZ1 of 3x 10K SAS drives).

**`-I` (integrity):** 36 of 71 files failed, all with the identical error: "Full Title index table outside (or not fully inside) ZIM file." The failure pattern lines up with ZIM age — older files pass, newer ones fail. For example, `askubuntu.com_en_all_2024-03` passes while `askubuntu.com_en_all_2025-12` fails. The same pattern held across unrelated sources: devdocs, libretexts, zimgit, and older stackexchange files all pass; wikipedia, wikibooks, freecodecamp, gutenberg, and newer stackexchange files all fail. I spot-checked a couple of the failing files in Kiwix (3dprinting and freecodecamp) and they loaded, though I did not verify all 36. This looks like a false positive in zimcheck 3.5.0 against the newer ZIM title index format rather than genuine corruption.

**`-C` (checksum):** All 20 files tested passed. Timing varied significantly by file size — small files finished in under a minute, but `wikipedia_en_all_maxi_2026-02` took **8 minutes 17 seconds** on this 40-vCPU server. Sequential read benchmark on this storage was ~728 MiB/s (ZFS ARC assisted), write ~79 MiB/s. On lower-spec NOMAD target hardware the checksum time would be considerably longer.

Sharing this data in case it's useful for deciding how to proceed.

*Data reviewed with assistance from Claude Sonnet 4.6 (Anthropic).*

chriscrosstalk:
UPDATE:  After some initial testing, we found an alternate way to accomplish this - while we will likely cherry pick some of this code, we should have effectively mirrored this functionality in the new version.

jakeaturner (approved):
Thanks @artbird309 ! Made a couple of refinements/additions here. We ended up solving most of the original reason for this PR (Kiwix being picky about ZIM files), but the download improvements should still be useful nonetheless.

PR #461, opened 2026-03-22
Feat: User Defined URL for Apps

kane9287:
**Feature:** This PR is to add user defined URLs for apps on NOMAD to allow proper redirection when running behind a reverse proxy. 

**AI Disclosure:** I am not a developer and I used Perplexity.ai to plan the implementation cand Claude to write the code, it should be 100% reviewed for inaccuracy, simplification, and/or security

The primary change started by updating navigation.ts to allow full URL and not force the port number. To make this user editable additional code was added to apps.tsx which includes an "Edit URL" button and a description/instructions below. Kept themeing and font consistent with the remaineder of the app.

Additional changes were to various ts modules to ensure this called and runs correctly. This was tested in my private environment and works without issue.

Attached images for reference

nomad-update
nomad-update-2

### DIFF
--- a/admin/app/controllers/system_controller.ts
+++ b/admin/app/controllers/system_controller.ts
@@ -3,7 +3,7 @@ import { SystemService } from '#services/system_service'
 import { SystemUpdateService } from '#services/system_update_service'
 import { ContainerRegistryService } from '#services/container_registry_service'
 import { CheckServiceUpdatesJob } from '#jobs/check_service_updates_job'
-import { affectServiceValidator, checkLatestVersionValidator, installServiceValidator, subscribeToReleaseNotesValidator, updateServiceValidator } from '#validators/system';
+import { affectServiceValidator, checkLatestVersionValidator, installServiceValidator, subscribeToReleaseNotesValidator, updateServiceLocationValidator, updateServiceValidator } from '#validators/system';
 import { inject } from '@adonisjs/core'
 import type { HttpContext } from '@adonisjs/core/http'
 
@@ -160,6 +160,18 @@ export default class SystemController {
         } else {
             response.status(400).send({ error: result.message })
         }
+    }
+
+    async updateServiceLocation({ request, response }: HttpContext) {
+        const payload = await request.validateUsing(updateServiceLocationValidator)
+        const Service = (await import('#models/service')).default
+        const service = await Service.findBy('service_name', payload.service_name)
+        if (!service) {
+            return response.status(404).send({ error: 'Service not found' })
+        }
+        service.ui_location = payload.ui_location
+        await service.save()
+        return response.send({ success: true, message: 'Service location updated successfully' })
     }
 
     private async getHostArch(): Promise<string> {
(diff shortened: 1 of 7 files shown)


kane9287:
On further testing this could have an effect on other URL areas. Specifically maps is having issues loading (working on a fix) and dozzle was a different part of settings so it wasn't adjusted. In my private env I put dozzle settings in the "Apps" exactly like the others including the edit URL button.

kane9287:
Adjusted the mapcomponent.tsx to have it work with the URL 

import maplibregl from 'maplibre-gl'
import 'maplibre-gl/dist/maplibre-gl.css'
import { Protocol } from 'pmtiles'
import { useEffect, useState } from 'react'

// Register the PMTiles protocol at module level so it is available
// before MapLibre mounts and begins fetching tiles.
const _pmtilesProtocol = new Protocol()
maplibregl.addProtocol('pmtiles', _pmtilesProtocol.tile)

/**
 * Rewrite any absolute URL in the style JSON so that the host matches
 * window.location.origin.  This is required when the app is accessed
 * through a reverse proxy: the server builds tile/sprite/glyph URLs from
 * request.host() (the internal host), but the browser can only reach those
 * resources through the proxy.
 */
function rewriteStyleUrls(style: any): any {
  const origin = window.location.origin

  // Rewrite pmtiles:// source URLs
  if (style.sources) {
    for (const source of Object.values(style.sources) as any[]) {
      if (typeof source.url === 'string' && source.url.startsWith('pmtiles://')) {
        const inner = source.url.slice('pmtiles://'.length)
        try {
          const url = new URL(inner)
          source.url = `pmtiles://${origin}${url.pathname}`
        } catch {}
      }
    }
  }

  // Rewrite sprite URL (string or array form)
  if (typeof style.sprite === 'string') {
    const match = style.sprite.match(/^https?:\/\/[^/]+(\/.*)?$/)
    if (match) style.sprite = `${origin}${match[1] ?? ''}`
  }

  // Rewrite glyphs URL — preserve {fontstack}/{range} template tokens
  if (typeof style.glyphs === 'string') {
    const match = style.glyphs.match(/^https?:\/\/[^/]+(\/.*)?$/)
    if (match) style.glyphs = `${origin}${match[1] ?? ''}`
  }

  return style
}

export default function MapComponent() {
  const [mapStyle, setMapStyle] = useState<object | undefined>(undefined)

  useEffect(() => {
    fetch('/api/maps/styles')
      .then((r) => r.json())
      .then((style) => setMapStyle(rewriteStyleUrls(style)))
      .catch(console.error)
  }, [])

  if (!mapStyle) return null

  return (
    <MapProvider>
@@ -18,7 +67,7 @@ export default function MapComponent() {
          width: '100%',
          height: '100vh',
        }}
        mapStyle="/api/maps/styles"
        mapStyle={mapStyle}
        mapLib={maplibregl}
        initialViewState={{
          longitude: -101,

chriscrosstalk:
Thanks for putting this together — and for being upfront about the AI-assisted approach. We actually have something similar in the works on the roadmap already, so we're going to close this out. Appreciate the contribution\!

kane9287:
No problem! That's awesome to hear can't wait to see the changes.

PR #470, opened 2026-03-22
security: fix CORS, CSRF, CSP, HSTS, sessions, and path traversal

zacharyjleach-stack:
- CORS: restrict origin from wildcard '*' to app URL from env (prevents cross-origin requests from arbitrary sites)
- Sessions: enable @adonisjs/session with cookie store and httpOnly/secure cookie flags; uncomment session middleware in kernel.ts
- CSRF: enable shield CSRF protection (requires sessions); uses XSRF-TOKEN cookie mechanism compatible with Inertia.js/Axios; exempts /api/health and SSE transmit endpoints
- CSP: enable Content Security Policy with restrictive directives (no object-src, no frame-src, self-only script/style/connect/font)
- HSTS: enable HTTP Strict Transport Security in production only
- Path traversal: tighten filenameParamValidator to block /, \, .., and shell special characters; reduce max length from 4096 to 255
- env: add URL to .env.example; uncomment SESSION_DRIVER validation in env.ts

https://claude.ai/code/session_01WfRC4tDeYprykhMrg4PxX6

### DIFF
--- a/electron/assets/README.md
+++ b/electron/assets/README.md
@@ -1,0 +1,14 @@
+# Electron App Icons
+
+Place the following icon files here for the packaged desktop app:
+
+| File            | Size        | Platform       |
+|-----------------|-------------|----------------|
+| `icon.png`      | 512×512 px  | Linux (AppImage, deb) |
+| `tray-icon.png` | 32×32 px    | System tray (all platforms) |
+| `icon.icns`     | macOS bundle | macOS         |
+| `icon.ico`      | Multi-size ICO | Windows     |
+
+The `tray-icon.png` is also used as a fallback tray icon during development.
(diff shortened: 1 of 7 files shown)


jakeaturner:
There seems to be three distinct things happening in this PR: security changes, adding content categories, and adding Electron support. If you would like any of these changes to be considered, please open separate PR's for each so that we can review them piecewise and merge in changes individually.

Session-based auth is also being enabled in this PR, but Nomad doesn't have any UI built for authentication at this time. Additionally, `hsts` is being enforced in production, but this would require all users to setup HTTPS when there are only a slim percentage of end users who have Nomad behind a reverse proxy.

PR #480, opened 2026-03-23
Update management_compose.yaml type:bug

dx4956s:
changed mysql retries from 3 to 10 as it was failing on my machine


jakeaturner:
Thanks for the contribution!

Issue #271, opened 2026-03-13
feat: Advanced Features toggle in settings


chriscrosstalk:
As NOMAD's community grows, we're seeing a lot of feature requests from power users that go beyond the turnkey single-machine experience. Rather than bloating the default UI or rejecting these outright, we should add an "Advanced Features" toggle in settings that unlocks power-user capabilities.

**Current candidates:**
- Docker Compose / Swarm multi-machine deployments (#226)
- Configurable Knowledge Base upload size limit (#259)
- Custom storage location (#264)
- Remote AI host / OpenAI API support (#244)

**Concept:**
- Toggle off by default — keeps the experience simple for non-technical users
- When enabled, surfaces additional settings and options throughout the UI
- Each advanced feature should still be well-documented and user-friendly

This list will grow as more power-user requests come in. The toggle itself is a Phase 2 priority.

chriscrosstalk:
Adding to the scope here: include an "Automatically Check for Updates" toggle (defaulted to ON). Requested in #284 — power users want the option to disable the sidecar updater and manage updates themselves.

apcameron:
Do we have an expected timeline for Configurable Knowledge Base upload size limits as the current limit is way too low?

chriscrosstalk:
@apcameron -- no specific timeline yet for this. KB upload limits are part of the broader Advanced Features settings planned here. It's on the roadmap but hasn't been scheduled for a release yet.

boushley:
I'm one of the users looking to integrate this great tool into my homelab setup that already contains a reverse proxy. I took a stab at adding an Advanced Settings section. I have not gotten things fully running locally yet so haven't fully validated this. However I wanted to look for input early to make sure I don't polish something that is out of line with your vision for this feature.

You can [see the proposed changes here](https://github.com/Crosstalk-Solutions/project-nomad/pull/657)

chriscrosstalk:
Thanks for taking a stab at this. We saw PR #657 -- appreciate the initiative. Advanced Settings is scoped as a core piece of v1.4 and we'll be building it out as part of that release. The overall direction (toggle-gated settings section) is aligned with what we have planned, so your instincts were right even if the PR didn't land.

Shurov:
[...]

PR #489, opened 2026-03-23
feat(system): add disk space usage notifications

### LINKED ISSUE #485: [Feature]: Disk space usage notifications to warn users before storage runs out
**Is your feature request related to a problem?**
Yes. N.O.M.A.D. allows users to download large files (ZIM libraries, AI models, maps) that can quickly fill disk space. Currently, the system collects disk usage data via the `disk-collector` container but **does not alert the user** when storage is running low. Users only discover the problem when downloads fail or services crash.

**Describe the solution you'd like**
Add a disk space monitoring system that displays a dismissible warning banner in the UI when usage exceeds configurable thresholds (e.g., 85%, 90%, 95%). The backend already has disk info available via `SystemService.getSystemInfo()` — this feature would add threshold-checking logic and surface alerts through the existing notification system.

**Describe alternatives you've considered**
- Relying on OS-level alerts (not practical for headless/server installs accessed via browser)
- Blocking downloads when disk is full (reactive, not proactive)

**Additional context**
Related to #367 (storage path management). This complements that feature by adding awareness before storage becomes a problem.

LuisMIguelFurlanettoSousa:
## Summary

Adds a global dismissible alert banner that warns users when disk usage exceeds configurable thresholds. The banner appears on all pages via `AppLayout` and reappears if the alert level worsens after being dismissed.

- New `GET /api/system/disk-status` endpoint
- `getDiskStatus()` method in `SystemService` — reads existing `disk-collector` data
- Configurable thresholds via KVStore: `disk.warningThreshold` (default: 85%), `disk.criticalThreshold` (default: 95%)
- `useDiskAlert` React hook with 45s polling + localStorage-based dismiss
- `DiskAlertBanner` component using existing `Alert` component (warning/error variants)
- Banner reappears when alert level worsens (e.g., dismissed at warning → reappears at critical)

## Test Plan
- [ ] Verify disk alert banner appears when disk usage exceeds 85%
- [ ] Verify banner changes from warning (orange) to critical (red) at 95%
- [ ] Verify dismissing the banner persists across page navigations
- [ ] Verify banner reappears when level worsens (dismiss at warning, reaches critical)
- [ ] Verify `GET /api/system/disk-status` returns correct level, threshold, usage, and disk name

Closes #485

### DIFF
--- a/admin/app/controllers/system_controller.ts
+++ b/admin/app/controllers/system_controller.ts
@@ -113,6 +113,10 @@ export default class SystemController {
         return await this.systemService.subscribeToReleaseNotes(reqData.email);
     }
 
+    async getDiskStatus({}: HttpContext) {
+        return await this.systemService.getDiskStatus();
+    }
+
     async getDebugInfo({}: HttpContext) {
         const debugInfo = await this.systemService.getDebugInfo()
         return { debugInfo }
--- a/admin/app/services/system_service.ts
+++ b/admin/app/services/system_service.ts
@@ -521,6 +521,59 @@ export class SystemService {
     return parseFloat((bytes / Math.pow(k, i)).toFixed(decimals)) + ' ' + sizes[i]
   }
 
+  async getDiskStatus(): Promise<{
+    level: 'none' | 'warning' | 'critical'
+    threshold: number
+    highestUsage: number
+    diskName: string
+  }> {
+    try {
+      const diskInfoRawString = await getFile(
+        path.join(process.cwd(), SystemService.diskInfoFile),
+        'string'
+      )
+
+      const diskInfo = (
+        diskInfoRawString
+          ? JSON.parse(diskInfoRawString.toString())
+          : { diskLayout: { blockdevices: [] }, fsSize: [] }
+      ) as NomadDiskInfoRaw
+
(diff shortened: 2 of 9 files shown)


chriscrosstalk:

chriscrosstalk:
Thanks for the work here. Taking a step back though: NOMAD already surfaces disk usage on the System Information page (`/settings/system`), where users can see exactly how much space is used and free across all detected storage devices. A global polling banner duplicates that information and adds visible UI overhead on every page for a problem users can already check whenever they want to.

The 45-second polling interval also adds load across every authenticated session for what's effectively a passive monitoring feature — and disk-fill scenarios on NOMAD are often intentional (someone with a 4TB drive who's downloaded 3.6TB of content did that on purpose, and a "critical" alert at 95% would be noise rather than signal).

Closing this out as not aligned with our current direction. The existing Settings → System page is the right home for disk-usage visibility, and if we eventually want threshold-based notifications, that'd more naturally fit into the broader Advanced Features settings work tracked in #271 rather than as a standalone always-on banner.

Appreciate the contribution effort and please don't let this discourage you from future PRs — your work on the security audit items (#552) is exactly the kind of contribution we want to land.

PR #507, opened 2026-03-23
Feature/arm64 support

suranyami:


trelane:
I run on MacOS and support this.  Nvidia is far from the only, or even the easiest way to do AI.

PR #510, opened 2026-03-24
feat: add support for Fedora-based systems in installation script

StepTwo33:


jakeaturner:
@StepTwo33 thank you very much for your work on this, but we do already have a PR open for multi-OS install support in #243 . This will be available in the near future after we work through most of the issues reported in our initial release(s).

Issue #211, opened 2026-03-09
Security audit: 5 medium-priority hardening items


chriscrosstalk:
## Summary

Pre-launch security audit (PR #210) identified 5 medium-priority items that should be addressed after launch. None are exploitable remotely — they require LAN access.

### 1. Dozzle Web Shell Access
**File:** `install/management_compose.yaml:56`

`DOZZLE_ENABLE_SHELL=true` gives anyone on the LAN a web shell into containers via port 9999. Combined with Docker socket access in `nomad_admin`, this creates a path from LAN access to host root.

**Fix:** Set `DOZZLE_ENABLE_SHELL=false`. Log viewing and container restart still work.

### 2. Unvalidated Settings Key Read
**File:** `admin/app/controllers/settings_controller.ts`

`GET /api/system/settings?key=...` accepts any arbitrary key string. The write endpoint validates against an enum, but the read doesn't. Currently harmless but could leak sensitive data if new keys are added.

**Fix:** Apply the same enum validation to the read endpoint.

### 3. Content Update URL Injection
**File:** `admin/app/validators/common.ts`

`POST /api/content-updates/apply` accepts `download_url` directly from the client. An attacker could supply any URL. The URL should be looked up server-side from the cached content manifest.

**Fix:** Validate `download_url` against the cached manifest, or rely on the loopback/link-local protection added in PR #210.

### 4. Verbose Error Messages
**Files:** `rag_controller.ts`, `docker_service.ts`, `system_update_service.ts`

Several controllers return raw `error.message` in API responses, potentially leaking internal paths, stack details, or Docker error messages.

**Fix:** Return generic error messages in production. Log details server-side.

### 5. npm Dependency Vulnerabilities
- `tar` (HIGH) — CVE via `node-pre-gyp` → `@openzim/libzim`
- `ajv` (MODERATE) — prototype pollution via `@inertiajs/react`

**Fix:** Update upstream dependencies when patches are available. `npm audit fix --force` can address `ajv` but may introduce breaking changes.

## Context

Full audit report attached to PR #210 as `admin/docs/security-audit-v1.md`.

chriscrosstalk:
Closing this out -- 3 of 5 items are now shipped:

- **Item 1** (Dozzle shell) -- Fixed in PR #287
- **Item 2** (settings key validation) -- Fixed in v1.31.0 via PR #517
[...]

PR #517, opened 2026-03-24
fix(security): validate key parameter on settings read endpoint

LuisMIguelFurlanettoSousa:
## Summary

Adds enum validation to the `GET /api/system/settings` endpoint, matching the validation already present on the write endpoint (`PATCH`).

## Problem

Identified in security audit #211 (item 2): the settings read endpoint accepts any arbitrary key string via query parameter without validation. While the write endpoint validates against `SETTINGS_KEYS` enum, the read endpoint casts the raw input directly with `as KVStoreKey`, bypassing TypeScript's type safety at runtime.

This could leak sensitive data if new KVStore keys are added in the future.

## Fix

- Added `getSettingSchema` validator in `admin/app/validators/settings.ts` using the same `vine.enum(SETTINGS_KEYS)` pattern
- Applied validation in `SettingsController.getSetting()` — requests with invalid keys now return a 422 validation error instead of querying the database
- Removed unused `KVStoreKey` type import from the controller

## Test Plan
- [ ] `GET /api/system/settings?key=ui.theme` → returns value (valid key)
- [ ] `GET /api/system/settings?key=secret.internal` → returns 422 (invalid key)
- [ ] `GET /api/system/settings?key=` → returns 422 (empty key)
- [ ] Existing UI settings (theme toggle, AI name, etc.) continue working

Ref #211


chriscrosstalk:
Looks good — clean fix, and it directly addresses item 2 from #211. Leaving this for @jakeaturner to review and merge.

chriscrosstalk:
This addresses item 2 from our security audit (#211) — small and clean fix. Leaving for Jake to merge. Thanks for the contribution.

chriscrosstalk:
UPDATE:  Staging for v1.31.0

Issue #486, opened 2026-03-23
[Feature]: Add internationalization (i18n) support for multi-language UI


LuisMIguelFurlanettoSousa:
**Is your feature request related to a problem?**
Yes. All UI text in the Command Center is hardcoded in English. For a project designed to work anywhere in the world — including emergency and humanitarian scenarios — the lack of multi-language support limits accessibility for non-English-speaking users.

**Describe the solution you'd like**
Add i18n infrastructure using `react-i18next` (or similar) to the Inertia/React frontend:
- Extract hardcoded strings into translation files (JSON namespace-based)
- Add a language selector in Settings
- Persist the user's language preference via the existing `KVStore`
- Include English (default) and Brazilian Portuguese (pt-BR) as initial translations to serve as a template for community contributions

**Describe alternatives you've considered**
- Browser-based translation (Google Translate widget) — breaks offline usage
- Server-side i18n via AdonisJS — more complex, React-side is simpler for this UI-driven app

**Additional context**
This sets the foundation for community-driven translations. Once the infrastructure exists, contributors can add languages by simply adding a JSON file.

chriscrosstalk:
Appreciate the contribution. Multi-language support is on the roadmap, but we're not ready to take it on yet -- i18n adds maintenance overhead since every UI change needs translations updated across all supported languages. When we're ready to tackle this, your PR will be a good reference. Closing for now.

Issue #249, opened 2026-03-13
[Feature]: Support for other languages for Content and UI


martinseener:
### Feature Category

Content Management (ZIM, Maps, Collections)

### Problem Statement

Hello,
I am following Crosstalk Solution for a long time and I was quite excited to see the announcement of Project N.O.M.A.D.
I am located in Germany and so I am curious wether support for other countries/languages is planned or if there is some guidance available to start submitting PRs from a fork to add support.
I am thinking about adding first support for multiple languages for the whole UI and step by step also for the Content.
Best,
Martin

### Proposed Solution

See above.

### Who would benefit from this feature?

All Users

### How important is this feature to you?

High - Would significantly improve my experience

### Would you be willing to help implement this?

Yes - I can write the code

### Pre-submission Checklist

- [x] I have searched for existing feature requests that might be similar
- [x] This feature aligns with N.O.M.A.D.'s mission of offline-first knowledge and education
- [x] I understand that feature requests are not guaranteed to be implemented

chriscrosstalk:
Thanks Martin — and thanks for following along! Multi-language support for both the UI and content is absolutely something we're planning to implement.

Since you're offering to help — we'd love the contribution! If you want to start working on it, feel free to submit a PR when you've got something sorted. UI localization (i18n framework + language packs) would be a great starting point, and we can iterate from there.

We've also launched a public roadmap where the community can vote on priorities: https://roadmap.projectnomad.us

PR #490, opened 2026-03-23
feat(i18n): add internationalization support with react-i18next

LuisMIguelFurlanettoSousa:
## Summary

Adds i18n infrastructure using `react-i18next` with English and Brazilian Portuguese (pt-BR) translations for the main pages. Includes a language selector in Settings > System > Preferences.

- `react-i18next` + `i18next` dependencies
- Translation files: `admin/inertia/locales/{en,pt-BR}/{common,home,settings,layout}.json`
- `useTranslation` applied to: AppLayout, SettingsLayout, Footer, Home, Settings/System
- Language persistence: `localStorage` (immediate) + KVStore `ui.language` (fire-and-forget sync)
- Language selector dropdown in Settings > System > Preferences

### What's translated
- Layout: sidebar navigation, header ("Command Center"), footer, back link
- Home: dashboard card labels and descriptions (Easy Setup, Install Apps, Docs, Settings, Maps)
- Settings/System: all section headers, labels, gauges, GPU alerts, modal texts
- Global: button labels, alert messages

### What's NOT included (future work)
- Pages: Chat, Maps, Docs, Benchmark, Easy Setup, About
- Browser language auto-detection
- SSR translations

## Test Plan
- [ ] Verify language selector in Settings > System > Preferences changes UI language
- [ ] Verify language persists after page reload
- [ ] Verify all translated pages render correctly in both English and pt-BR
- [ ] Verify English remains the default language for new users
- [ ] Verify untranslated pages still render correctly (fallback to English keys)

Closes #486

PR #518, opened 2026-03-24
Added initial i18n framework and most german translations

martinseener:
References #249 and #486 

- Add i18next, react-i18next, i18next-browser-languagedetector packages
- Configure i18n initialization with language detector in lib/i18n.ts
- Created en/de translation files and moved most hard-coded strings into the files and translated them
- Uses locale-aware date formatting where applicable
- Added language-specific Wikipedia content files (wikipedia.en.json, wikipedia.de.json) and updated download URLs
- Added NOMAD_REPO_URL env variable for fork-friendly URL resolution (easier testing and rollout independent of Crosstalk repo)


martinseener:
Additional comment: I was not able to fully test this yet due to time constraints, for example for german wikipedia changes but I did test all the string translations throughout the UI. I probably have missed some as I worked on this for a while now but this can be a good starting point as it allows for easy multi-language support as well, not just my mother tongue that I added.

martinseener:
Screenshot 2026-03-24 at 13 30 22
Screenshot 2026-03-24 at 13 29 10

chriscrosstalk:

martinseener:
@chriscrosstalk I would continue enhancing this PR. Some things are definitely missing and I try to keep it up with the latest versions. So when it's due in the Roadmap, you can easily take it :)

chriscrosstalk:
Thanks for the substantial work on i18n here. Quick status update so you know where this stands:

There are currently three open i18n PRs using two different frameworks (#490 with react-i18next + Portuguese, #518 with react-i18next + German, #648 with plain i18next + Chinese), each covering a different language. Before landing any i18n work, we need to make a framework decision so we end up with one coherent approach rather than a mix of incompatible implementations.

**i18n isn't in scope for the next release cycle.** It's on the roadmap for a future release, but no firm timeline.

**What this means for your PR:**
- No rush to rebase against current `dev` — these are long-lived PRs
- The framework decision will happen when we have capacity to actually integrate i18n, not before
- We'll coordinate a single direction with all three contributors when that time comes
- Your translation work is genuinely valued — we don't want to waste it by forcing a premature decision between branches

Leaving this open and labeling as Roadmap Item. Thanks for your patience and for contributing to NOMAD's international reach.

chriscrosstalk:
Thanks for the contribution and for putting real work into this.

Internationalization is something Jake and I are planning to design and build in-house rather than accept piecewise community submissions. The framework choice, key naming, locale loading, content-language wiring, and downstream translation maintenance all need to be coordinated decisions, and forking one of three competing i18n PRs would leave the other two stranded.

Closing this PR for now. If you'd like to vote/follow the work, please add your support at https://roadmap.projectnomad.us. We'll credit the prior community PRs when the in-house framework lands.

PR #525, opened 2026-03-24
feat(maps): add global map download from Protomaps

bgauger:
## Summary
- Adds a "Download Global Map" option to Maps Manager that downloads the latest full-planet PMTiles build from build.protomaps.com (~125 GB)
- Fetches build metadata from Protomaps to show the latest available build date and size
- Uses resumable downloads so interrupted transfers pick up where they left off

## Test plan
- [x] Tested full download on test server (125.4 GB completed successfully)
- [x] Verified map renders correctly after download
- [x] Verified download resumes after interruption instead of restarting
- [x] Verified UI shows progress in Active Downloads section

### DIFF
--- a/admin/app/controllers/maps_controller.ts
+++ b/admin/app/controllers/maps_controller.ts
@@ -73,6 +73,18 @@ export default class MapsController {
     return await this.mapService.listRegions()
   }
 
+  async globalMapInfo({}: HttpContext) {
+    return await this.mapService.getGlobalMapInfo()
+  }
+
+  async downloadGlobalMap({}: HttpContext) {
+    const result = await this.mapService.downloadGlobalMap()
+    return {
+      message: 'Download started successfully',
+      ...result,
+    }
+  }
+
   async styles({ request, response }: HttpContext) {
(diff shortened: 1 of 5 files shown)


kane9287:
Neat idea, for the purpose how would a global map be beneficial? I live in North American so I downloaded Canada from the Yukon down to Panama before it ferries to Columbia. This is the only realistically feasible travel people can handle in a catastrophic scenario. 

I think it would be better to chunk vectors of the world and make them available like the chunks of the US are. To DL Canada I used this command on Win 11 with pmtiles.exe: 

`pmtiles.exe extract https://build.protomaps.com/20260320.pmtiles canada.pmtiles --bbox=-141,41.5,-52,83.5 --maxzoom=16`

With the help of an LLM it would be easy to adjust the bbox to include any specific region. This would help to minimize total storage requirements allowing the system to stay lean and low power.

chriscrosstalk:
UPDATE:  Staging for v1.31.0

jakeaturner (approved):
Great stuff, thanks!

PR #536, opened 2026-03-25
fix(security): move hardcoded HMAC benchmark secret to environment variable (CWE-798)

andesyteoss:
## Vulnerability Summary

**CWE:** [CWE-798 — Use of Hard-coded Credentials](https://cwe.mitre.org/data/definitions/798.html)
**Severity:** Medium (leaderboard integrity impact; no RCE / data loss)
**Confidence:** High — the same vulnerability was **already exploited in the wild** (see commit `baf16ae`)

### Data Flow

1. `admin/app/services/benchmark_service.ts` defines `BENCHMARK_HMAC_SECRET` as a hardcoded hex string (`778ba65d…`).
2. `BenchmarkService.submitToRepository()` uses the secret to HMAC-sign POST requests to `https://benchmark.projectnomad.us/api/v1/submit`.
3. Because this repository is public, **anyone** can read the secret from source (or git history) and forge signed benchmark submissions.

### Prior Exploitation

Commit `baf16ae` documents that the **previous** hardcoded secret (`nomad-benchmark-v1-2026`) was already extracted and used to submit a fake leaderboard entry. The maintainer rotated the secret but re-hardcoded the replacement, repeating the same mistake.

---

## Fix Description

**3 files changed, 22 insertions, 6 deletions — minimal, targeted diff.**

| File | Change |
|---|---|
| `admin/app/services/benchmark_service.ts` | Replace hardcoded string with `env.get('BENCHMARK_HMAC_SECRET')`. Add a runtime guard that throws a clear error if the variable is not set, preventing unsigned submissions. |
| `admin/start/env.ts` | Register `BENCHMARK_HMAC_SECRET` as an optional string in the AdonisJS env schema. Optional so the app still starts for users who don't use the benchmark feature. |
| `admin/.env.example` | Document the new variable with a generation hint (`openssl rand -hex 24`). |

### Rationale

- **Removes the secret from source code and git history going forward.** The old secret `778ba65d…` should be considered burned and rotated server-side.
- **Follows existing project conventions** — the codebase already uses `env.get()` for `DB_PASSWORD`, `APP_KEY`, etc.
- **Fails safely** — if the env var is missing, submission is refused with a descriptive error rather than silently sending an unsigned request.
- **No breaking changes** — all other benchmark functionality (local scoring, display) is unaffected.

---

## Test Results

| # | Test | Result |
|---|---|---|
| 1 | TypeScript compilation (`node ace build`) | ✅ Compiles cleanly with the new `env.get()` import |
| 2 | Runtime guard — missing env var | ✅ Throws `Error: Benchmark submission signing secret is not configured…` |
| 3 | Runtime guard — env var set | ✅ HMAC is computed and attached to request headers as before |
| 4 | No regressions in env schema | ✅ `BENCHMARK_HMAC_SECRET` is optional; app starts without it |
| 5 | `.env.example` is valid | ✅ Variable is commented out with generation instructions |
| 6 | Diff is minimal (3 files) | ✅ No unrelated changes |

---

## Disprove Analysis

We systematically tried to invalidate this finding before submitting. Summary of the devil's-advocate checks:

### Checks Performed

| Check | Finding |
|---|---|
| **Auth check** | No authentication on any routes — but irrelevant; the secret is in *public source code* |
| **Network check** | CORS `origin: ['*']`, binds `0.0.0.0:8080` — but the secret is on GitHub, not behind a firewall |
| **Deployment check** | Secret is baked into the Docker image at build time — extractable from image or repo |
| **Caller trace** | Secret is used *only* in `submitToRepository()` to sign benchmark POSTs — the HMAC *is* the authentication |
| **Validation check** | No additional auth layer on the benchmark endpoint beyond the HMAC signature |
| **Prior reports** | Issue #211 covers other security items; this specific secret is not listed but is in the same category |
| **Similar vulns** | No other hardcoded secrets in the codebase (`APP_KEY`, `DB_PASSWORD` use `replaceme` placeholders ✅) |

### Counterarguments Considered

1. **"The code comments already acknowledge the weakness"** — Acknowledgment ≠ mitigation. The secret was already exploited despite the comment.
2. **"Shared-secret model is architecturally weak anyway"** — True, but moving to env vars raises the bar from "anyone on the internet" to "only operators given the secret." That is a meaningful improvement.
3. **"Low-impact target (just a leaderboard)"** — The maintainers themselves treated it as worth rotating (commit `baf16ae`). Leaderboard integrity matters to the community.

### Exploit Sketch (for reviewer reference)

```bash
# Read the hardcoded secret from the public GitHub repo
SECRET="778ba65d0bc0e23119e5ffce4b3716648a7d071f0a47ec3f"
TIMESTAMP=$(date +%s)000
BODY='{"nomad_score":100,"builder_tag":"forged"}'
SIG=$(echo -n "${TIMESTAMP}${BODY}" | openssl dgst -sha256 -hmac "$SECRET" | awk '{print $2}')
curl -X POST https://benchmark.projectnomad.us/api/v1/submit \
  -H "Content-Type: application/json" \
  -H "X-NOMAD-Timestamp: $TIMESTAMP" \
  -H "X-NOMAD-Signature: $SIG" \
  -d "$BODY"
```

This is essentially what already happened per commit `baf16ae`.

**Verdict: CONFIRMED_VALID** — the finding is real, was previously exploited, and this fix follows project conventions.

---

## Recommended Follow-up

1. **Rotate the secret server-side** — the value `778ba65d…` remains in git history and must be considered compromised.
2. **(Optional)** Add `BENCHMARK_HMAC_SECRET` to `management_compose.yaml` so Docker Compose deployments get a placeholder.
3. **(Future)** Consider challenge-response or per-instance token authentication for stronger leaderboard protection.

---

### DIFF
--- a/admin/.env.example
+++ b/admin/.env.example
@@ -16,3 +16,6 @@ REDIS_PORT=6379
 # On Windows dev, use an absolute path like: C:/nomad-storage
 # On Linux production, use: /opt/project-nomad/storage
 NOMAD_STORAGE_PATH=/opt/project-nomad/storage
+# HMAC secret used to sign benchmark submissions to benchmark.projectnomad.us
+# Generate with: openssl rand -hex 24
+# BENCHMARK_HMAC_SECRET=
--- a/admin/app/services/benchmark_service.ts
+++ b/admin/app/services/benchmark_service.ts
@@ -23,16 +23,17 @@ import type {
   RepositoryStats,
 } from '../../types/benchmark.js'
 import { randomUUID, createHmac } from 'node:crypto'
+import env from '#start/env'
 import { DockerService } from './docker_service.js'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
 import { BROADCAST_CHANNELS } from '../../constants/broadcast.js'
 import Dockerode from 'dockerode'
 
-// HMAC secret for signing submissions to the benchmark repository
-// This provides basic protection against casual API abuse.
-// Note: Since NOMAD is open source, a determined attacker could extract this.
-// For stronger protection, see challenge-response authentication.
-const BENCHMARK_HMAC_SECRET = '778ba65d0bc0e23119e5ffce4b3716648a7d071f0a47ec3f'
+// HMAC secret for signing submissions to the benchmark repository.
+// Must be provided via the BENCHMARK_HMAC_SECRET environment variable.
+// The benchmark server uses this to verify that submissions originate from
+// a genuine NOMAD instance.  Never commit the real secret to source control.
+const BENCHMARK_HMAC_SECRET = env.get('BENCHMARK_HMAC_SECRET')
 
 // Re-export default weights for use in service
 const SCORE_WEIGHTS = {
@@ -157,6 +158,11 @@ export class BenchmarkService {
     }
 
     try {
+      // Refuse to submit if the signing secret is not configured
+      if (!BENCHMARK_HMAC_SECRET) {
+        throw new Error('Benchmark submission signing secret is not configured. Set the BENCHMARK_HMAC_SECRET environment variable.')
+      }
+
       // Generate HMAC signature for submission verification
       const timestamp = Date.now().toString()
       const payload = timestamp + JSON.stringify(submission)
--- a/admin/start/env.ts
+++ b/admin/start/env.ts
@@ -60,4 +60,11 @@ export default await Env.create(new URL('../', import.meta.url), {
   |----------------------------------------------------------
   */
   NOMAD_API_URL: Env.schema.string.optional(),
+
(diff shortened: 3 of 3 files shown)


Kirderf420:
Really well work and documentation.

chriscrosstalk:
Good security hygiene, but we've already addressed this. The HMAC secret was rotated (PR #238) and the leaderboard repo was made private. Moving to an env var is the right long-term approach, but the install script would also need to generate and inject the secret — otherwise fresh installs would have no HMAC key and benchmark submissions would silently fail.

Closing this one since the immediate vulnerability is already mitigated. If we revisit the HMAC architecture later, we'll factor env vars into the design.

PR #553, opened 2026-03-26
feat(ollama): expand offline fallback models and add API transparency

### LINKED ISSUE #528: [Bug]: External URL for API calls
### Issue Category

AI Assistant (Ollama)

### Bug Description

When combing the code for ways to improve and personalize I came across an ENV inside of admin>constants>misc.ts that points to an external API.

`export const NOMAD_API_DEFAULT_BASE_URL = 'https://api.projectnomad.us'`

I wasn't sure what the API was for so I removed it to test. When I tested the custom image I noticed that the Ollama model list was reduced to 3 models. I ran this issue through Claude and learned that the https://api.projectnomad.us is called to generate a list of models for Ollama and when that URL isn't there it _silently_ fails and returns a truncated list.

Here's the rub: it shouldn't silently fail, it should tell us it failed to call the domain. We should also have disclosure that this calls out to a domain and we have not alternative resource. This goes against the privacy first mantra since everytime the AI settings is loaded it pings Project NOMADs domain. This may include sensitive data like IP addresses, headers, etc.

Further this could have been handled with an internal list, even a local URL redirect to an API embedded in the software. This may have been an oversight but I would encourage quick correction, things like this don't go over well in open source communities and we've seen a lot of projects struggle with PR lately.

### Steps to Reproduce

1. Remove the misc.ts from the repo
2. Rebuild the docker image as a custom
3. Alter compose.yml in /opt/project-nomad to the custom image above
4. Compose up, compose down
5. Go to AI settings, only 3 modals are available to download 

### Expected Behavior

There should be an internal list that the software references. Alternatively there should be a localhost URL with the list at a reachable API. Even better, offer a way to customize this so people can easily download models of their preference. 

Just don't hide an API call in a random misc file. Not cool.

### Actual Behavior

Removing a remote API URL from a file caused models to not be available for download in the settings

### N.O.M.A.D. Version

1.30.2

### Operating System

Ubuntu 24.04

### Docker Version

_No response_

No

### GPU Model (if applicable)

_No response_

### System Specifications

_No response_

_No response_

### Relevant Logs

```shell

```

```javascript

```

### Screenshots

Image

### Additional Context

_No response_

### Pre-submission Checklist

LuisMIguelFurlanettoSousa:
## Summary

Addresses the privacy and transparency concerns raised in #528 regarding the external API dependency on `api.projectnomad.us`:

- **Expanded fallback model list:** Increased from 3 to 15 popular models covering general text, code, reasoning, embedding, multimodal, and small/embedded categories. Users now have a meaningful selection even when fully offline.

- **Explicit fallback indicator:** Added a `source` field (`'api'` | `'fallback'`) to the model list API response. The frontend displays an informational banner when using the offline list, and the refresh button shows a warning notification instead of a false success.

- **Fixed hardcoded URL:** `system_service.ts` was hardcoding `https://api.projectnomad.us` instead of using the shared `NOMAD_API_DEFAULT_BASE_URL` constant and `NOMAD_API_URL` env var override.

- **Sanitized error messages:** Error responses in the release notes subscription no longer leak internal details to the client.

### Files Changed

| File | Change |
|------|--------|
| `constants/ollama.ts` | Expand fallback from 3 → 15 models |
| `services/ollama_service.ts` | Add `source` field to all return paths |
| `inertia/lib/api.ts` | Add `source` to response type |
| `inertia/pages/settings/models.tsx` | Show fallback Alert + context-aware refresh notification |
| `services/system_service.ts` | Use shared URL constant + sanitize error |

## Related Issues

Fixes #528

## Test Plan

- [ ] Block `api.projectnomad.us` (e.g., via `/etc/hosts`) and verify 15 models appear with info banner
- [ ] Unblock and force refresh — verify banner disappears and success notification shows
- [ ] Verify `NOMAD_API_URL` env var overrides the default URL for release notes subscription
- [ ] Verify release notes subscription errors show generic message, not internal details

### DIFF
--- a/admin/app/services/system_service.ts
+++ b/admin/app/services/system_service.ts
@@ -14,6 +14,7 @@ import env from '#start/env'
 import KVStore from '#models/kv_store'
 import { KV_STORE_SCHEMA, KVStoreKey } from '../../types/kv_store.js'
 import { isNewerVersion } from '../utils/version.js'
+import { NOMAD_API_DEFAULT_BASE_URL } from '../../constants/misc.js'
 
 
 @inject()
@@ -384,8 +385,9 @@ export class SystemService {
 
   async subscribeToReleaseNotes(email: string): Promise<{ success: boolean; message: string }> {
     try {
+      const baseUrl = env.get('NOMAD_API_URL') || NOMAD_API_DEFAULT_BASE_URL
       const response = await axios.post(
-        'https://api.projectnomad.us/api/v1/lists/release-notes/subscribe',
+        `${baseUrl}/api/v1/lists/release-notes/subscribe`,
         { email },
         { timeout: 5000 }
       )
@@ -399,13 +401,13 @@ export class SystemService {
 
       return {
         success: false,
-        message: `Failed to subscribe: ${response.statusText}`,
+        message: 'Failed to subscribe to release notes. Please try again later.',
       }
     } catch (error) {
-      logger.error('Error subscribing to release notes:', error)
+      logger.error({ err: error }, '[SystemService] Error subscribing to release notes')
       return {
         success: false,
-        message: `Failed to subscribe: ${error instanceof Error ? error.message : error}`,
+        message: 'Failed to subscribe to release notes. The service may be temporarily unavailable.',
       }
     }
   }


chriscrosstalk:
Thanks for the work here, but we don't think expanding the fallback list adds much value. The fallback kicks in when `api.projectnomad.us` is unreachable — which in practice means NOMAD is offline. If NOMAD is offline, you can't download any of these models regardless of how many are in the list.

The only scenario where a bigger list helps is if our API is down but the user still has internet — a pretty narrow edge case that doesn't justify +303 lines of code.

Closing this one, but appreciate the contribution.

PR #556, opened 2026-03-26
feat: add MiniMax as optional cloud LLM provider

octo-patch:
## Summary

Adds [MiniMax](https://platform.minimax.io) as an optional cloud LLM provider alongside the existing Ollama local models. When `MINIMAX_API_KEY` is configured, MiniMax cloud models (MiniMax-M2.7 and MiniMax-M2.7-highspeed) appear in the model selector, giving users access to powerful cloud AI when internet is available while preserving the offline-first Ollama experience.

## Changes

- **`admin/app/services/minimax_service.ts`** — New service implementing MiniMax OpenAI-compatible chat API with streaming support
- **`admin/app/controllers/ollama_controller.ts`** — Route MiniMax models to MiniMax service; include cloud models in installed models list
- **`admin/start/env.ts`** — Add optional `MINIMAX_API_KEY` environment variable
- **`admin/.env.example`** — Document MiniMax API key configuration
- **`admin/inertia/components/chat/index.tsx`** — Show "Cloud" label for cloud models instead of file size
- **`README.md`** — Mention MiniMax alongside Ollama in capabilities list

## How It Works

- When `MINIMAX_API_KEY` is set, MiniMax models appear in the model dropdown marked as "Cloud"
- The chat controller detects MiniMax model names (prefix `MiniMax-`) and routes requests to the MiniMax service
- MiniMax service uses the OpenAI-compatible API at `https://api.minimax.io/v1` with native fetch (no new dependencies)
- Both streaming and non-streaming chat work identically to Ollama models
- Without the API key, behavior is unchanged — purely offline Ollama

## Models

| Model | Description |
|-------|-------------|
| `MiniMax-M2.7` | Peak Performance. Ultimate Value. Master the Complex |
| `MiniMax-M2.7-highspeed` | Same performance, faster and more agile |

## API Reference

- [MiniMax Chat API (OpenAI Compatible)](https://platform.minimax.io/docs/api-reference/text-openai-api)

## Test Plan

- [x] 8 unit tests for MiniMaxService (model detection, config, error handling)
- [x] 3 integration tests (basic chat, streaming, system messages)
- [x] All tests pass with `node ace test`
- [ ] Verify MiniMax models appear in dropdown when API key is set
- [ ] Verify chat works with MiniMax-M2.7 model
- [ ] Verify streaming output works correctly
- [ ] Verify Ollama models continue working normally

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ For more control over the installation process, copy and paste the [Docker Compo
 N.O.M.A.D. is a management UI ("Command Center") and API that orchestrates a collection of containerized tools and resources via [Docker](https://www.docker.com/). It handles installation, configuration, and updates for everything — so you don't have to.
 
 **Built-in capabilities include:**
-- **AI Chat with Knowledge Base** — local AI chat powered by [Ollama](https://ollama.com/), with document upload and semantic search (RAG via [Qdrant](https://qdrant.tech/))
+- **AI Chat with Knowledge Base** — local AI chat powered by [Ollama](https://ollama.com/), with optional cloud models via [MiniMax](https://platform.minimax.io), document upload and semantic search (RAG via [Qdrant](https://qdrant.tech/))
 - **Information Library** — offline Wikipedia, medical references, ebooks, and more via [Kiwix](https://kiwix.org/)
 - **Education Platform** — Khan Academy courses with progress tracking via [Kolibri](https://learningequality.org/kolibri/)
 - **Offline Maps** — downloadable regional maps via [ProtoMaps](https://protomaps.com)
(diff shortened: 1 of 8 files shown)


jakeaturner:
Thank you for opening this PR! I am not familiar with MiniMax myself, but because it's OpenAI compatible, I suspect this may tee'd up for support by #244 which enables remote Ollama and other OpenAI-compatible remotes. @hestela or @octo-patch might either of you know if this is the case?

hestela:
In theory it should work ok after 244 but we'd need to do a quick follow on PR to add ability to set tokens in the header for all the AI related requests made. It may also need more tweaks to context length. Im not already paying for api credits anywhere so not sure who would even test this. I can test auth because LM studio supports setting auth in the openai way I believe.

But nomad wasn't originally designed to connect to paid AI providers so I wouldn't he surprised if there's some suboptimal prompts or settings there which would waste more API cost.

chriscrosstalk:
Thanks for the effort here, but this conflicts with NOMAD's core mission — offline-first, no cloud dependencies. Adding a cloud LLM provider goes against what the project is about. Users who want cloud AI already have plenty of options; NOMAD's value is that everything runs locally.

Closing this one, but appreciate the contribution.

PR #559, opened 2026-03-26
fix: add DNS delay to allow dhcpcd resolution on Debian (#557)

### LINKED ISSUE #557: [Bug]: 500 Error on reboot due to Docker DNS race condition with dhcpcd on Debian
### Issue Category

Docker/Container Issues

### Bug Description

On Debian systems, Docker starts before dhcpcd finishes writing nameservers to /etc/resolv.conf, causing containers to launch with no upstream DNS. Any feature requiring external DNS resolution returns a 500 error after every reboot.

### Steps to Reproduce

- Install Project N.O.M.A.D. on a Debian system using dhcpcd
- Reboot the VM
- Navigate to any page requiring external network access

### Expected Behavior

Project N.O.M.A.D. loads correctly after reboot as it does after initial installation.

### Actual Behavior

500 error on pages requiring external DNS. Inspecting the container confirms no upstream nameservers:
```
docker exec nomad_admin cat /etc/resolv.conf
# NO EXTERNAL NAMESERVERS DEFINED
```

Error: `getaddrinfo EAI_AGAIN browse.library.kiwix.org`

### N.O.M.A.D. Version

1.30.2

### Operating System

Debian 13 (Trixie)

### Docker Version

_No response_

No

### GPU Model (if applicable)

_No response_

### System Specifications

_No response_

_No response_

### Relevant Logs

```shell

```

```javascript

```

### Screenshots

Image

### Additional Context

Workaround: create `/etc/docker/daemon.json` with explicit DNS servers:
```
json{
  "dns": ["8.8.8.8", "8.8.4.4"]
}
```
Then `sudo systemctl restart docker`. 

The believed root cause is a race condition on boot where Docker starts and captures DNS configuration before dhcpcd has finished writing nameservers to /etc/resolv.conf. On this Debian system, dhcpcd does not appear to block network-online.target, meaning Docker cannot reliably depend on DNS being ready when it starts. 

Note: the exact mechanism by which dhcpcd is launched on this system could not be fully determined during debugging, so there may be additional factors at play.

Running the update script appears to resolve the issue temporarily but only because containers are recreated long after boot — it does not address the root cause.

### Pre-submission Checklist

- [ ] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

sjhddh:
Fixes #557. On Debian systems with dhcpcd, Docker containers can start before nameservers are written to `/etc/resolv.conf`, causing 500 errors. Added a brief delay loop to ensure DNS is ready before startup.\n\n*(Refactored using the open-source [0-editor](https://github.com/0-protocol/0-editor))*\n\n

### DIFF
--- a/install/entrypoint.sh
+++ b/install/entrypoint.sh
@@ -21,4 +21,7 @@ node ace queue:work --all &
 
 # Start the AdonisJS application
 echo "Starting AdonisJS application..."
+    # Wait for DNS resolution (Debian dhcpcd race condition)
+    until ping -c1 8.8.8.8 >/dev/null 2>&1 || ping -c1 google.com >/dev/null 2>&1 || [ $SECONDS -gt 10 ]; do sleep 1; done
+
 exec node bin/server.js


systemofapwne:
If I understand #557 correctly, the docker deamon starts before dhcpcd has fully set up name resolution on the system. 

I am not so sure, if this PR will fix #557, as it looks like it just will make the container wait for DNS to work. But a container and/or docker-deamon restart still might be required after dhcpcd has set up resolv.conf: E.g. I have just edited my /etc/resolv.conf with an invalid config (nameserver 0.0.0.0) on my docker host where an example container was still running. DNS resolution remained working until I restarted the deaemon and container.

Further more, the approach here tries to cure a symptom (for nomad only) rather than fixing the problem (for all containers)

IMHO, the user should simply override his docker.service unit file to start after dhcpcd
```
# /etc/systemd/system/docker.service.d/wait-for-dhcpcd.conf
[Unit]
After=dhcpcd.service
```
TBH, I did not test this configuration, but in principle it will ask docker.service to start after dhcpcd. (All other default `After` instructions of the original service still hold, as this dropin unit will only append to `After`, not override it)

Alternatively, one could add some additional wait time to give dhcpcd more time to set things up.
```
# /etc/systemd/system/docker.service.d/wait-for-dhcpcd.conf
[Service]
ExecStartPre=sleep 15
```

Ideally, combine both approaches

sjhddh:
Thank you for the detailed feedback and the `systemd` suggestions! You are absolutely right that modifying `docker.service` addresses the root timing issue at the host level for all containers.

However, for Project N.O.M.A.D., we prefer to keep the resilience localized to the container itself. Modifying the host's global Docker systemd service is a bit intrusive and can behave inconsistently across different Linux distributions or custom user setups. By handling the wait loop inside `entrypoint.sh`, the container guarantees its own startup prerequisites without altering the host's global daemon configuration. 

We will stick with the localized container-level delay for now to keep the application footprint as isolated and robust as possible.

chriscrosstalk:
UPDATE:  Appreciate the effort, but we're gonna close this one out.  Super edge case, and there's a decent workaround for those who are affected.

Issue #502, opened 2026-03-23
[Bug]: Fresh install - Kiwix failure


YEMX0621:
### Issue Category

Service Management (Start/Stop/Update)

### Bug Description

Project NOMAD Debug Info
========================
App Version: 1.30.1
Environment: production

System:
  OS: Ubuntu 24.04.4 LTS
  Hostname: nomadnode1
  Kernel: 6.8.0-106-generic
  Architecture: x64
  Uptime: 2h 18m

Hardware:
  CPU: Xeon® E7- 4870 (80 cores)
  RAM: 251.8 GB total, 4.9 GB used, 247 GB available
  GPU: None detected
  Disk: 40.3 TB, 0% used

Installed Services:
  Information Library (nomad_kiwix_server): restarting
  Education Platform (nomad_kolibri): running
  AI Assistant (nomad_ollama): running
  Notes (nomad_flatnotes): running
  Data Tools (nomad_cyberchef): running

Internet Status: Online
Update Available: No (1.30.1 is latest)
Browser: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36 Edg/146.0.0.0

Upon initial install to fresh linux host I ran the following commands:
- sudo apt update -y && sudo apt full-upgrade -y && sudo apt autoremove -y && sudo apt clean -y && sudo apt autoclean -y
- sudo apt install -y htop git curl wget unzip net-tools iftop nano
- sudo curl -fsSL https://raw.githubusercontent.com/Crosstalk-Solutions/project-nomad/refs/heads/main/install/install_nomad.sh -o install_nomad.sh && sudo bash install_nomad.sh
- sudo reboot now

upon reboot I log into the server page at http://<local_IP>:8080 and start the "easy setup" step.
- I select everything as far as apps go to be installed
- I select the data I wish to have installed (everything)
- After the apps are installed I run the update check and install the updates identified
-  - No idea why a brand new fresh install does not pull app updates on initial install?
-  - Dozzle was hardcoded to version 10.0 so I changed it in the yml file to pull latest.
- The system begins to process the request but fails.

### Steps to Reproduce

nomad_admin@nomadnode1:~$ sudo docker logs --tail 50 nomad_kiwix_server
[sudo] password for nomad_admin:
/data/nhs.uk_en_medicines_2025-12.zim
[...]

Issue #499, opened 2026-03-23
[Bug]: kiwix_server is not starting after updating to version 3.8.2


deepshaswat:
### Issue Category

Service Management (Start/Stop/Update)

### Bug Description

Kiwix_server was updated to 3.8.2 using the check for updates, after that the server is not starting and is on crashback loop

### Steps to Reproduce

1. go to settings
2. check for updates
3. install the latest 3.8.2 for kiwix_server

### Expected Behavior

server should start automatically

### Actual Behavior

server is failing to start and is in loop of restart

### N.O.M.A.D. Version

1.30.1

### Operating System

Ubuntu 24.04

### Docker Version

29.3.0

### Do you have a dedicated GPU?

No

### Relevant Logs

```shell
[...]

Issue #464, opened 2026-03-22
[Bug]: Information library :8090 not working. Unable to add the ZIM file '*.zim'


mikilbc:
### Issue Category

Docker/Container Issues

### Bug Description

Nomad internal docker nomad_kiwix_server restarts in loop. Zim files are in storage/zim/ but not listed.
From container log: "/usr/local/bin/kiwix-serve --port=8080 *.zim --address=all
Unable to add the ZIM file '*.zim' to the internal library.
Here is the content of /data:
"

### Steps to Reproduce

Fresh installation, Easy setup, Information library.

### Expected Behavior

I want to see Information library content.

### Actual Behavior

Firefox cannot connect to 192.168.50.3:8090, tested on localhost:8090 on server too.

### N.O.M.A.D. Version

1.30.1

### Operating System

Other (not yet officially supported)

### Docker Version

29.2.1

### Do you have a dedicated GPU?

No

### Relevant Logs

```shell
docker logs nomad_kiwix_server --tail=50
Unable to add the ZIM file '*.zim' to the internal library.
Here is the content of /data:
/usr/local/bin/kiwix-serve --port=8080 *.zim --address=all
Unable to add the ZIM file '*.zim' to the internal library.
Here is the content of /data:
/usr/local/bin/kiwix-serve --port=8080 *.zim --address=all
Unable to add the ZIM file '*.zim' to the internal library.
Here is the content of /data:
```

### Browser Console Errors (if UI issue)

```javascript

```

### Pre-submission Checklist

- [x] I have searched for existing issues that might be related to this bug
- [x] I am running the latest version of Project N.O.M.A.D. (or have noted my version above)
- [x] I have redacted any personal or sensitive information from logs and screenshots
- [ ] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

SxthGear:
Same issue here.

kennethbrewer3:
I ran into this issue as well because my installation is not in the expected place. My system is set up in a way that the partition that has /opt is too small for the zim files. I found that I can fix the issue by doing the following. This creates a symbolic link that tricks nomad into trying to get the zim files in the folder where you have them. Replace my path with your installation path.


`sudo ln -s /home/kbb3/docker/project-nomad /opt/project-nomad`

This isn't an official fix. I'm not on the development team. It's kind of a hack that has worked for me. I have read another issue where they intend to eventually push an update that has an official way to alter the path.

[...]

PR #562, opened 2026-03-26
fix(kiwix): validate ZIM files after download to prevent restart loops

bgauger:
## Problem

kiwix-serve exits immediately when it encounters a ZIM file it cannot load (`Unable to add the ZIM file to the internal library`). Since the container has `restart: unless-stopped`, a single bad file puts the entire Information Library into an infinite restart loop.

This affects fully downloaded files that are corrupt or incompatible with the current kiwix-serve version — distinct from partial downloads (addressed by #448). Users in #499 reported having to manually identify and delete bad files one at a time via Dozzle logs, sometimes repeatedly after re-downloading.

Ref #499, #502, #464

## Fix

Adds a `validateZimFile()` method to `ZimService` that spins up a temporary kiwix-serve container with just the single file. If kiwix-serve stays running for 5 seconds, the file is valid. If it exits immediately with "Unable to add", the file is invalid and gets deleted before the main kiwix container is restarted.

This validation runs in `downloadRemoteSuccessCallback()` after each download completes, before the kiwix container restart.

## Relationship to #448

PR #448 prevents kiwix from seeing **partial** downloads by staging to `.tmp`. This PR validates **complete** files that kiwix still can't load (corruption, version incompatibility, etc.). They cover different failure modes and are complementary.

## Changes

Single file modified: `admin/app/services/zim_service.ts`

- Add `validateZimFile(filename)` — creates a temporary kiwix-serve container to test a single ZIM file, cleans up after
- Modify `downloadRemoteSuccessCallback()` — validate each newly downloaded file; delete invalid files and skip their `InstalledResource` entries

## Testing

Tested on a live NOMAD instance (v1.30.3) that was experiencing this exact crash loop with 8 bad ZIM files:

| Test | File | Expected | Result |
|------|------|----------|--------|
| Valid file (devdocs_en_bash) | Small, known good | VALID | VALID |
| Invalid file (librepathology) | Full size, kiwix can't load | INVALID | INVALID |
| Non-existent file | Missing from disk | INVALID | INVALID |
| Valid file (wikipedia_en_100_mini) | Larger, known good | VALID | VALID |

### DIFF
--- a/admin/app/services/zim_service.ts
+++ b/admin/app/services/zim_service.ts
@@ -26,6 +26,7 @@ import { RunDownloadJob } from '#jobs/run_download_job'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
 import { CollectionManifestService } from './collection_manifest_service.js'
 import type { CategoryWithStatus } from '../../types/collections.js'
+import Service from '#models/service'
 
 const ZIM_MIME_TYPES = ['application/x-zim', 'application/x-openzim', 'application/octet-stream']
 const WIKIPEDIA_OPTIONS_URL = 'https://raw.githubusercontent.com/Crosstalk-Solutions/project-nomad/refs/heads/main/collections/wikipedia.json'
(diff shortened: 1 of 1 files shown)


artbird309:
I tested `zimcheck` against 71 ZIM files (Proxmox LXC, 40 vCPUs, Debian Trixie, zimcheck 3.5.0, storage on RAIDZ1 of 3x 10K SAS drives) while reviewing a related suggestion on #448 - sharing the data here as background context since it covers the same problem space, though my testing used zimcheck directly rather than the kiwix-serve validation approach this PR takes.

**`-I` (integrity):** 36 of 71 files failed, all with the identical error: "Full Title index table outside (or not fully inside) ZIM file." The failure pattern lines up with ZIM age - older files pass, newer ones fail. For example, `askubuntu.com_en_all_2024-03` passes while `askubuntu.com_en_all_2025-12` fails. The same pattern held across unrelated sources: devdocs, libretexts, zimgit, and older stackexchange files all pass; wikipedia, wikibooks, freecodecamp, gutenberg, and newer stackexchange files all fail. I spot-checked a couple of the failing files in Kiwix (3dprinting and freecodecamp) and they loaded, though I did not verify all 36. This looks like a false positive in zimcheck 3.5.0 against the newer ZIM title index format rather than genuine corruption. Using `-I` as a hard gate would reject valid files.

**`-C` (checksum):** All 20 files tested passed. Timing varied significantly by file size - small files finished in under a minute, but `wikipedia_en_all_maxi_2026-02` took **8 minutes 17 seconds** on this 40-vCPU server. Sequential read benchmark on this storage was ~728 MiB/s (ZFS ARC assisted), write ~79 MiB/s. On lower-spec NOMAD target hardware the checksum time would be considerably longer.

Full data and context: https://github.com/Crosstalk-Solutions/project-nomad/pull/448#issuecomment-4148751215

*Data reviewed with assistance from Claude Sonnet 4.6 (Anthropic).*

bgauger:
Great data, thanks for running this. The zimcheck `-I` false positives on newer files are exactly why this PR uses kiwix-serve directly as the validator — it tests what actually matters (can kiwix-serve load the file) without getting tripped up by format changes between ZIM versions.

The checksum timing is also worth noting. 8+ minutes for the full Wikipedia on a 40-vCPU server rules out `-C` as a validation gate on target NOMAD hardware.

One thing your timing data surfaces — the BullMQ download worker uses the default `lockDuration` of 30 seconds. This validation runs inside the job's `onComplete` callback (before `updateProgress(100)`), so on slower hardware the container spin-up + 5s wait could push past that lock window and trigger a stall. That's a separate queue config issue though, not specific to this PR. I'll open an issue for the `lockDuration`.

chriscrosstalk:

PR #575, opened 2026-03-27
feat(web): Switch all png except favicon to webp format

### LINKED ISSUE #561: [Feature]: Optimize the nomad png logo size
### Feature Category

Performance Optimization

### Problem Statement

It would be great if we could optimize the size of the nomad logo that gets used in a bunch of places. At 952kb it is large in size, and when the pages refresh it seems to lag behind the page in some cases, I suspect due to it being really high quality, and used often

### Proposed Solution

You could run it through one of many open source image ompitimizers it should cut it in half. I did it on my own fork, but didn't want to make a PR without bringing it up here. 

Logo file is here:  
https://github.com/Crosstalk-Solutions/project-nomad/blob/main/admin/public/project_nomad_logo.png

Ideally run the above one through an optimizer, it can then be normal size, and if you want to retain the original you could hold it in the repo for any other use in the future, and minimize code changes. 

### Alternative Solutions

In my own installation I replaced the file with one that is ~200kb but it could be further optimized still. 

### Use Case

As a nomad consumer, I want to shrink the nomad png logo so that it uses less resources and bandwidth on each page load in the app. It is almost 1mb in size, and seems excessive for the logo in the corner of the screen, or even on a page at 200-300px in size. 

All Users

Low - Minor convenience

Run the png through a tool like ImageOptim and tweak settings to get it down to a bit smaller of a size if possible. 

### Examples or References

Image

Image

Yes - I can help test

### Additional Context

Less code, more a new image being uploaded. I can test it locally and offer some versions that are different sizes. All the other png files look pretty well optimized. 

### Pre-submission Checklist

hestela:
Fixes #561 
Before:
```
git ls-files | grep png | xargs -I{} du -b "{}" | awk '{sum+=$1} END {print sum/1024/1024 " MB"}'
3.29008 MB
```
After:
```
git ls-files | grep -E 'webp|png' | xargs -I{} du -b "{}" | awk '{sum+=$1} END {print sum/1024/1024 " MB"}'
0.995811 MB
```
So saving about 2.3MB.
command used to convert:
`git ls-files | grep png |grep -v favicon | xargs -I{} mogrify -format webp "{}"`
I skipped favicon as I wasn't completely sure if favicons support webp. But otherwise all modern browsers support webp. Github supports it as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <div align="center">
-<img src="https://raw.githubusercontent.com/Crosstalk-Solutions/project-nomad/refs/heads/main/admin/public/project_nomad_logo.webp" width="200" height="200"/>
+<img src="admin/public/project_nomad_logo.webp" width="200" height="200"/>
 
 # Project N.O.M.A.D.
 ### Node for Offline Media, Archives, and Data


hestela:
Also wasn't sure if i was supposed to send to dev or main. but there's no conflicts on either branch.

chriscrosstalk:
UPDATE:  Staging for v1.31.0

jakeaturner (approved):
Great stuff, thanks!

Issue #576, opened 2026-03-28
[Feature]: Replaceable mirrors


freimer:
### Feature Category

Content Management (ZIM, Maps, Collections)

### Problem Statement

Build in download source for ZIM files is very slow.  There are other mirrors available that are significantly faster.  Locations could also be blocked for some users.  There should be the ability to override download sources.

### Proposed Solution

I already have a working solution.  I will submit the PR.  I built the new container and it works fine.  Was actually able to download things, including the largest Wikipedia file.

### Use Case

As a user I want to be able to specify my own mirror overrides for any download locations.

### Who would benefit from this feature?

All Users

### How important is this feature to you?

High - Would significantly improve my experience

### Implementation Ideas (Optional)

See PR I will submit.

### Would you be willing to help implement this?

Yes - I can write the code

### Pre-submission Checklist

- [x] I have searched for existing feature requests that might be similar
- [x] This feature aligns with N.O.M.A.D.'s mission of offline-first knowledge and education
- [x] I understand that feature requests are not guaranteed to be implemented

chriscrosstalk:
PR #577 is open for this. Closing the issue -- tracking via the PR.

PR #577, opened 2026-03-28
add configurable download mirrors

freimer:
See #576 .

This implements the ability to replace URL prefixes for mirrors.  As simple as adding this to the compose:

```
      # Optional: rewrite one or more download URL prefixes to mirrors.
      # Example:
      # - DOWNLOAD_MIRROR_RULES={"https://download.kiwix.org/zim/":"https://cdimage.debian.org/mirror/kiwix.org/zim/"}
      - DOWNLOAD_MIRROR_RULES=
```

That is what I used, after I rebuilt the container image and changed the compose to use it, and it works awesome.  Use if you want.

### DIFF
--- a/admin/.env.example
+++ b/admin/.env.example
@@ -16,3 +16,7 @@ REDIS_PORT=6379
 # On Windows dev, use an absolute path like: C:/nomad-storage
 # On Linux production, use: /opt/project-nomad/storage
 NOMAD_STORAGE_PATH=/opt/project-nomad/storage
+# Optional: rewrite download URLs to one or more mirrors using JSON source-prefix => target-prefix mappings.
+# Example for Kiwix:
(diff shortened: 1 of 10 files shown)


chriscrosstalk:
Thanks for identifying this pain point -- slow downloads from the default Kiwix CDN is a real issue. We took your idea and ran with it in a slightly different direction in PR #593: instead of a compose env var, we built a GUI-based library browser so users can switch between mirrors without touching any config files.

PR #593 includes:
- 5 pre-seeded official Kiwix mirrors (US, Germany, Denmark, UK, Global CDN)
- A browsable directory structure so users can explore any mirror's ZIM catalog
- Users can also add their own custom libraries through the UI

Closing this in favor of #593, but appreciate you raising the issue and putting together a working solution -- it helped shape the final approach.

freimer:
Awesome, thank you!

PR #554, opened 2026-03-26
feat(downloads): rich progress, friendly names, cancel, and live status

chriscrosstalk:
## Summary

- **Rich progress data**: Downloads now report `downloadedBytes`/`totalBytes` via BullMQ progress objects, showing "2.3 GB / 5.1 GB" instead of the old "78% / 100%"
- **Friendly names**: Title metadata flows through from curated categories, Content Explorer library, Wikipedia selector, and map collections
- **Cancel button**: Users can cancel active or queued downloads. Uses Redis-based cross-process signaling to abort the HTTP stream in the worker, then cleans up partial files. Confirmation step prevents accidental cancels.
- **Live status indicator**: Green pulsing dot with smoothed transfer speed (e.g., "Downloading... 12.7 MB/s"), orange stall warning after 60s of no data, gray dot for queued items

Backward compatible -- in-flight jobs with integer-only progress fall back to percentage display.

## Files changed (10)

| File | Change |
|------|--------|
| `types/downloads.ts` | New `DownloadProgressData` type, extended job params and response |
| `jobs/run_download_job.ts` | Progress object, AbortController + Redis cancel signal check |
| `services/download_service.ts` | Parse progress object, `cancelJob()` with Redis signal + file cleanup |
| `controllers/downloads_controller.ts` | `cancelJob` action |
| `controllers/zim_controller.ts` | Pass metadata to `downloadRemote()` |
| `services/zim_service.ts` | Wire title/totalBytes through all dispatch paths |
| `services/map_service.ts` | Wire title through collection dispatch |
| `start/routes.ts` | `POST /api/downloads/jobs/:jobId/cancel` |
| `inertia/lib/api.ts` | `cancelDownloadJob()` client method |
| `inertia/components/ActiveDownloads.tsx` | Full UI redesign |

## Test plan

- [x] Start curated tier download -- friendly names, byte progress, green dot + speed all visible
- [x] Cancel a queued (not yet active) download -- removed immediately
- [x] Cancel an active download -- Redis signal aborts stream, partial file deleted
- [x] Backward compat: in-flight jobs from before upgrade show percentage fallback
- [ ] Start download from Content Explorer library -- title passes through
- [ ] Check Easy Setup wizard step 4 -- active downloads render correctly

Tested on NOMAD1 (192.168.200.199) running v1.30.3 with hot-patched container.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #599, opened 2026-03-30
fix: handle locked BullMQ jobs when dismissing failed downloads

### LINKED ISSUE #598: 500 error when dismissing failed download alerts
## Description

Clicking the X button to dismiss a failed download on the Remote Explorer page (`/settings/zim/remote-explorer`) returns a 500 Internal Server Error.

## Steps to Reproduce

1. Trigger a download that will fail (e.g., a large ZIM file that gets aborted due to timeout or DNS failure)
2. Wait for the download to show as "failed" in the UI with the X dismiss button
3. Click the X to remove the failed download alert
4. Observe: `An internal error occurred. Please try again or check the console for details. Request failed with status code 500`

## Root Cause

The issue is in `DownloadService.removeFailedJob()` (`admin/app/services/download_service.ts:54-63`). It calls `job.remove()` on BullMQ jobs that are still locked by the worker.

The chain of events:

1. A download fails (aborted, DNS timeout, etc.)
2. BullMQ's retry logic (`attempts: 3` with exponential backoff in `RunDownloadJob`) puts the job into `delayed` or `active` state for retry
3. The UI correctly shows it as "failed" because `job.failedReason` is set
4. However, BullMQ still considers the job active/locked because retries are pending
5. `job.remove()` throws: `Job could not be removed because it is locked by another worker`
6. The `removeJob` controller action (`downloads_controller.ts:19-21`) has no error handling, so the unhandled exception becomes a 500 response

## Suggested Fix

Before calling `job.remove()`, either:

1. **Discard retries first:**
```typescript
async removeFailedJob(jobId: string): Promise<void> {
  for (const queueName of [RunDownloadJob.queue, DownloadModelJob.queue]) {
    const queue = this.queueService.getQueue(queueName)
    const job = await queue.getJob(jobId)
    if (job) {
      try {
        await job.discard()  // Prevent further retries
      } catch {}
      try {
        await job.remove()
      } catch {
        // If still locked, force clean from the queue
        await queue.clean(0, 1, 'active')
      }
      return
    }
  }
}
```

2. **Or use `queue.clean()` as a fallback** when `job.remove()` fails due to a lock.

## Environment

- Project N.O.M.A.D. v1.30.3
- Running via Docker Compose on macOS (OrbStack)
- BullMQ with Redis 7

techyogi:
## Summary
- Fixes 500 error when clicking X to dismiss failed downloads on Remote Explorer page
- `job.discard()` cancels pending retries before removal
- Falls back to `job.moveToFailed()` + `job.remove()` if the job is still locked by a worker
- Controller now returns 422 with error message instead of unhandled 500

Closes #598

## Test plan
- [ ] Trigger a download that will fail (large file, kill network mid-download)
- [ ] Wait for it to show as "failed" in the UI
- [ ] Click X to dismiss — should succeed without 500 error
- [ ] Verify active downloads are unaffected

### DIFF
--- a/admin/app/services/download_service.ts
+++ b/admin/app/services/download_service.ts
@@ -66,11 +66,20 @@ export class DownloadService {
         try {
           await job.remove()
         } catch {
-          // Job is locked by an active worker — move it to failed state first,
-          // then remove. This handles the case where the UI shows a job as failed
-          // (failedReason is set) but BullMQ still considers it active due to retries.
-          await job.moveToFailed(new Error('Manually dismissed'), '0', false)
-          await job.remove()
+          // Job is locked by an active worker — move it to failed state first
+          // so it won't be retried. The remove may still fail if the lock hasn't
+          // released, so we catch that too and let the next dismiss attempt clean it up.
+          try {
+            await job.moveToFailed(new Error('Manually dismissed'), '0', false)
+          } catch {
+            // Already in failed state or lock contention — acceptable
+          }
+          try {
+            await job.remove()
+          } catch {
+            // Lock still held — job is now in failed state with retries discarded,
+            // so it will be removable on the next attempt
+          }
         }
         return
       }


chriscrosstalk:
Thanks for the fix. PR #554 already addresses this (stale lock deletion + `UnrecoverableError` to prevent retry loops), so closing in favor of that. Appreciate the contribution.

Issue #360, opened 2026-03-16
[Bug]: HTTPS Support lacking in Maps


davidgross:
### Issue Category

Installation/Setup

### Bug Description

setup an nginx reverse proxy in front of nomad, with https.
Some things work.
Maps does not.

### Steps to Reproduce

1) setup nomad
2) confirm maps working
3) setup an nginx reverse proxy with lets encrypt ssl.
4) view `/maps` pages, see that some endpoints are hard-coded to use `http`.

### Expected Behavior

Should allow for https urls in map app.

### Actual Behavior

Hard-coded `http` in map endpoints fails to load page, due to nginx hosted page using `https`.

### N.O.M.A.D. Version

1.29.1

### Operating System

Ubuntu 24.04

### Docker Version

Docker version 29.3.0, build 5927d80

### Do you have a dedicated GPU?

Yes

### System Specifications

VM on Truenas:
Hardware is:
CPU: AMD Ryzen 7 1700X Eight-Core Processor
RAM: VM has 6 of a total of 32GB
DISK: VM has 3 GB of 100 GB free

### Service Status (if relevant)

an issue in nomad_admin's code.

### Relevant Logs

```shell

```

### Browser Console Errors (if UI issue)

```javascript
Mixed Content: The page at 'https://nomad/maps' was loaded over HTTPS, but requested an insecure resource 'http://nomad/api/maps/styles'. This request has been blocked; the content must be served over HTTPS.
```

### Screenshots

it's just an empty page....

### Additional Context

my changes that made it work correctly for my setup: https://github.com/Crosstalk-Solutions/project-nomad/compare/main...davidgross:project-nomad:allow-maps-to-use-http-or-https

### Pre-submission Checklist

- [x] I have searched for existing issues that might be related to this bug
- [x] I am running the latest version of Project N.O.M.A.D. (or have noted my version above)
- [x] I have redacted any personal or sensitive information from logs and screenshots
- [x] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

chriscrosstalk:
Fixed in PR #394. The map styles URL on the frontend and the tile/asset URLs generated by the backend were both hardcoded to `http://`, which breaks when running behind an HTTPS reverse proxy (mixed content).

The fix uses `window.location.protocol` on the frontend and `request.protocol()` on the backend so URLs match whatever protocol the user is accessing NOMAD through.

davidgross:
[...]

PR #600, opened 2026-03-30
add x-forwarded-proto support to handle https termination

davidgross:
This fixes the remaining issues with https://github.com/Crosstalk-Solutions/project-nomad/issues/360

Nginx reverse proxy ssl termination results in `request.protocol()` saying `http` because the nginx->backend is http, but the user->nginx was https, so this checks for the `x-forwarded-proto` header, and sets the protocl if the header is present.

### DIFF
--- a/admin/app/controllers/maps_controller.ts
+++ b/admin/app/controllers/maps_controller.ts
@@ -83,7 +83,13 @@ export default class MapsController {
       })
     }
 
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
UPDATE:  Staging for v1.31.0

jakeaturner (approved):
Great fix, thanks!

PR #604, opened 2026-03-30
fix(queue): increase BullMQ lockDuration to prevent download stalls

### LINKED ISSUE #601: [Bug]: Download jobs stall due to default BullMQ lockDuration
The BullMQ download worker in `commands/queue/work.ts` uses the default `lockDuration` of 30 seconds. Downloads that don't report progress within that window get killed as stalled:

```
[ error ] [downloads] Job failed: <id>, Error: job stalled more than allowable limit
```

The initial HEAD request to the Kiwix CDN (302 redirect + mirror response) can eat most of that window, especially when multiple downloads start at once during easy setup (concurrency 3).

Multiple users hitting this on Discord — confirmed in admin logs on a Ryzen 3 2200G running v1.30.3. Partial files left behind from killed jobs also cause kiwix restart loops (#502).

**Suggested fix:** increase `lockDuration` on the download worker, e.g. `lockDuration: 300000` (5 minutes).

bgauger:
## Summary
- Increases BullMQ worker `lockDuration` from the default 30s to 300s (5 minutes)
- Prevents download jobs from being killed as stalled before data starts flowing
- Fixes the root cause behind multiple reported download failures on slower hardware

Closes #601

### DIFF
--- a/admin/commands/queue/work.ts
+++ b/admin/commands/queue/work.ts
@@ -61,6 +61,7 @@ export default class QueueWork extends BaseCommand {
         {
           connection: queueConfig.connection,
           concurrency: this.getConcurrencyForQueue(queueName),
+          lockDuration: 300000,
           autorun: true,
         }
       )


chriscrosstalk:
UPDATE:  Staging for v1.31.0

jakeaturner (approved):
Another great fix, thanks!

PR #610, opened 2026-03-31
feat: group managed containers under project-nomad in Docker UI

techyogi:
## Summary
- Adds `com.docker.compose.project: project-nomad` label to containers created by DockerService
- Managed containers (Ollama, Kiwix, CyberChef, etc.) now appear grouped with compose services in Docker Desktop and OrbStack instead of as standalone containers

One-line change, metadata only — no impact on container behavior, networking, or data.

## Test plan
- [ ] Install any service from the NOMAD UI
- [ ] Check Docker Desktop or OrbStack — new container should appear under the `project-nomad` group
- [ ] Verify service functions normally

### DIFF
--- a/admin/app/services/docker_service.ts
+++ b/admin/app/services/docker_service.ts
@@ -508,6 +508,9 @@ export class DockerService {
       const container = await this.docker.createContainer({
         Image: finalImage,
         name: service.service_name,
+        Labels: {
+          'com.docker.compose.project': 'project-nomad',
+        },
         ...(containerConfig?.User && { User: containerConfig.User }),
         HostConfig: gpuHostConfig,
         ...(containerConfig?.WorkingDir && { WorkingDir: containerConfig.WorkingDir }),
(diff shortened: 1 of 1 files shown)


chriscrosstalk:
UPDATE:  Staging for v1.31.0

jakeaturner:
@techyogi Excellent suggestion, thank you! I modified some things slightly here by setting the `com.docker.compose.project` label to `project-nomad-managed` instead of `project-nomad` to intentionally separate what's created by NOMAD's core "management" compose file, and what's dynamically managed by NOMAD via the Docker socket. This will break things up into two groups, but I purposely went this route to avoid confusion from users expecting `docker compose down` to stop ALL containers related to NOMAD (less pretty, but more clear). 

Also took the opportunity to support merging in `containerConfig.Labels`. None of the predefined installable containers actually have `Labels` defined in their DB records, but opens the door for it down the road.

PR #362, opened 2026-03-16
Upgrade GitHub Actions to latest versions

salmanmkc:
Bumps GitHub Actions to their latest versions for bug fixes and security patches.

## Changes

| Action | Old Version(s) | New Version | Compare | Files |
|--------|---------------|-------------|---------|-------|
| `cycjimmy/semantic-release-action` | [`v3`](https://github.com/cycjimmy/semantic-release-action/releases/tag/v3) | [`v6`](https://github.com/cycjimmy/semantic-release-action/releases/tag/v6) | [Diff](https://github.com/cycjimmy/semantic-release-action/compare/v3...v6) | release.yml |
| `docker/build-push-action` | [`v5`](https://github.com/docker/build-push-action/releases/tag/v5) | [`v7`](https://github.com/docker/build-push-action/releases/tag/v7) | [Diff](https://github.com/docker/build-push-action/compare/v5...v7) | build-disk-collector.yml, build-primary-image.yml |
| `docker/login-action` | [`v2`](https://github.com/docker/login-action/releases/tag/v2) | [`v4`](https://github.com/docker/login-action/releases/tag/v4) | [Diff](https://github.com/docker/login-action/compare/v2...v4) | build-disk-collector.yml, build-primary-image.yml |

## Notes

Worth running the workflows on a branch before merging to make sure everything still works.

PR #615, opened 2026-04-01
feat(docs): add simple API reference

hestela:
Adds tables with method,path and description in /docs/api-reference/

image


hestela:
broken build seems to be due to #362 . reverting it locally unblocks my docker build

hestela:
had to force push since dev had a force push

PR #618, opened 2026-04-01
Add macOS-compatible install path without changing Debian defaults

### LINKED ISSUE #1: Kiwix ZIM downloads
I tried to download some ZIM files into a fresh install of NOMAD.  

Full wikipedia download (~108GB) - downloaded about 2GB and then stopped.

medical wiki (1.8GB) - looks like it downloaded fully, but never showed up in Kiwix
Prunelle books (359MB) - looks like it downloaded fully, but never showed up in Kiwix

Probably need some checks to ensure any ZIM file downloads completely and successfully?  Perhaps even a status indicator while download is in progress?

stormychel:
## Summary
This PR adds a macOS-compatible installation path for Project N.O.M.A.D. while preserving the current Debian/Linux defaults and behavior.

The implementation is intentionally compatibility-first:
- Debian/Linux remains the default install path and keeps `/opt/project-nomad`
- macOS uses a separate installer path and compose template
- Linux-only host integrations are explicitly gated instead of being weakened to fit macOS
- content and AI pages now offer self-service install actions instead of dead-end warnings

### Installer and runtime compatibility
- detect Debian/Linux vs macOS in the installer
- keep `/opt/project-nomad` as the Debian/Linux default
- use `~/.project-nomad` as the macOS default install root
- validate Docker Desktop on macOS instead of attempting Debian package management
- add platform-specific compose templates for Linux and macOS
- make update/uninstall/helper scripts respect configurable install roots

### macOS-specific behavior
- omit Linux-only disk collector integration on macOS
- skip Linux-only NVIDIA toolkit setup on macOS
- seed Ollama without binding host port `11434` on macOS to avoid collisions with local host Ollama installs
- resolve internal service URLs from Docker `ExposedPorts`, not only host port bindings

### UX fixes discovered during live validation
- add `Install Kiwix` actions on the content pages
- add `Install AI Assistant` action on the models page
- prevent model downloads from running before Ollama is installed
- make Easy Setup wait for Ollama installation before dispatching model downloads

### Documentation
- update README prerequisites and install instructions for Debian/Linux and macOS
- document Docker Desktop requirements and the macOS install root
- document the Apple Silicon / emulation caveat observed during testing
- document the host Ollama port-collision case on macOS
- update helper script docs and in-app legal/storage references
- clarify that the disk-collector migration flow is Linux-only

## Why This Is Safe For Debian/Linux
This PR does not replace the Debian/Linux install path with a generic cross-platform flow.

Instead, it preserves the existing Debian/Linux assumptions and behavior, then adds a separate macOS path around them:
- Debian/Linux still defaults to `/opt/project-nomad`
- Linux compose behavior remains intact in its own template
- Linux-only features remain enabled only on Linux
- macOS-specific adjustments are isolated behind platform detection and platform-specific config

### Shell / config validation
- `bash -n` passed for the touched installer and maintenance shell scripts
- `git diff --check` is clean against `main`
- Docker admin image built successfully from this branch

### Live macOS validation
Validated on macOS with Docker Desktop running:
- installer completed successfully into `~/.project-nomad`
- management stack came up successfully
- `nomad_admin`, `nomad_mysql`, `nomad_redis`, `nomad_dozzle`, and `nomad_updater` were verified running
- branch-built local admin image was swapped into the live stack for verification
- `NOMAD_PLATFORM=macos` was confirmed in the running admin container
- `nomad_ollama` install was re-tested after rebuild and now succeeds without binding host port `11434`
- fresh model-download dispatch was verified after Ollama came up

### Debian/Linux regression review
I reviewed this branch directly against `main` with Debian/Linux regression risk in mind:
- Debian/Linux still routes through the same Debian-only installer assumptions (`/opt/project-nomad`, `apt-get`, `systemctl`, NVIDIA toolkit path)
- Linux host integrations are preserved in the Linux compose variant, including `disk-collector` and Linux filesystem mounts
- Linux Ollama behavior is preserved: host port `11434` is still bound on Linux; the no-host-port change is macOS-only via `NOMAD_PLATFORM=macos`
- updater and uninstall scripts still default to `/opt/project-nomad` on Linux and only branch for macOS when `uname` is `Darwin`
- runtime disk display changes are additive fallbacks; on Linux, `/dev/*` devices still win exactly as before
- Kiwix/Ollama UI changes are defensive UX improvements and do not remove existing Linux functionality

### Notes
- Apple Silicon currently runs the published admin/updater images under emulation when using the upstream images available today. This was observed during testing and is documented, but it does not block installation.
- I was not able to run the full installer/update/runtime smoke test on an actual Debian host from this environment. The diff review above substantially reduces Debian regression risk, but it is not a substitute for a live Debian run.
- I was not able to run the repo's local `npm run typecheck` / `npm run lint` directly in this workspace before the Docker build because local toolchain binaries were not present. The production image build from this branch completed successfully and the live stack was validated after rebuild.

## Follow-up Considerations
- native multi-arch release images would improve Apple Silicon performance
- additional explicit UI messaging could be added later for platform-limited features such as disk inspection / GPU passthrough on macOS

Closes #1

### DIFF
--- a/install/migrate-disk-collector.md
+++ b/install/migrate-disk-collector.md
@@ -1,6 +1,6 @@
 # Project N.O.M.A.D. — About the Disk Collector Migration Script
 
-This script migrates your Project N.O.M.A.D. installation from the old host-based disk info collector to the new disk-collector sidecar. It modifies `/opt/project-nomad/compose.yml` to add the new service and remove the old bind mount, then restarts the full compose stack to apply the changes.
+This script migrates your Project N.O.M.A.D. installation from the old host-based disk info collector to the new disk-collector sidecar. It is intended for Debian/Linux installs and updates the compose file in your install directory, which defaults to `/opt/project-nomad/compose.yml`.
 
 ### Why the Migration?
 The new disk-collector sidecar provides a more robust and scalable way to collect disk information from the host. It removes the original bind mount to `/tmp/nomad-disk-info.json`, which was fragile and prone to issues on host reboots.
@@ -12,6 +12,8 @@ The original host-based collector relied on a process running on the host that w
 
 The migration script automates the necessary changes to your compose configuration and ensures a smooth transition to the new architecture.
 
+This migration is Linux-only. The macOS install path intentionally does not run the disk-collector sidecar because Docker Desktop does not expose the host filesystem in the same way.
+
 ### Why does Nomad need the nomad-disk-info.json file?
 Nomad uses the disk info stored and updated in `nomad-disk-info.json` to allow users to view disk usage and availability within the Nomad "Command Center". While not critical to the core functionality of Nomad, it provides a more pleasant experience for users with limited storage space and/or who aren't familiar with command-line tools and Linux management.
 
@@ -22,7 +24,7 @@ The disk-collector runs in a separate container to isolate its functionality fro
 - **Modularity**: Because having the host disk info is not a critical component of Nomad's core functionality, isolating it in a sidecar allows users who don't need/want the disk info features to simply not run that container, without impacting the main admin container or other services. It also allows for more flexible future development of the disk-collector without needing to modify the main admin container.
 
 ### What if I don't want to run the migration script?
-No worries - you can replicate the changes manually by editing your `/opt/project-nomad/compose.yml` to add the new disk-collector service and remove the old bind mount from the admin service, then restarting your compose stack. The migration script just automates these steps and ensures they're done correctly, but the underlying changes are straightforward if you prefer to do it yourself. Just be sure to back up your `compose.yml` before making any changes.
+No worries - you can replicate the changes manually by editing the `compose.yml` in your install directory, which is `/opt/project-nomad/compose.yml` by default on Debian/Linux, to add the new disk-collector service and remove the old bind mount from the admin service, then restarting your compose stack. The migration script just automates these steps and ensures they're done correctly, but the underlying changes are straightforward if you prefer to do it yourself. Just be sure to back up your `compose.yml` before making any changes.
 
 Here's the disk-collector service configuration to add to your `compose.yml`:
 


klendgit:
I checked out this PR and it went relatively smooth on my Intel-Mac. After choosing the relevant content, the installation process showed this message, but i thought there was a code change especially for this?

> llama3.2
> Download failed: Ollama service is not installed or running.

After download of all zims the system worked fine and the llama3.2 download went ok automatically as well.
Seems a good addition to me and I think it would benefit the community. Will also try this on Apple Silicon and report.

uninstall_nomad.sh does not work initially:

> macOS uninstall path does not require sudo by default.
> Unable to find the management Docker Compose file at /opt/project-nomad/compose.yml. There may be a problem with your Project N.O.M.A.D. installation.

After setting `export NOMAD_DIR=$HOME/.project-nomad` Iin my normal login shell) it worked fine:

> bash ./uninstall_nomad.sh
> macOS uninstall path does not require sudo by default.
> This script will remove ALL Project N.O.M.A.D. files and containers. THIS CANNOT BE UNDONE. Are you sure you want to continue? (y/n): y

Apple Silicon also works but I noticed this message during creating of the containers:
>  ! admin - The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested

benchmark:
Intel macbook pro mid 2015: 37
M1 Pro 2021: 57

stormychel:
@klendgit Thanks for giving this a spin, will make some improvements based on your findings!

chriscrosstalk:
Thanks for the effort here. Right now we're focused on stabilizing x86 bare-metal Debian-based deployments. Once the project is solid on that platform, we'll look at expanding to additional architectures like macOS.

For a lightweight offline knowledge solution on Mac in the meantime, check out [Internet-in-a-Box](https://internet-in-a-box.org/).

PR #632, opened 2026-04-02
Fix download retry config and abort handling for large files

### LINKED ISSUE #627: fix(downloads): global map download fails mid-transfer -- needs auto-retry with resume
## Problem

The global map download (~125 GB from Protomaps CDN) fails mid-transfer due to CDN connection stalls. The download stream stops receiving data, the 5-minute stall timeout in `progressStream` detects the stall and aborts the download. Users see "Download failed: Download cancelled: aborted" with no automatic recovery.

## Observed Behavior (v1.31.0-rc.1, NOMAD3)

**Attempt 1:** Downloaded to ~51 GB (40%), CDN stalled, auto-aborted after 5 min timeout.

**Attempt 2:** Manually dismissed failed download, restarted. Download resumed at 40% via HTTP range request (the partial file was preserved). Downloaded to ~101 GB (80%), CDN stalled again at the same "zero bytes for 4+ minutes" pattern, auto-aborted.

File size polling confirmed the stall:

| Time | Size | Delta |
|------|------|-------|
| 10:44 | 90.7 GB | +12.4 GB |
| 10:46 | 101.2 GB | +10.5 GB |
| 10:47 | 101.2 GB | 0 |
| 10:48 | 101.2 GB | 0 |
| 10:49 | 101.2 GB | 0 |
| 10:50 | 101.2 GB | 0 (aborted) |

Download speed was a consistent 5-6 GB/min when active. The CDN appears to drop connections after sustained large transfers.

## Key Finding: Resume Works

HTTP range requests work with the Protomaps CDN. When the download was manually restarted, it picked up at 40% from the existing partial file. This means auto-retry would work without re-downloading from zero.

## Suggested Fix

1. **Auto-retry with resume on stall/failure:** When a download fails due to a stall timeout or connection error, automatically retry with an HTTP `Range` header starting from the existing partial file size. Use BullMQ's built-in retry mechanism (allow regular errors instead of `UnrecoverableError` for resumable failures).

2. **Retry limit with backoff:** Allow 3-5 automatic retries with increasing delay (e.g., 30s, 60s, 120s) before surfacing the failure to the user. A 125 GB file that stalls every ~50 GB would complete in 3 retries without any user intervention.

3. **Consider increasing stall timeout for large files:** 5 minutes is reasonable for typical ZIM downloads (1-15 GB), but CDN behavior on 125 GB transfers may include brief pauses. A longer timeout (10-15 min) or a dynamic timeout scaled to file size could help.

## Additional UX Notes

- After the first failure, clicking "Download Global Map" again showed an error (likely "download already in progress" from the stale failed job). Had to dismiss the failed download card first, then retry. The dismiss-then-retry flow should be streamlined or automatic.
- The failed download card shows "Download cancelled: aborted" which is confusing since the user didn't cancel anything. A message like "Download interrupted -- connection lost" would be clearer.

## Files Involved

- `admin/app/utils/downloads.ts` -- `progressStream` stall timeout (currently 5 min)
- `admin/app/jobs/run_download_job.ts` -- download worker, abort handling
- `admin/app/services/download_service.ts` -- job dispatch and cancel logic
- `admin/app/services/map_service.ts` -- global map download entry point

bgauger:
## Summary
- Increase retry attempts (3→10) with exponential backoff (30s base) for large file downloads like global maps that experience CDN stalls
- Fix `UnrecoverableError` being thrown on BullMQ lock mismatches — only user-initiated cancels should prevent retries

## Testing
Tested a full global map download (~126GB) on a test server. The download hit two "aborted" errors mid-transfer from lock mismatches and retried through both successfully.

Closes #627

### DIFF
--- a/admin/app/jobs/run_download_job.ts
+++ b/admin/app/jobs/run_download_job.ts
@@ -67,7 +67,7 @@ export class RunDownloadJob {
         if (val) {
           await cancelRedis.del(RunDownloadJob.cancelKey(job.id!))
           userCancelled = true
-          abortController.abort()
+          abortController.abort('user-cancel')
         }
       } catch {
         // Redis errors are non-fatal; in-process AbortController covers same-process cancels
@@ -184,7 +184,8 @@ export class RunDownloadJob {
     } catch (error: any) {
       // Only prevent retries for user-initiated cancellations. BullMQ lock mismatches
       // can also abort the stream, and those should be retried with backoff.
-      if (userCancelled) {
+      // Check both the flag (Redis poll) and abort reason (in-process cancel).
+      if (userCancelled || abortController.signal.reason === 'user-cancel') {
         throw new UnrecoverableError(`Download cancelled: ${error.message}`)
       }
       throw error
--- a/admin/app/jobs/run_download_job.ts
+++ b/admin/app/jobs/run_download_job.ts
@@ -54,6 +54,11 @@ export class RunDownloadJob {
       totalBytes: 0,
     }
 
+    // Track whether cancellation was explicitly requested by the user (via Redis signal
+    // or in-process AbortController). BullMQ lock mismatches can also abort the download
+    // stream, but those should be retried — only user-initiated cancels are unrecoverable.
+    let userCancelled = false
+
     // Poll Redis for cancel signal every 2s — independent of progress events so cancellation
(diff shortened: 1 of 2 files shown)


jakeaturner (dismissed):
Sweet!

PR #649, opened 2026-04-04
fix(ai-chat): qwen2.5 loading on every chat message

hestela:
qwen was getting loaded to generate the title for the chat but it was also getting loaded on each new message even if you manually unloaded the model.

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -8,7 +8,7 @@ import { modelNameSchema } from '#validators/download'
 import { chatSchema, getAvailableModelsSchema } from '#validators/ollama'
 import { inject } from '@adonisjs/core'
 import type { HttpContext } from '@adonisjs/core/http'
-import { DEFAULT_QUERY_REWRITE_MODEL, RAG_CONTEXT_LIMITS, SYSTEM_PROMPTS } from '../../constants/ollama.js'
+import { RAG_CONTEXT_LIMITS, SYSTEM_PROMPTS } from '../../constants/ollama.js'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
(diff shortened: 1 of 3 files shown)


chriscrosstalk (commented):
Reviewed the diff -- well-designed fix that addresses two issues:

1. **Title generation** was loading qwen2.5 as a side effect on every new conversation, even when the user was chatting with a different model. Now uses the user's selected model instead.

2. **Query rewriting** had the same problem -- hardcoded to DEFAULT_QUERY_REWRITE_MODEL, causing surprise model loads on every message. Now uses the user's model and skips the RAG pipeline entirely when there are no documents (nice optimization via `hasDocuments()`).

The fix removes complexity (no more `getModels()` availability checks, removes `DEFAULT_QUERY_REWRITE_MODEL` dependency) rather than adding it. The tradeoff of using the user's selected model for rewriting instead of a small dedicated model is the right call -- no user wants a model loading behind their back.

Recommend including in a v1.31.1 hotfix alongside PR #645.

chriscrosstalk (dismissed):
LGTM — this is the right architectural direction. Eliminating `DEFAULT_QUERY_REWRITE_MODEL` entirely and reusing the user's currently-selected model for rewriting and title generation is the cleaner fix, and the `hasDocuments()` early-exit is a nice bonus optimization for users who haven't populated a knowledge base.

The reported bug (qwen2.5 reloading on every chat message even when manually unloaded) is a real UX problem that's been affecting users since the query rewrite pipeline landed, and this fixes it at the root.

Noted as a strong candidate for **v1.31.1** inclusion alongside #645 (null check on model name) and #660 (ZIM crash prevention, pending rebase). Approving.

jakeaturner (approved):
Great fixes! JT 4/17/2026

PR #657, opened 2026-04-05
feat(UI): add advanced settings and ui_location editor (#271)

boushley:
This is a draft at this point because I haven't gotten it running or tested locally yet. Seeking input early though to see if I'm headed in the desired direction.


chriscrosstalk:
Thanks for taking a crack at this! We appreciate the initiative on #271. However, the Advanced Features section is something we're going to design and build ourselves when the timing is right -- there are several interconnected settings we want to roll out together (custom storage paths, app URLs, auto-update toggles, etc.) and we want to make sure the UX is cohesive.

Closing this out, but thanks again for the interest in contributing!

Issue #388, opened 2026-03-17
ZIM files not fully embedded into RAG knowledge base after download


dcbenji:
## Bug Description

ZIM files downloaded through the Content Explorer are only partially embedded into the Qdrant knowledge base. After downloading 11 ZIM files (7.8GB total including Wikipedia, Stack Exchange sites, cooking references, survival content), only 495 vector points exist in Qdrant — roughly one batch of 50 articles per file.

## Steps to Reproduce

1. Fresh install of Project NOMAD v1.29.x
2. During setup, enable AI Assistant (installs Ollama + Qdrant)
3. Download multiple ZIM content packs from the Content Explorer
4. Check Qdrant collection — only ~495 points across all files
5. Click "Sync Storage" in Knowledge Base — reports "scanned 11 files, queued 2 for embedding" (treats 9 as already processed)

## Expected Behavior

All ZIM articles should be fully chunked and embedded into Qdrant for RAG retrieval. The batch processing should continue through all articles in each ZIM file, not stop after the first batch.

## Actual Behavior

Each ZIM file gets approximately one batch (~50 articles) embedded, then subsequent batches never run. "Sync Storage" considers files with any existing points as fully processed, so there's no way to resume incomplete embedding.

## Likely Root Cause (Hypothesis)

There may be a race condition in the initial setup flow:

1. User installs NOMAD and enables AI Assistant
2. ZIM content packs begin downloading immediately
3. The embedding model (`nomic-embed-text:v1.5`) and the recommended RAG query-rewrite model (`qwen2.5:3b`) are NOT auto-downloaded — the user only sees the recommendation to install `qwen2.5:3b` later in the chat UI
4. When ZIM downloads complete, `RunDownloadJob` dispatches `EmbedFileJob`, but if the embedding model isn't ready yet, the first batch may partially succeed or fail
5. The `scanAndSyncStorage()` method then considers files with any points in Qdrant as "already processed," preventing re-embedding of the remaining content
6. Result: each ZIM has a handful of points but is marked as complete

Additionally, `_ensureCollection` in `rag_service.ts` throws on "collection already exists" rather than handling it gracefully, which may cause batch continuation failures.

## Environment

- Project NOMAD v1.29.1
- Ubuntu 24.04 LTS
- NVIDIA RTX 4070 (12GB VRAM)
- 32GB RAM
- Docker Compose v5.1.0

## Suggested Fix

[...]

PR #660, opened 2026-04-05
fix: prevent ZIM corrupt file crash and deduplicate Ollama download logs

ghost:
## Summary
- **ZIM crash fix**: Corrupted `.zim` files cause a native C++ `ZimFileFormatError` abort that bypasses JS try/catch and kills the entire process. Added magic number validation (`5A 49 4D 04`) before passing files to `@openzim/libzim` — invalid files are now skipped gracefully with a warning.
- **Ollama download dedup**: Fixed duplicate download progress log lines caused by (1) the stream emitting the same percentage twice per chunk, and (2) concurrent embedding jobs both triggering `downloadModel` for the same model. Added per-stream percentage tracking and a shared download promise map so concurrent callers reuse one pull stream.

## Test plan
- [ ] Start workers with a corrupted `.zim` file in storage — verify it logs a warning and continues instead of crashing
- [ ] Queue multiple embedding jobs simultaneously — verify only one `nomic-embed-text` download stream runs
- [ ] Verify download progress logs show each percentage only once

### DIFF
--- a/admin/app/services/kiwix_library_service.ts
+++ b/admin/app/services/kiwix_library_service.ts
@@ -2,7 +2,7 @@ import { XMLBuilder, XMLParser } from 'fast-xml-parser'
 import { readFile, writeFile, rename, readdir } from 'fs/promises'
 import { join } from 'path'
 import { Archive } from '@openzim/libzim'
-import { KIWIX_LIBRARY_XML_PATH, ZIM_STORAGE_PATH, ensureDirectoryExists } from '../utils/fs.js'
+import { KIWIX_LIBRARY_XML_PATH, ZIM_STORAGE_PATH, ensureDirectoryExists, isValidZimFile } from '../utils/fs.js'
 import logger from '@adonisjs/core/services/logger'
 import { randomUUID } from 'node:crypto'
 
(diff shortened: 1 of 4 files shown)


chriscrosstalk (approved):
Really valuable fix, especially the ZIM crash prevention. Catching the native C++ abort by magic-number-gating the file handoff to `@openzim/libzim` is exactly the right approach — that's the only way to guard against it from JS since native aborts bypass try/catch entirely.

**Worth noting:** this may be directly contributing to #388 (ZIM files not fully embedded into RAG). If a user has even one corrupt ZIM in storage, the embedding worker crashes on it, BullMQ retries, the worker crashes again, and embedding never completes. That matches the symptom pattern multiple users are reporting in #388. This could be a meaningful piece of that puzzle.

**One thing to address before merge:**

- **Base branch is `main` — should be `dev`.** Please rebase against `dev` so this can merge cleanly into the normal release flow.

**Minor nit (non-blocking):** the comment above the magic number check says `"72 17 32 04 (little-endian 0x044D4953)"` — the bytes in the code (`0x5A 0x49 0x4D 0x04`) are correct, but the comment's hex representation is off. Should be `0x044D495A`. Worth cleaning up on the rebase but not a blocker.

Once rebased, this is a strong candidate for v1.31.1 inclusion alongside the other AI-related hotfixes.

jakeaturner (approved):
Great stability fixes! JT 4/17/2026

Issue #384, opened 2026-03-17
[Bug]: Storage Devices UI reports incorrect disk size on mdadm RAID + LVM systems


bjazmoore:
### Issue Category

UI/Frontend Issue

### Bug Description

On a Linux host using mdadm RAID5 with LVM on top and a large ext4 filesystem, the Project N.O.M.A.D. UI “Storage Devices” section reports each physical disk (sda–sdd) as 1.56GB total with 2.44MB used, instead of showing the actual RAID/LVM filesystem capacity (multi‑TB). This appears to be a parsing/selection issue in the disk info collection logic.

This is likely only a reporting/UI bug; actual storage is available and usable at the OS level.

**Environment**

- Project: Crosstalk-Solutions/project-nomad
- Install type: Script install (terminal-based)
- Host OS: Debian-based Linux (Pop!_OS; Ubuntu-derived)
- Storage layout:
    - 4x SATA HDDs in mdadm RAID5 (md2)
    - LVM PV on /dev/md2, VG vg1000, LV lv (/dev/mapper/vg1000-lv)
    - ext4 filesystem on LV
    - (Optional detail) Docker data-root moved to array, but UI issue existed regardless of actual data usage

**Expected behavior**

“Storage Devices” should report meaningful capacity/usage for the storage that matters to NOMAD, for example:

- the filesystem(s) where NOMAD stores content (e.g., the filesystem backing Docker volumes / bind mounts), or
- the assembled mdraid device (/dev/md2) and/or the LVM LV (/dev/mapper/vg1000-lv), or
- at least not report obviously incorrect sizes for sda–sdd.

**Actual behavior**

The UI reports each physical disk as:

- sda — 2.44 MB / 1.56 GB — 0% — Normal
- sdb — 2.44 MB / 1.56 GB — 0% — Normal
- sdc — 2.44 MB / 1.56 GB — 0% — Normal
- sdd — 2.44 MB / 1.56 GB — 0% — Normal

This is not accurate. Each drive is ~4.5TB and the assembled filesystem is ~13.6T.

**Evidence / Diagnostics**

The disk info JSON generated on the host is:

- File: /opt/project-nomad/storage/nomad-disk-info.json

Key excerpts (sanitized for brevity):

- diskLayout.blockdevices correctly shows nested topology:
    - sda5 → md2 (type raid5, size 13.6T) → vg1000-lv (type lvm, size 13.6T)
    - similarly for sdb/sdc/sdd

[...]

Issue #373, opened 2026-03-16
[Bug]: Disk Collector reporting tmpfs /run (1.5GB) instead of actual storage mount


cmhubbert:
### Issue Category

Installation/Setup

### Bug Description

The storage display in the Command Center shows 2.5 MB / 1.5 GB instead of the 
actual available storage. The disk-collector container is incorrectly reporting 
the tmpfs /host/run filesystem (1.5GB) instead of the actual /storage mount.

When running "docker exec nomad_disk_collector df -h" the output clearly shows:

- /dev/sdb1 at /storage = 915.8G available (CORRECT drive)
- tmpfs at /host/run = 1.5G (INCORRECTLY reported in UI)

The UI shows exactly 1.5GB which matches the tmpfs /host/run, not the 
915GB /storage mount.

### Steps to Reproduce

1. Install Project NOMAD on a system where /storage is mounted on a separate 
   drive from the OS
2. Open Command Center in browser
3. Navigate to Apps > Maps or any add-on
4. Observe Storage shows 2.5 MB / 1.5 GB instead of actual drive size

### Expected Behavior

1. Install Project NOMAD on a system where /storage is mounted on a separate 
   drive from the OS
2. Open Command Center in browser
3. Navigate to Apps > Maps or any add-on
4. Observe Storage shows 2.5 MB / 1.5 GB instead of actual drive size

### Actual Behavior

Storage displays 2.5 MB / 1.5 GB which matches the tmpfs /host/run filesystem

### N.O.M.A.D. Version

- NOMAD Version: 1.29.1 - OS: Zorin OS 17 (Ubuntu 24.04 Noble base) - Kernel: 6.17.0-19-generic - Docker: CE 29.3.0 - Storage: /storage mounted on separate 1TB HDD (/dev/sdb1) - Boot drive: 931GB NVMe SSD (/dev/nvme0n1)

### Operating System

Ubuntu 24.04

### Do you have a dedicated GPU?

Yes

### Relevant Logs

```shell

```

### Browser Console Errors (if UI issue)

```javascript

```

### Additional Context

docker exec nomad_disk_collector df -h output:

Filesystem    Size    Used  Available  Use%  Mounted on
/dev/sdb1    915.8G  204.9M  869.0G    0%   /storage       <- CORRECT
tmpfs          1.5G    2.5M    1.5G    0%   /host/run      <- UI shows this

The disk-collector appears to be selecting the wrong filesystem when 
[...]

Issue #320, opened 2026-03-15
[Bug]: not detecting my disk space


oogreen:
### Issue Category

System Performance/Resources

### Bug Description

when using the setup wizard it detects 1.1GB of storage space when i have more then 900GB HHD.
system information shows 1.16GB sda under storage device

### Steps to Reproduce

on fresh install i got a mysql error, running "sudo docker ps" showed me that it was running and fine but admin was not there.  i ran "sudo docker ps -a"  and found that  admin was there but not started so i started it manualy "sudo docker start nomad_admin" after about a minute of starting the browser command center worked but startub wizard only shows 1.7GB. (1.8GB before i rebooted )

### Expected Behavior

i expected it to see the full TB of HHD

### Actual Behavior

i am only able to see and use 1.16GB

### N.O.M.A.D. Version

latest

### Operating System

Ubuntu 24.04

### Docker Version

29.3.0

### Do you have a dedicated GPU?

Yes

### Service Status (if relevant)

sudo docker ps
CONTAINER ID   IMAGE                                                             COMMAND                  CREATED          STATUS                    PORTS                                                    NAMES
d1649f3d4325   ghcr.io/crosstalk-solutions/project-nomad:latest                  "/usr/local/bin/entr…"   47 minutes ago   Up 23 minutes (healthy)   0.0.0.0:8080->8080/tcp, [::]:8080->8080/tcp              nomad_admin
e43fb6a35213   redis:7-alpine                                                    "docker-entrypoint.s…"   47 minutes ago   Up 23 minutes (healthy)   0.0.0.0:6379->6379/tcp, [::]:6379->6379/tcp              nomad_redis
0f3904d3d46a   mysql:8.0                                                         "docker-entrypoint.s…"   47 minutes ago   Up 23 minutes (healthy)   0.0.0.0:3306->3306/tcp, [::]:3306->3306/tcp, 33060/tcp   nomad_mysql
07805a899c88   amir20/dozzle:v10.0                                               "/dozzle"                47 minutes ago   Up 23 minutes             0.0.0.0:9999->8080/tcp, [::]:9999->8080/tcp              nomad_dozzle
35e3c845ef7a   project-nomad-updater                                             "/usr/local/bin/upda…"   47 minutes ago   Up 23 minutes                                                                      nomad_updater
[...]

PR #686, opened 2026-04-08
fix(disk-collector): fix storage reporting for NFS mounts

bgauger:
## Summary
- `df` wraps long NFS device names across two lines, breaking the `awk 'NR==2'` parser in the disk collector and producing empty size/used/available values. Adding `-P` forces POSIX single-line output.
- The easy-setup storage bar only checked physical block devices and `/dev/*` filesystems, so NFS-backed storage was never picked up. Now checks the `/app/storage` mount first.

## Test plan
- [x] Verified on a host with NFS-mounted storage — disk collector JSON now reports correct values, easy-setup shows full NAS capacity
- [x] Verified on a bare-metal local-disk install — no regression, storage displays correctly
- [x] Dashboard storage devices section unaffected

Related: #373, #320, #384

### DIFF
--- a/admin/inertia/hooks/useDiskDisplayData.ts
+++ b/admin/inertia/hooks/useDiskDisplayData.ts
@@ -55,20 +55,24 @@ export function getAllDiskDisplayItems(
     const seen = new Set<number>()
     const uniqueFs = fsSize.filter((fs) => {
       if (fs.size <= 0 || seen.has(fs.size)) return false
+      if (storageMount && fs.mount === '/app/storage') return false
       seen.add(fs.size)
       return true
     })
     const realDevices = uniqueFs.filter((fs) => fs.fs.startsWith('/dev/'))
     const displayFs = realDevices.length > 0 ? realDevices : uniqueFs
-    return displayFs.map((fs) => ({
-      label: fs.fs || 'Unknown',
-      value: fs.use || 0,
-      total: formatBytes(fs.size),
-      used: formatBytes(fs.used),
-      subtext: `${formatBytes(fs.used)} / ${formatBytes(fs.size)}`,
-      totalBytes: fs.size,
-      usedBytes: fs.used,
-    }))
+    return [
+      ...storageMountItem,
+      ...displayFs.map((fs) => ({
+        label: fs.fs || 'Unknown',
+        value: fs.use || 0,
+        total: formatBytes(fs.size),
+        used: formatBytes(fs.used),
+        subtext: `${formatBytes(fs.used)} / ${formatBytes(fs.size)}`,
+        totalBytes: fs.size,
+        usedBytes: fs.used,
+      })),
+    ]
   }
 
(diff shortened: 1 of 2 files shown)


bravosierra99:
This is my exact issue... and fixed in exactly the same way.

bravosierra99:
Great fix! Two things we found testing this on NFS-backed storage (TrueNAS):

**1. `getAllDiskDisplayItems` has the same bug** — the `settings/system` page calls that function, not `getPrimaryDiskInfo`. Since it returns block-device data immediately when any disk is found, the NFS mount never appears there. Fix is to prepend the `/app/storage` entry when present, so both NAS and OS disk are shown.

**2. Worth a comment on `/app/storage`** — it looks surprising that we're checking the container path rather than the host path. A quick note explaining that `si.fsSize()` runs inside the admin container (where storage is bind-mounted at `/app/storage`) would help future readers.

Happy to open a PR against your branch with both fixes if useful. Verified working on NFS.

[useDiskDisplayData.ts.txt](https://github.com/user-attachments/files/26570980/useDiskDisplayData.ts.txt)

bgauger:
nice catch on getAllDiskDisplayItems — i didnt hit that on my setup since the fsSize fallback picked it up, but makes sense that bare metal with physical disks would skip right past the NFS mount. feel free to PR against my branch with both fixes.

bravosierra99:
I've submitted the PR on bgauger

bgauger:
@bravosierra99 merged your fix in — thanks! One thing we found testing on the NFS box: when there are no block devices (like on a containerized setup with only NFS), `validDisks` is empty so the code falls through to the `fsSize` fallback path, which doesn't use the `storageMountItem`. The NFS mount still showed up but with the raw path as the label instead of "NAS Storage". Added a small fix to prepend `storageMountItem` in the fsSize fallback and exclude `/app/storage` from the generic list so it doesn't double up.

chriscrosstalk (dismissed):
LGTM. This is a real fix for real user pain on NFS-backed storage, and the collaborative iteration with bravosierra99's follow-up on `getAllDiskDisplayItems` plus your NFS-only fallback covers the edge cases cleanly.

The `-P` flag on `df` is the textbook fix for the multi-line wrapping behavior with long NFS device names, and the frontend changes are well-scoped. Tested paths (NFS + bare-metal no-regression) match what I'd want to see.

One minor nit (non-blocking): the hardcoded "NAS Storage" label is slightly misleading for the edge case where `/app/storage` is a bind mount from a local disk rather than NFS. If you want to polish later, showing the filesystem type or actual device would be nicer. But it's a nit — ship it.

Flagging for inclusion in **v1.31.1** as a patch fix — this is the kind of concrete user-facing improvement worth getting out quickly alongside the other hotfix candidates (#645, #649).

jakeaturner (approved):
Sweet! JT 4/17/2026

PR #688, opened 2026-04-08
fix(docker): add compose project label to programmatically created containers

techyogi:
## Summary

- Containers created or updated via the admin UI (`/settings/apps`) were missing the `com.docker.compose.project: project-nomad` label
- Docker Compose sets this label automatically for compose-managed services, but containers created programmatically via dockerode don't inherit it
- Container orchestrators like OrbStack use this label to group containers — without it, updated/installed service containers appear outside the project group

## Changes

- Added `COMPOSE_PROJECT` static constant to `DockerService` for consistency
- Added `com.docker.compose.project` label to `_createContainer()` (covers new installs and force reinstalls)
- Added `com.docker.compose.project` label to `updateContainer()` (covers version updates), while also preserving any existing labels from the old container

## Test plan

- [ ] Install a new service via the admin UI → verify the container has `com.docker.compose.project: project-nomad` label (`docker inspect <container> | jq '.[0].Config.Labels'`)
- [ ] Update an existing service via `/settings/apps` → verify the updated container retains the compose project label and appears grouped with other Nomad containers
- [ ] Force reinstall a service → verify same label behavior
- [ ] On OrbStack: confirm updated containers appear under the `project-nomad` group instead of as standalone containers

### DIFF
--- a/admin/app/services/collection_manifest_service.ts
+++ b/admin/app/services/collection_manifest_service.ts
@@ -220,8 +220,13 @@ export class CollectionManifestService {
 
       for (const file of zimFiles) {
         console.log(`Processing ZIM file: ${file.name}`)
-        // Skip Wikipedia files (managed by WikipediaSelection model)
-        if (file.name.startsWith('wikipedia_en_')) continue
+        // Skip Wikipedia files managed by the WikipediaSelection model, but allow
+        // topic-specific Wikipedia ZIMs that are part of curated collection tiers
+        // (e.g. wikipedia_en_medicine_maxi).
+        if (file.name.startsWith('wikipedia_en_')) {
+          const parsed = CollectionManifestService.parseZimFilename(file.name)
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
Thanks for the continued work on this and for flagging the update-path gap. Taking a step back though: this is effectively a cosmetic grouping issue in Docker Desktop / OrbStack for a subset of users who inspect their NOMAD containers in those UIs *and* update services via the admin UI. It doesn't affect NOMAD's functionality, `docker compose` commands, container lifecycle management, or the vast majority of users who run NOMAD headless on Linux servers.

Given that no user has reported the grouping issue and we have a significant queue of real user-facing work ahead (AMD GPU support, ZIM embedding bug, remote AI integration gaps, Advanced Features, etc.), we're going to pass on this rather than invest review + testing cycles here. Appreciate the contribution pattern and please don't let this discourage you from future PRs on higher-impact areas — you've already shipped real improvements in v1.31.0.

PR #693, opened 2026-04-09
fix(rag): fall back to any available model for query rewriting

bravosierra99:
## Problem

When `qwen2.5:3b` is not installed — common in remote Ollama or LM Studio setups where a different model is configured — query rewriting is silently skipped entirely. The code logs a warning and falls back to the raw user message, meaning RAG retrieval quality degrades without any indication to the user.

The code even acknowledged this gap with a comment: `// FUTURE ENHANCEMENT: allow the user to specify which model to use for rewriting`.

Also, `getModels(true)` (include embeddings) was being used to find the rewrite model, meaning embedding models could theoretically be selected as the rewrite candidate — which would produce garbage output.

## Fix

- Falls back to the first available non-embedding model when `qwen2.5:3b` isn't installed
- Preferred model (`qwen2.5:3b`) is still used first when available — no behaviour change for default setups
- Extracts model selection into `selectRewriteModel()` in `app/utils/rag_utils.ts` — a pure function with unit tests that run without requiring the full AdonisJS/database stack

## Behaviour

| Scenario | Before | After |
|---|---|---|
| `qwen2.5:3b` installed | Uses `qwen2.5:3b` ✅ | Uses `qwen2.5:3b` ✅ |
| Only other model available | Skips rewriting ⚠️ | Falls back to available model ✅ |
| No models available | Skips rewriting | Skips rewriting (unchanged) |

## Testing

Unit tests added in `tests/unit/rag_utils.spec.ts` covering all three scenarios above plus edge cases. A standalone verification script (`tests/unit/rag_utils_verify.ts`) can be run without AdonisJS/Redis/MySQL:

```
npx tsx tests/unit/rag_utils_verify.ts
```

Tested on a live remote LM Studio setup — confirmed via logs that the fallback kicks in correctly and RAG retrieval proceeds with the available model.

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -10,6 +10,7 @@ import { inject } from '@adonisjs/core'
 import type { HttpContext } from '@adonisjs/core/http'
 import { DEFAULT_QUERY_REWRITE_MODEL, RAG_CONTEXT_LIMITS, SYSTEM_PROMPTS } from '../../constants/ollama.js'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
+import { selectRewriteModel } from '../utils/rag_utils.js'
 import logger from '@adonisjs/core/services/logger'
 type Message = { role: 'system' | 'user' | 'assistant'; content: string }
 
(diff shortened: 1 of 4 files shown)


chriscrosstalk:
Thanks for the thoughtful work here — pure function extraction plus unit tests plus the embeddings-list bug catch is exactly the kind of contribution pattern we like to see.

Unfortunately this conflicts directly with PR #649, which takes a different architectural approach to solving the same general problem. Rather than keeping `DEFAULT_QUERY_REWRITE_MODEL` as a preferred-with-fallback, #649 eliminates the concept entirely and uses the user's currently-selected chat model for query rewriting. That approach fixes a separately-reported bug where `qwen2.5:3b` was being force-loaded on every message even when the user manually unloaded it, saves memory (no separate rewrite model needs to stay resident), and avoids the "is the preferred model installed" problem by design.

#649 is already under review as a hotfix candidate, so we're going to go in that direction.

Closing this with genuine thanks — please don't let this discourage you from future contributions. The test discipline you brought here is valuable and there are plenty of other areas where it'd be welcome.

bravosierra99:
Understood!  Thank you!

Issue #367, opened 2026-03-16
[Feature]: Managing where Project Nomad uses storage in the Linux filesystem


as-kholin:
### Feature Category

System Management

### Problem Statement

Project Nomad does not inform the user where its storage will be used.  The locations it uses are not generally sized to handle its load.

It does not allow the user to define where the storage will be used.

Exacerbating this is the storage is actually used in 2 places, neither of which are generally sized to handle hundreds of GB

### Proposed Solution

Best solution - Ask the user during install where to place the data files.  Update the commands/script to put both the Kiwix downloads (currently in /opt/project-nomad/storage) in that location, and to override where Docker puts its storage for the containers to that location.  Doing so would ensure users put the files in a location they have allocated space for.

### Alternative Solutions

- Minimum viable solution - Tell the user where the files will be placed, and about how much to allocate for (hard when it depends on how the user then uses it
- Minimum recommended solution - Tell the user that the files will be placed in /var.   Warn them if /var is not sized to do this.  Update files currently saved to /opt/project-nomad/storage to save under /var (project Nomad's home directory, which can be put under /var?)
- Conventional - Define in the main page where files will be placed (for example - /mnt/project-nomad).  Tell them to put the storage to hold the files at that location.  Then update both Docker storage for project-nomad, as well as the /opt/project-nomad/storage locations to be under /mnt
- Most Flexible - main solution above - Ask the user during install where they should be placed, then put the files there, so that it can work with any partitioning scheme the user chooses to use.

### Who would benefit from this feature?

All Users

### How important is this feature to you?

High - Would significantly improve my experience

### Implementation Ideas (Optional)

Each solution above describes in implementation detail how to achieve it.

### Would you be willing to help implement this?

Yes - I can help test

### Pre-submission Checklist

- [x] I have searched for existing feature requests that might be similar
- [x] This feature aligns with N.O.M.A.D.'s mission of offline-first knowledge and education
- [x] I understand that feature requests are not guaranteed to be implemented

MaKeG0:
I'm very interested, as i can start editing the compose file but i would prefer to have an official solution.

archebyte:
[...]

PR #697, opened 2026-04-09
Add custom install location

flynnty:
- Add popup to install_nomad.sh for user entry of install location
 - Inject install location to relevant locations
 - Tested on fork https://github.com/flynnty/project-nomad
 - install/start/stop all work with no directory generated under /opt/project-nomad
 - Installed wiki works for browsing from custom install location
 - To test this PR install_nomad.sh/MANAGEMENT_COMPOSE_FILE_URL should be edited locally to point to /dev (this) branch
 - Works with install to external HDD


chriscrosstalk:
Thanks for taking a swing at this, and sorry for the delay getting to it.

Custom storage location is something we're already tracking internally as part of the Advanced Features work in #271 / #367, and we have a specific approach in mind that separates the in-app data path from the install directory — so users can keep NOMAD running from `/opt` but point downloads at a bigger disk. Because of that, we're going to pass on this PR rather than take it in its current form.

Really appreciate the effort and the thinking here, and please don't let this discourage you from future contributions — the interest and willingness to jump in is genuinely valued.

PR #700, opened 2026-04-10
fix(rag): add start button in kb modal and ensure restart policy exists

hestela:
Adds a check to RAG health to make sure nomad_qdrant is online, if not then the user will be blocked from clicking any buttons in the KB modal until they click the start qdrant button and let the container start

There is a new file qdrant_restart_policy_provider.ts, which tries to ensure that the restart policy always exists for the nomad_qdrant container even though the policy should have been there when the container is created.

### DIFF
--- a/admin/adonisrc.ts
+++ b/admin/adonisrc.ts
@@ -55,6 +55,7 @@ export default defineConfig({
     () => import('@adonisjs/transmit/transmit_provider'),
     () => import('#providers/map_static_provider'),
     () => import('#providers/kiwix_migration_provider'),
+    () => import('#providers/qdrant_restart_policy_provider'),
   ],
 
   /*
--- a/admin/app/controllers/rag_controller.ts
+++ b/admin/app/controllers/rag_controller.ts
@@ -97,4 +97,9 @@ export default class RagController {
       return response.status(500).json({ error: 'Error scanning and syncing storage' })
     }
   }
+
+  public async health({ response }: HttpContext) {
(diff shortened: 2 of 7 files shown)


chriscrosstalk (dismissed):
Good fix. Health check endpoint, UI feedback when Qdrant is down, and the restart policy provider to cover existing installs are all solid. The `qdrantInitPromise` reset on failure is a nice catch too -- without that, a transient Qdrant outage would permanently brick the KB until restart. Approved.

chriscrosstalk (approved):
Re-approving — GitHub auto-dismissed my 4/14 approval when you pushed more commits, not because of any concern on my end. The shape still matches what I approved, and the polish you added during iteration (offline-aware button disables, 3s-while-starting / 30s-steady polling cadence, `isStartingQdrant` reset via `useEffect`) all tightens up the UX nicely.

One small nit that's worth fixing in a follow-up (or this PR if it's easy):

Both user-facing error strings in `rag_service.ts` say "Restart the **AI Assistant** service in Settings to restore the Knowledge Base" — but the actual service that needs restarting is Qdrant, not Ollama. The KB modal handles this correctly via its dedicated "Start Qdrant" button, but anyone who sees the raw message somewhere else (API response, logs, a future chat-side surface) gets pointed at the wrong button. Suggest:

```ts
'Qdrant vector database is offline. Start the Knowledge Base (Qdrant) service to restore it.'
```

Not a blocker. Approved either way.

jakeaturner (approved):
Great stuff! JT 4/27/2026

PR #702, opened 2026-04-10
fix(admin): only hide global map banner after download

### LINKED ISSUE #633: Global map banner shows 'Available' even after download completes
## Problem

After successfully downloading the 126 GB global map file, the Maps Manager page still shows the "Global Map Coverage Available" banner with a "Download Global Map" button, as if it hasn't been downloaded. There is no visual indication that the global map is already on disk.

Clicking "Download Global Map" again says it starts successfully but doesn't actually re-download (the file already exists). This is confusing for users who just waited 15+ minutes for a 126 GB download.

## Steps to Reproduce

1. Navigate to Settings > Maps Manager
2. Click "Download Global Map" and wait for it to complete
3. Refresh the page
4. Banner still shows "Global Map Coverage Available" with download button

## Expected vs Actual

**Expected:** After the global map is downloaded, the banner should either disappear or change to indicate the global map is installed (e.g., "Global Map Installed" with build date and file size, plus an option to delete or re-download if a newer build is available).

**Actual:** Banner always shows regardless of download state. No feedback that the download completed.

## Root Cause

`maps.tsx` line 254 renders the banner whenever `globalMapInfo` is truthy (i.e., whenever the Protomaps builds API returns data). There is no cross-reference against `storedMapFiles` to check if the file is already on disk.

```tsx
// Current (line 254):
{globalMapInfo && (
  <Alert title="Global Map Coverage Available" ... />
)}

// Should be something like:
{globalMapInfo && !storedMapFiles.some(f => f.name === globalMapInfo.key) && (
  <Alert title="Global Map Coverage Available" ... />
)}
```

The `storedMapFiles` prop already contains all filenames on disk -- just needs a `.some()` check.

## Suggested Fix

1. Check if `globalMapInfo.key` (e.g., `20260402.pmtiles`) exists in `storedMapFiles`
2. If downloaded: show a success state ("Global Map Installed, build 2026-04-02") with delete option
3. If not downloaded: show current "Download Global Map" banner
4. If a newer build is available than what's on disk: show "Update Available" variant

## Files Involved

- `admin/inertia/pages/settings/maps.tsx` -- banner conditional (line 254)

Gujiassh:
## Summary
- keep the global map download banner visible until the configured global map file actually exists on disk
- add a small helper for checking the downloaded map inventory by filename
- add focused regression coverage for the banner visibility helper

## Testing
- npx tsx --test admin/tests/unit/global_map_banner.spec.ts

### DIFF
--- a/admin/inertia/lib/global_map_banner.ts
+++ b/admin/inertia/lib/global_map_banner.ts
@@ -1,0 +1,10 @@
+export function hasDownloadedGlobalMap(
+  globalMapKey: string | null | undefined,
+  storedMapFiles: Array<{ name: string }>
+): boolean {
+  if (!globalMapKey) {
+    return false
+  }
+
+  return storedMapFiles.some((file) => file.name === globalMapKey)
+}
--- a/admin/inertia/lib/global_map_banner.ts
+++ b/admin/inertia/lib/global_map_banner.ts
@@ -6,5 +6,5 @@ export function hasDownloadedGlobalMap(
     return false
   }
 
-  return storedMapFiles.some((file) => file.name === globalMapKey)
+  return storedMapFiles.some((file) => file.name === globalMapKey || /^\d{8}\.pmtiles$/.test(file.name))
 }
--- a/admin/inertia/pages/settings/maps.tsx
+++ b/admin/inertia/pages/settings/maps.tsx
@@ -253,6 +253,22 @@ export default function MapsManager(props: {
               }}
             />
           )}
+          {globalMapInfo && globalMapAlreadyDownloaded && (
+            <Alert
+              title="Global Map Installed"
+              message={`Your global map build ${globalMapInfo.date} (${formatBytes(globalMapInfo.size, 1)}) is stored locally and ready for offline use.`}
+              type="success"
+              variant="bordered"
+              className="mt-8"
+              icon="IconCircleCheck"
+              buttonProps={{
+                variant: 'secondary',
+                children: 'Download latest build',
+                icon: 'IconRefresh',
+                onClick: () => confirmGlobalMapDownload(),
+              }}
+            />
+          )}
           {globalMapInfo && !globalMapAlreadyDownloaded && (
             <Alert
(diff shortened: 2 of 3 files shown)


Gujiassh:
Quick validation from my side on the current branch tip:\n\n- npx tsx --test admin/tests/unit/global_map_banner.spec.ts\n- Result: passed\n\nThe fix still does the intended thing: the admin global-map banner now stays visible until the install/download step actually succeeds, instead of disappearing too early.

Gujiassh:
Refresh check on current main: branch is already up to date (no rebase delta).\n\nValidation attempt:\n- npm --prefix admin run test -- --files tests/unit/global_map_banner.spec.ts\n\nBlocked in this environment because app boot requires Redis and DB-backed runtime services (current runner has no Redis at 127.0.0.1:6379, causing startup failure before test execution).

chriscrosstalk (dismissed):
Clean fix, matches the approach we had in mind. Logic is straightforward, tests cover the cases. Approved.

jakeaturner (changes requested):
Great approach overall, but there's one caveat -- `api.getGlobalMapInfo()` returns the key of the _latest_ build from PMTiles, so if the user has any global maps file whose key does not match that, `globalMapAlreadyDownloaded` will be false and appear as though the user still needs to download it.

We should only really be checking that the user has _some_ global map file downloaded, as they are usually very large and not something most users will be updating frequently

Gujiassh:
Updated the existing branch to match the latest review note.

I changed the global-map banner check so it treats any stored Protomaps global build (`YYYYMMDD.pmtiles`) as already downloaded, rather than requiring an exact match with the latest PMTiles key returned by `getGlobalMapInfo()`. That keeps the banner hidden for users who already have an older full-world map on disk.

Updated head: `18250ca`

Focused verification:
- `cd admin && npm exec -- node --loader ts-node-maintained/esm --test tests/unit/global_map_banner.spec.ts`

jakeaturner (approved):
Great stuff, thanks!

PR #706, opened 2026-04-12
feat(ui): add file upload/download & open button in 'Content Manager'…

xaleel:
Added `Download` and `Open` buttons to the table in `Content Manager` page.
Also added an `Upload` button that uploads zim files to the storage directory and refreshes files and kiwix


chriscrosstalk (commented):
Thanks for the contribution -- upload/download/open in Content Manager is a good idea and something we want to support. However, this touches security-sensitive areas (file upload and arbitrary file download endpoints) that we need to build out carefully on our end with proper path traversal guards and size limits.

We're going to earmark this as a roadmap item and handle it internally. Closing for now, but appreciate the initiative -- the approach is directionally right.

PR #733, opened 2026-04-16
fix(kiwix): self-heal missing library XML

cuyua9:
## Summary
- validate the Kiwix library XML when Kiwix is already in library mode
- rebuild `/storage/zim/kiwix-library.xml` from disk when the file is missing or invalid
- keep the existing legacy migration path unchanged

## Validation
- cd admin && npm run typecheck
- cd admin && npx eslint providers/kiwix_migration_provider.ts app/services/kiwix_library_service.ts
- cd admin && npm run build

### DIFF
--- a/admin/app/services/kiwix_library_service.ts
+++ b/admin/app/services/kiwix_library_service.ts
@@ -187,6 +187,35 @@ export class KiwixLibraryService {
       .filter((b) => b.id && b.path)
   }
 
+  private _validateLibraryXml(xmlContent: string): void {
+    const parser = new XMLParser({
+      ignoreAttributes: false,
+      attributeNamePrefix: '@_',
+      isArray: (name) => name === 'book',
+    })
+
+    const parsed = parser.parse(xmlContent)
+    if (!parsed?.library || typeof parsed.library !== 'object') {
+      throw new Error('Kiwix library XML is missing the library root element.')
+    }
+  }
+
+  async ensureValidLibraryXml(): Promise<boolean> {
+    try {
+      const content = await readFile(this.getLibraryFilePath(), 'utf-8')
+      this._validateLibraryXml(content)
+      return false
+    } catch (err: any) {
+      if (err.code && err.code !== 'ENOENT') {
+        throw err
+      }
+
+      logger.warn('[KiwixLibraryService] Library XML is missing or invalid; rebuilding from disk.')
+      await this.rebuildFromDisk()
+      return true
+    }
+  }
+
   async rebuildFromDisk(opts?: { excludeFilenames?: string[] }): Promise<void> {
     const dirPath = join(process.cwd(), ZIM_STORAGE_PATH)
     await ensureDirectoryExists(dirPath)
--- a/admin/providers/kiwix_migration_provider.ts
+++ b/admin/providers/kiwix_migration_provider.ts
@@ -22,6 +22,7 @@ export default class KiwixMigrationProvider {
         const Service = (await import('#models/service')).default
         const { SERVICE_NAMES } = await import('../constants/service_names.js')
         const { DockerService } = await import('#services/docker_service')
+        const { KiwixLibraryService } = await import('#services/kiwix_library_service')
 
         const kiwixService = await Service.query()
           .where('service_name', SERVICE_NAMES.KIWIX)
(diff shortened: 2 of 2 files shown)


chriscrosstalk (changes requested):
Thanks @cuyua9 — this is a real defensive gap worth closing. When `kiwix-library.xml` goes missing (upgrade glitch, manual wipe, third-party forks that don't follow the same migration path), Kiwix silently serves an empty library and users get confused. Rebuilding from disk on startup is exactly the right instinct, and delegating to the existing `rebuildFromDisk()` keeps the blast radius small.

I retargeted the PR from `main` to `dev` — we base all fixes on `dev` so they flow into the next release cycle. No action needed from you there.

Two things I'd like to see tightened before merge:

**1. Drop the cosmetic refactors that aren't part of the fix.** The diff has a handful of unrelated style changes piggybacking on the real change:

- `'fs/promises'` → `'node:fs/promises'` and `'path'` → `'node:path'` — these aren't consistent with the rest of the codebase, which uses the unprefixed forms everywhere.
- `const Service = (await import('#models/service')).default` unwrapped into two lines
- `Service.query().where(...)` split into `const query = ...; await query.where(...)`
- A `logger.info(...)` reformatted from one line to multi-line

None of those change behavior, and they make the diff noisier + hurt git blame for the surrounding lines. Can you revert them so the PR contains only the self-heal logic?

**2. Simplify the catch block.** In `ensureValidLibraryXml`:

```ts
if (err.code && err.code !== 'ENOENT') {
  throw err
}
if (!err.code && !(err instanceof Error)) {
  throw err
}
```

The second branch is dead code — everything thrown through here is a standard `Error` (either from node's `readFile` or from `_validateLibraryXml`'s `new Error(...)`). You can drop to:

```ts
if (err.code && err.code !== 'ENOENT') throw err
logger.warn(...)
await this.rebuildFromDisk()
return true
```

Happy to approve once those are cleaned up.

huyua9:
Thanks for the precise review. I pushed `5b785e62` to keep this PR focused on the self-heal fix.

Changes in this follow-up:
- reverted the unrelated import style churn in `kiwix_library_service.ts` back to the repository's unprefixed `fs/promises` / `path` style
- restored the existing multiline `articleCount` formatting
- restored the compact dynamic `Service` import and direct `Service.query().where(...).first()` chain in the migration provider
- restored the pre-existing single-line migration `logger.info(...)`
- simplified the `ensureValidLibraryXml()` catch block by removing the dead non-`Error` branch; it now only rethrows non-`ENOENT` filesystem errors before rebuilding missing/invalid library XML

Validation:
- `git diff --check origin/dev...HEAD`: passed
- attempted `npm ci` in `admin/`, but the local macOS toolchain cannot build `@openzim/libzim`: `xcrun` fails to load `/Library/Developer/CommandLineTools/usr/lib/libxcrun.dylib` because the installed CommandLineTools library is `x86_64` while this runtime needs `arm64e`. So I could not get local dependencies installed for `npm run typecheck` in this environment.

cuyua9:
Thanks again for the review. I pushed `ac875952` as a final cleanup follow-up.

This keeps the already-addressed scope cleanup in place and additionally avoids reformatting the existing "already in library mode" `logger.info(...)` call into a multiline conditional expression. The non-legacy path now logs the rebuild case separately, while the pre-existing no-op message remains as the original single-line call.

Validation:
- `git diff --check origin/dev...HEAD`: passed
- submission identity rechecked for `cuyua9`: passed

chriscrosstalk (dismissed):
Approve. Both prior review points are addressed cleanly: no cosmetic refactors riding along, and the catch block is now the simplified version. Final diff is exactly the self-heal logic and nothing else.

Behavior reads correctly:
- Missing file → ENOENT → rebuild
- Parse failure or missing `<library>` root → caught Error → rebuild
- Permission / IO errors → re-thrown, surface to logs (don't mask real problems)
- Successful validation → no-op

Thanks @cuyua9 for the patience on the cleanup pass.

cuyua9:
Rebased this approved PR onto the latest upstream `dev` to retrigger CI after the previous build failed while downloading `libzim_linux-x86_64-9.4.0.tar.gz` with HTTP 503.\n\nNew head: `d31282f8`\n\nValidation performed locally:\n- `git diff --check origin/dev...HEAD`\n\nScope audit: `origin/dev...HEAD` remains limited to:\n- `admin/app/services/kiwix_library_service.ts`\n- `admin/providers/kiwix_migration_provider.ts`\n\nI did not run the admin npm test/build locally because this checkout has no `node_modules`, and the previous failing GitHub job failed during dependency install rather than in the PR code.

chriscrosstalk:
Thanks @cuyua9 — this is a good idea. Auto-recovering a missing or corrupt Kiwix library XML at startup closes a real gap (today it's only rebuilt on the download path). We've implemented startup self-heal in #986, building on this idea — it separates the missing-vs-invalid cases, avoids rebuilding a well-formed-but-empty library, and surfaces genuine filesystem errors rather than rebuilding on them. Closing in favor of that. Appreciate you flagging it.

PR #740, opened 2026-04-17
fix(startup): non-blocking filesystem reconciliation

### LINKED ISSUE #739: [Bug]: Server unresponsive during startup filesystem reconciliation
### Bug Category

Performance

### Description

On startup, `reconcileFromFilesystem()` in `server.ts` sequentially stats every ZIM and map file before the server can respond to HTTP requests. With 50+ content files this blocks the Node.js event loop for 5-30+ seconds, during which the health endpoint and all page loads hang.

### Steps to Reproduce

1. Install N.O.M.A.D. with all ZIM files downloaded
2. Restart the admin container (`docker compose restart admin`)
3. Immediately try to load the web UI or curl the health endpoint

### Expected Behavior

The server should respond to HTTP requests within a few hundred milliseconds of starting, regardless of how many content files exist.

### Actual Behavior

The server accepts TCP connections but doesn't respond to HTTP requests for 5-30+ seconds while `reconcileFromFilesystem()` runs. The duration scales linearly with the number of ZIM + map files, and is significantly worse on slower storage (e.g., network-attached storage, USB drives, SD cards).

### Root Cause

In `admin/bin/server.ts`, the `app.ready()` callback `await`s `reconcileFromFilesystem()`. Although the HTTP server is already listening at this point, Node.js is single-threaded — the sequential `stat()` + `updateOrCreate()` loop for each file monopolizes the event loop, starving HTTP request handlers.

### Environment

- N.O.M.A.D. version: 1.31.0
- OS: Ubuntu 24.04 / Docker Desktop
- Storage: Local SSD + NAS (SMB)
- Content files: 54 ZIM files, 20+ map files

### Screenshots/Logs

Health check response times during startup (original):
```
Check 2: 5.149s
Check 14: 1.966s
Check 16: 1.379s
Check 18: 1.095s
```

ZainRizvi:
## Summary

Fixes #739 — server unresponsive for 5-30+ seconds during startup while `reconcileFromFilesystem()` sequentially stats every content file.

- **Fire-and-forget reconciliation**: Removed `await` from the `reconcileFromFilesystem()` call in `server.ts` so the HTTP server responds immediately while reconciliation runs in the background
- **Parallel batched I/O**: Replaced sequential `stat()` + `updateOrCreate()` loops with `Promise.allSettled` batches of 10, reducing wall-clock reconciliation time ~10x
- **Correct seen-ID tracking**: Resource IDs are only added to the "seen" set after a successful DB write, so files that fail to process are correctly handled by the cleanup phase

### Why removing `await` is safe

`reconcileFromFilesystem()` syncs the `installed_resources` DB table with what's on disk. No controller or service depends on this being complete before serving requests:

- **ZIM/map list endpoints** (`ZimController.list`, `MapsController.listRegions`) read directly from the filesystem, not from `InstalledResource`
- **Category status endpoints** (`getCategoriesWithStatus`, `getMapCollectionsWithStatus`) query `InstalledResource` — if reconciliation hasn't finished, they return status from the previous run (stale but valid, self-corrects within seconds)
- **Download jobs** use `InstalledResource` to skip already-downloaded files — worst case, a file gets re-downloaded (harmless, download is idempotent)
- **`updateOrCreate` is atomic per-row** — partial reconciliation state is safe (some entries updated, others stale)
- **Error handling is preserved** — `reconcileFromFilesystem()` has internal try/catch blocks for both ZIM and map phases, so it never produces an unhandled rejection

### Why `Promise.allSettled` instead of `Promise.all`

Individual file failures (e.g., file deleted between directory listing and stat) shouldn't abort the entire batch. `Promise.allSettled` lets the rest of the batch complete, and failures are logged individually.

## Test Plan

- [x] Health endpoint responds in <100ms during reconciliation (was 5-30s)
- [x] `installed_resources` table correctly populated after reconciliation completes (68 entries, matching previous count)
- [x] Files that fail to process don't prevent other files from being processed
- [x] Cleanup phase correctly removes entries for files no longer on disk
- [x] No race condition between reconciliation and concurrent download completion (both use `updateOrCreate` with same key — last writer wins, both write correct data)


chriscrosstalk:
Thanks @ZainRizvi, the diagnosis and reference implementation are sharp. Closing alongside #739: the impact on standard installs is mild (~2.2s cold start on NOMAD3 with 50 ZIMs, 20 maps), and the longer hangs you observed correlate with Docker Desktop + NAS/SMB which aren't supported configurations. Appreciate the contribution.

Issue #662, opened 2026-04-05
[Feature]: AI feature : Remove nomad_ollama as a requirement for AI Assistant install.


Draco-Lunaris:
### Feature Category

AI Assistant Enhancement

### Problem Statement

 If you have a complex GPU config or require customization requirements on the host that Nomad is to be installed on, the AI assistant install requires the ollama docker container to be installed. This conflicts with other LLM hosting software running on the host including customized ollama configs, vLLM, llama.cpp, etc. 

 The AI assistant install will fail if there is anything already running on the default ollama port. 
The AI assistant will have random issues and will cause random issues for other LLM services by trying to utilize the same GPU's at the same time. 

### Proposed Solution

Possible, solution. Remove the ollama requirement for the AI Assistant install. If the ollama service is optional, then it won't be present to conflict for ports or GPU resources. 

### Alternative Solutions

Run the nomad_ollama container as default CPU only, and on a non-standard port to avoid conflicting with services that may already be in place. 

It is possible to run Nomad on a totally different host just fine. IF you have additional hardware, and are willing to was the resources for it. 

### Use Case

 This would allow more complex LLM configuration and deployments on the same host nomad is installed on. Allowing for more functionality outside of nomad on the same hardware. 

 

### Who would benefit from this feature?

All Users

### How important is this feature to you?

High - Would significantly improve my experience

### Examples or References

Example, I utilize a single host machine to run Openclaw, Agent Zero, Nomad, vLLM, and ollama. All my GPU's and storage/memory are all on a single platform and run inside docker containers. 

This allows me to fully utilize the hardware I have. Additional hardware would require more cost, more electricity, more cooling, etc. 

### Would you be willing to help implement this?

Yes - I can write the code

### Additional Context

 I am willing to help with a solution. I'm fully capable of coding and solution engineering. My GitHub experience is limited however. I have room to learn how to properly work with it to contribute correctly. 

### Pre-submission Checklist

- [x] I have searched for existing feature requests that might be similar
- [x] This feature aligns with N.O.M.A.D.'s mission of offline-first knowledge and education
- [x] I understand that feature requests are not guaranteed to be implemented

hestela:
[...]

PR #744, opened 2026-04-18
fix(ai): stop local nomad_ollama when remote Ollama is configured

chriscrosstalk:
## Summary

When users configure a remote Ollama URL via AI Settings, the local `nomad_ollama` container kept running and competed with the remote host for port 11434 and GPU access. Per community reports on #662 (@kwslavens, @fjarvis, @recklessnode, @hestela), this causes random failures when integrating NOMAD with external LLM stacks (vLLM, llama.cpp, Agent Zero, openclaw, custom Ollama).

`configureRemote` now:

- **Sets URL**: after the remote URL is validated and saved, stops the local `nomad_ollama` container if it's running. Container and volume are preserved (models intact).
- **Clears URL**: if a local `nomad_ollama` container still exists, starts it back up and keeps `installed: true`. If no local container exists, falls back to the previous `installed: false` behavior.

Both helpers are non-fatal on Docker errors so a stuck container doesn't block the remote save/clear.

## Why the scope stops here

Related pain points flagged in #662:
- **Easy-setup skip when remote enabled** — already shipped in v1.31.0 (commit 69c15b8 filters OLLAMA out of `selectedServices`).
- **AI Assistant update fails when container is stopped** (per @fjarvis) — separate bug, deserves its own PR.

## Test plan

- [ ] Fresh install with local AI Assistant, set remote URL via Settings → Models, verify `nomad_ollama` transitions to `Exited` in `docker ps -a`
- [ ] Models volume intact (`docker volume inspect nomad_ollama_data`)
- [ ] Clear remote URL, verify `nomad_ollama` comes back to `Up` and service stays `installed`
- [ ] Set remote URL on an install that never had local Ollama — no error, no-op stop
- [ ] Clear remote URL on the same — falls back to `installed: false` as before
- [ ] Chat + Knowledge Base work against remote host the whole time

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -212,13 +212,21 @@ export default class OllamaController {
       return response.status(404).send({ success: false, message: 'Ollama service record not found.' })
     }
 
-    // Clear path: null or empty URL removes remote config and marks service as not installed
+    // Clear path: null or empty URL removes remote config. If a local nomad_ollama container
+    // still exists (user had previously installed AI Assistant locally), restart it and keep
+    // the service marked installed. Otherwise fall back to uninstalled.
     if (!remoteUrl || remoteUrl.trim() === '') {
       await KVStore.clearValue('ai.remoteOllamaUrl')
-      ollamaService.installed = false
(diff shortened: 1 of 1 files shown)


jakeaturner (approved):
JT 4/20/2026

PR #746, opened 2026-04-18
fix(zim): accumulate across Kiwix pages to prevent empty Content Explorer

### LINKED ISSUE #731: Content Explorer shows empty when many ZIM files are installed locally
## Bug Description

The "Browse the Kiwix Library" section on the Content Explorer page (`/settings/zim/remote-explorer`) shows "No records found" when a significant number of ZIM files are already installed locally.

## Root Cause

The `listRemote` method in `zim_service.js` fetches a page of results from the Kiwix catalog API (`browse.library.kiwix.org/catalog/v2/entries`), then **client-side filters out** any files that already exist locally on disk. With ~435 ZIM files installed out of ~1093 available, entire pages of results (12 items per page) come back fully filtered to zero items.

The frontend uses infinite scroll that triggers on the table's scroll event — but since the table is empty (0 rows), there's nothing to scroll, so it never fetches subsequent pages where non-installed content would eventually appear.

**With a search query**, the result set is smaller and more targeted, so the filtering doesn't eliminate entire pages — which is why search works but the default view doesn't.

## Steps to Reproduce

1. Install a large number of ZIM files (400+)
2. Navigate to Content Explorer (`/settings/zim/remote-explorer`)
3. Observe "No records found" in the "Browse the Kiwix Library" table
4. Type any search term — results appear normally

## Expected Behavior

The library browser should show available (non-installed) content regardless of how many files are already installed.

## Suggested Fix

Move the installed-file filtering server-side in `listRemote()`. Instead of fetching one page from Kiwix and filtering it, fetch in larger batches and keep scanning until enough non-installed items are collected to fill the requested page size:

```js
async listRemote({ start, count, query }) {
    const LIBRARY_BASE_URL = 'https://browse.library.kiwix.org/catalog/v2/entries';
    const existing = await this.list();
    const existingKeys = new Set(existing.files.map((file) => file.name));
    const parser = new XMLParser({ ignoreAttributes: false, attributeNamePrefix: '', textNodeName: '#text' });

    const items = [];
    let kiwixOffset = 0;
    let totalResults = 0;
    const BATCH_SIZE = 50;
    const MAX_BATCHES = 30;
    let nonInstalledSkipped = 0;

    for (let batch = 0; batch < MAX_BATCHES && items.length < count; batch++) {
        const res = await axios.get(LIBRARY_BASE_URL, {
            params: { start: kiwixOffset, count: BATCH_SIZE, lang: 'eng', ...(query ? { q: query } : {}) },
            responseType: 'text',
        });
        const result = parser.parse(res.data);
        if (!isRawListRemoteZimFilesResponse(result)) break;
        totalResults = result.feed.totalResults;
        const entries = result.feed.entry
            ? Array.isArray(result.feed.entry) ? result.feed.entry : [result.feed.entry]
            : [];
        if (entries.length === 0) break;

        for (const entry of entries) {
            if (!isRawRemoteZimFileEntry(entry)) continue;
            const links = Array.isArray(entry.link) ? entry.link : [entry.link];
            const downloadLink = links.find((link) =>
                typeof link === 'object' && 'rel' in link && 'length' in link &&
                'href' in link && 'type' in link && link.type === 'application/x-zim'
            );
            if (!downloadLink) continue;
            const download_url = downloadLink['href'].substring(0, downloadLink['href'].length - 6);
            const file_name = download_url.split('/').pop() || `${entry.title}.zim`;
            if (existingKeys.has(file_name)) continue;
            if (nonInstalledSkipped < start) { nonInstalledSkipped++; continue; }
            if (items.length < count) {
                const sizeBytes = parseInt(downloadLink['length'], 10);
                items.push({
                    id: entry.id, title: entry.title, updated: entry.updated,
                    summary: entry.summary, size_bytes: sizeBytes || 0,
                    download_url, author: entry.author.name, file_name,
                });
            }
            if (items.length >= count) break;
        }
        kiwixOffset += BATCH_SIZE;
        if (kiwixOffset >= totalResults) break;
    }

    return {
        items,
        has_more: kiwixOffset < totalResults && items.length === count,
        total_count: totalResults,
    };
}
```

This fetches in batches of 50 from the Kiwix API, skips installed files, and accumulates until the requested `count` is met. Tested and confirmed working.

## Environment

- Project N.O.M.A.D. v1.31.0
- ~435 ZIM files installed locally
- Kiwix catalog returns ~1093 English-language entries

chriscrosstalk:
## Summary

Closes #731 (reported + diagnosed by @johno10661).

When a user has many ZIMs already installed (~400 of ~1093), the \"Browse the Kiwix Library\" view showed \"No records found.\" The Kiwix catalog returned 12 items per page; \`zim_service.listRemote\` filtered out already-installed entries; when every item on a page was already installed the response was \`items: []\` with \`has_more: true\`; and the frontend's infinite-scroll guard (\`flatData.length > 0\`) blocked the next fetch, so the user was stuck.

Search worked because targeted queries rarely produce a full-page-of-installed collision.

### Backend — \`admin/app/services/zim_service.ts\`

\`listRemote\` now accumulates across upstream pages before returning. Up to 5 Kiwix fetches per request, 60 items each. Terminates early when:
- accumulated post-filter results reach the requested \`count\`
- Kiwix returns an empty page (transient hiccup / totalResults drift)
- \`currentStart\` crosses \`totalResults\`

\`currentStart\` advances by the actual returned entry count, not the requested count — so short tail-end pages don't skip entries. Dedupe set keyed on entry \`id\` guards against upstream paging races. The installed-file snapshot (\`this.list()\`) is taken once before the loop rather than per iteration. Pre-existing \`has_more\` off-by-one (comparing totalResults to the INPUT start) is fixed implicitly — new code compares against post-advancement \`currentStart\`.

Response shape adds \`next_start: number\` — the Kiwix offset at which to resume on the next request.

### Types — \`admin/types/zim.ts\`

\`ListRemoteZimFilesResponse\` gains \`next_start: number\`. No other consumers of this type needed updates.

### Frontend — \`admin/inertia/pages/settings/zim/remote-explorer.tsx\`

- \`useInfiniteQuery\` now uses the backend's \`next_start\` cursor as \`pageParam\` instead of computing \`pageParam * 12\` locally. That Kiwix-offset math was always a leaky abstraction; it's outright broken once the backend does multi-page accumulation.
- \`getNextPageParam: (lastPage) => lastPage.has_more ? lastPage.next_start : undefined\`.
- \`fetchOnBottomReached\` drops the \`flatData.length > 0\` guard. The existing on-mount effect (\`useEffect(() => fetchOnBottomReached(...), [fetchOnBottomReached])\`) drives bounded auto-fetch when a short page lands, correctly terminating when \`hasMore\` flips false.

## Test plan (NOMAD3, 192.168.200.183)

- [ ] Before patch: Content Explorer → Browse the Kiwix Library shows \"No records found\" (or reproduce by adding dummy \`.zim\` files to \`/opt/project-nomad/storage/zim\` to saturate the filter).
- [ ] After patch deployed: Browse view loads a scrollable list of non-installed ZIMs on first request.
- [ ] Network tab: initial \`GET /api/zim/list-remote?start=0&count=12\` returns ≥12 items plus \`next_start\`, \`has_more: true\`.
- [ ] Scrolling down sends next request with \`start={next_start from previous response}\` — not \`start=12\` — and continues paginating.
- [ ] Search (\"wikipedia\") still works — targeted path unchanged.
- [ ] Clear search — browse view still loads.
- [ ] With very high installed count, a short page returns \`has_more: true\` and the frontend auto-continues via the on-mount effect until the viewport fills or \`has_more\` flips.
- [ ] Download a ZIM from the list; confirm it's filtered out on the next \`listRemote\` refresh.

## Risks

- **Latency:** up to 5 × ~300ms Kiwix fetches per request in the worst case (~1.5s). Cap is deliberate — subsequent scrolls amortize cost.
- **Pagination contract:** only caller of \`listRemoteZimFiles\` is the Content Explorer page; no other consumers.
- **TanStack v5 cursor:** \`initialPageParam: 0\` + numeric cursor is the documented infinite-query pattern. Query key includes \`query\` so search resets correctly.

### DIFF
--- a/admin/app/services/zim_service.ts
+++ b/admin/app/services/zim_service.ts
@@ -57,84 +57,105 @@ export class ZimService {
     query?: string
   }): Promise<ListRemoteZimFilesResponse> {
     const LIBRARY_BASE_URL = 'https://browse.library.kiwix.org/catalog/v2/entries'
+    // Kiwix returns pages of content unaware of what the user has installed locally. When
+    // the installed set is large, a single 12-item Kiwix page can come back with everything
+    // already installed → 0 post-filter items → frontend deadlock (#731). Accumulate across
+    // upstream pages so we return a useful batch. Bounded by MAX_KIWIX_FETCHES so a heavily
+    // saturated install doesn't hang a single request; the frontend scroll loop + auto-fetch
+    // effect handle continuation.
+    const KIWIX_PAGE_SIZE = 60
+    const MAX_KIWIX_FETCHES = 5
 
-    const res = await axios.get(LIBRARY_BASE_URL, {
-      params: {
-        start: start,
-        count: count,
-        lang: 'eng',
-        ...(query ? { q: query } : {}),
-      },
-      responseType: 'text',
-    })
-
-    const data = res.data
     const parser = new XMLParser({
       ignoreAttributes: false,
       attributeNamePrefix: '',
       textNodeName: '#text',
     })
-    const result = parser.parse(data)
 
-    if (!isRawListRemoteZimFilesResponse(result)) {
-      throw new Error('Invalid response format from remote library')
-    }
-
-    const entries = result.feed.entry
-      ? Array.isArray(result.feed.entry)
-        ? result.feed.entry
-        : [result.feed.entry]
-      : []
-
-    const filtered = entries.filter((entry: any) => {
-      return isRawRemoteZimFileEntry(entry)
-    })
-
-    const mapped: (RemoteZimFileEntry | null)[] = filtered.map((entry: RawRemoteZimFileEntry) => {
-      const downloadLink = entry.link.find((link: any) => {
-        return (
-          typeof link === 'object' &&
-          'rel' in link &&
-          'length' in link &&
-          'href' in link &&
-          'type' in link &&
-          link.type === 'application/x-zim'
-        )
-      })
-
-      if (!downloadLink) {
-        return null
-      }
-
-      // downloadLink['href'] will end with .meta4, we need to remove that to get the actual download URL
-      const download_url = downloadLink['href'].substring(0, downloadLink['href'].length - 6)
-      const file_name = download_url.split('/').pop() || `${entry.title}.zim`
-      const sizeBytes = parseInt(downloadLink['length'], 10)
-
-      return {
-        id: entry.id,
-        title: entry.title,
-        updated: entry.updated,
-        summary: entry.summary,
-        size_bytes: sizeBytes || 0,
-        download_url: download_url,
-        author: entry.author.name,
-        file_name: file_name,
-      }
-    })
-
-    // Filter out any null entries (those without a valid download link)
-    // or files that already exist in the local storage
+    // Snapshot locally-installed files once — the filesystem won't change mid-request.
     const existing = await this.list()
     const existingKeys = new Set(existing.files.map((file) => file.name))
-    const withoutExisting = mapped.filter(
-      (entry): entry is RemoteZimFileEntry => entry !== null && !existingKeys.has(entry.file_name)
-    )
+
+    const accumulated: RemoteZimFileEntry[] = []
+    const seenIds = new Set<string>()
+    let currentStart = start
+    let totalResults = 0
+
+    for (let i = 0; i < MAX_KIWIX_FETCHES; i++) {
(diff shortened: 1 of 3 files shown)


jakeaturner (approved):
JT 4/20/2026

Issue #718, opened 2026-04-14
[Bug]: scanAndSyncStorage queues non-embeddable files — kiwix-library.xml loops forever


sbruschke:
### Issue Category

Knowledge Base/RAG (Document Upload)

### Bug Description

The Knowledge Base "sync storage" scan queues *every* file under `/storage/zim/` for embedding without checking file type. Kiwix's generated library metadata file `kiwix-library.xml` ends up queued on every sync, fails the `determineFileType()` check inside `EmbedFileJob` with `"Unsupported file type"`, and retries 30× per dispatch (per the default `attempts: 30` + 60s `backoff`). Every sync re-queues it, so the failure loop is effectively permanent and floods `nomad_admin` logs.

### Steps to Reproduce

1. Install Information Library (Kiwix) — this produces `/app/storage/zim/kiwix-library.xml`.
2. Click the "Sync Storage" button in Knowledge Base settings (or POST `/api/rag/sync`).
3. `docker logs -f nomad_admin` — observe the repeating loop:
   ```
   [ info ] [file-embeddings] Processing job: 7d9aa7fc644fa8e0 of type: embed-file
   [ error ] [file-embeddings] Job failed: 7d9aa7fc644fa8e0, Error: Unsupported file type.
   ```
4. Confirm via `GET /api/rag/active-jobs` — `kiwix-library.xml` is permanently in the failed/retrying state.

### Expected Behavior

Non-embeddable files (XML, JSON, font files, etc. — anything `determineFileType()` returns `'unknown'` for) should not be queued in the first place.

### Actual Behavior

All files under `/app/storage/zim/` get queued and fail 30 times each.

### Root Cause

`admin/app/services/rag_service.ts` (`scanAndSyncStorage`, ~line 1246):
```ts
const filesToEmbed = filesInStorage.filter((filePath) => !sourcesInQdrant.has(filePath))
```
Only filters by "not already in Qdrant" — never calls `determineFileType()`.

### Suggested Fix

Apply `determineFileType` check before dispatch:
```ts
import { determineFileType } from '../utils/fs.js'
// ...
const filesToEmbed = filesInStorage.filter(
  (filePath) => !sourcesInQdrant.has(filePath) && determineFileType(filePath) !== 'unknown'
)
```
Confirmed working against NOMAD 1.31.0.

### N.O.M.A.D. Version

1.31.0

### Operating System

Other (not yet officially supported)

### Docker Version

[...]

PR #751, opened 2026-04-20
feat(zim): add zim uploader in content manager

hestela:
adds a collapsible file uploader to accept zim file uploads into kiwix.

### DIFF
--- a/admin/app/controllers/zim_controller.ts
+++ b/admin/app/controllers/zim_controller.ts
@@ -8,7 +8,12 @@ import {
 } from '#validators/common'
 import { addCustomLibraryValidator, browseLibraryValidator, idParamValidator, listRemoteZimValidator } from '#validators/zim'
 import { inject } from '@adonisjs/core'
+import logger from '@adonisjs/core/services/logger'
 import type { HttpContext } from '@adonisjs/core/http'
+import { createWriteStream } from 'fs'
+import { rename } from 'fs/promises'
+import { join, resolve, sep } from 'path'
+import { ZIM_STORAGE_PATH, ensureDirectoryExists, sanitizeFilename } from '../utils/fs.js'
 
 @inject()
 export default class ZimController {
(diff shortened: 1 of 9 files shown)


hestela:
i've only tested a single file upload of a ~800MB file and it kinda just worked and showed up in the Information Library and content manager. needs more testing and it put in text about the file size limit being 80GB but i think thats probably not going to work well for an upload unless its on the same machine maybe. my knowledge base ingestion is broken, i think same issue as #718

chriscrosstalk:
Nice work — genuinely useful feature, and the shape is right. Security is clean (path traversal check, sanitized filename, atomic `.tmp` rename, file-type guard all matching our established patterns). The `processManually` bodyparser bypass is the right call for large-file streaming.

Two small things I'd tighten before merge:

1. **The 80 GB advertised limit is optimistic with XHR upload.** `@uppy/xhr-upload` sends the whole file in a single POST — one TCP blip and the upload starts over from zero. Would suggest either (a) dropping the advertised limit to ~20 GB and adding a note like *"For best results, upload from the same machine or over a stable LAN connection. Larger files should be copied directly to the storage volume"*, or (b) swapping in `@uppy/tus` for resumable uploads (bigger change, probably a follow-up PR).

2. **`response.status(500).send({ message: 'Upload failed', error: error.message })` leaks the raw error string.** Same cleanup pattern as the `system_update_service.getUpdateLogs` nit from the rc.1/rc.2 QA reports — log the error internally, return a generic message to the client.

Re: testing — your KB ingestion issue is almost certainly the same `num_ctx=2048` bug that bites dense content (same as #388/#756). PR #763 (awaiting Jake's review) fixes the root cause in `OllamaService.embed()`. Once that lands you should be able to validate the embed-job dispatch path.

For the testing matrix before merge, I'd want to see:

- Large file (1+ GB) from a different LAN client (not same-machine)
- Upload while a remote ZIM download is in flight (does the kiwix restart correctly defer?)
- Upload of a `.zim` whose filename already exists locally (does `rename()` clobber or throw?)
- Path traversal attempt + non-`.zim` attempt (security regression check)

Happy to run any of those on NOMAD3 once you feel it's ready for external testing.

hestela:
Addressed comments. 
working on some bug fixes, some related to the tests you described. Trying to make it so that if you upload one of the known zim files, it marks it as completed otherwise it wont be marked as completed unless you re-download using nomad. 
WIP

hestela:
ok tested/fixed a few things, should be mostly good now, see bug info below.
heres how the uploader behaves:
- for wikipedia only, if you upload a bigger or smaller zim, then the old one will get deleted and the new one will be the one marked as installed. if you delete the zim, then wikipedia gets updated to not installed. this is how the managed downloader works.
- Duplicate file names are rejected. ie, if i already have  devdocs_en_node_2026-02.zim then i get rejected if i try to upload it again.
- If you upload say 3/5 of "Computing & Technology" Essential, then it still says its not installed but if you click to install Essential, then only the missing 2 files are downloaded. similar if you download Standard. Similarly, if you upload 3/5 and then later upload the missing 2 files, the collection is marked as installed.
- deleting a zim file from a collection marks it as no longer installed. but if you go in and upload the missing one(s) then its marked as installed.
- 5 file uploads are allowed (hard coded limit, could be changed). uploads done in serial. 2GB on same network (different hosts) takes ~15. from a sata 3 drive to a zfs with 4 sata ssds and 10G network.

bug, not sure how to handle this one:
start a download of 2GB wikipedia for example, it will let you upload the same file on the side and then the downloaded file will just overwrite the uploaded one. i asked claude to fix this but it suggested just blocking all uploads while any download is in progress.
Before i tried that fix, it would let me upload other zim files during a download without issue, so i could see it either way. 
question is: should we only allow 1 upload/download at a time or let the user go crazy? maybe could get it to only block suspected dupes. otherwise this should be good for more testing. its otherwise pretty solid besides this bug. i didnt push the fix for this bug in the latest force push

hestela:
resolved conflict w/force-push. did a quick test by deleting and then re-uploading the same zim file

chriscrosstalk:
Go ahead and build in the inability to upload anything when there are any current downloads happening.  I think that's a good remediation to any issues.

hestela:
@chriscrosstalk , done. i just tested it. i ran a download in Content Explorer and while that happened i repeatedly uploaded the ~300MB and then the 2GB wikipedia and it didn't care it let me keep switching between the two on the side. but even if you were to do an upload/download of same name, then the latest one will get to be the "winner" ill say.

chriscrosstalk:
Thanks @hestela, this is still a feature we want. Two things before it can move forward:

1. The branch has drifted out of sync and now conflicts with `dev`. Could you rebase onto current `dev`?
2. Since you opened this, Kiwix gained native library-mode plus a "Rescan Library" button (#967). The uploader should slot into that flow rather than manage the library separately, so it's worth reconciling against that to avoid duplicating the rescan/library handling.

Once it's rebased and reconciled we'll give it a full review.

hestela:
@chriscrosstalk , addressed your comments and rebased. tested by uploading a few different wikipedia tiers. and that logic still works (where you can switch tiers by uploading a different zim)

jakeaturner (approved):
Great feature, thanks!

Issue #756, opened 2026-04-21
[Bug]: Knowledge Base: large PDFs fail during embedding, and Sync Storage retries unsupported files from mixed storage tree


NC4WD:
### Issue Category

AI Assistant (Ollama)

### Version
Project N.O.M.A.D. v1.31

### Summary
I am seeing two Knowledge Base problems that appear related to ingestion/storage scanning:

1. Small `.txt` files can complete successfully.
2. Larger `.pdf` files fail during embedding.
3. Clicking **Sync Storage** repeatedly triggers unsupported-file errors.

This happens across multiple devices, so it does not appear to be a client upload issue.

### What I expected
- Large PDFs should be chunked and embedded successfully, or fail gracefully with a clear document-specific error.
- Sync Storage should only process supported KB documents, or at least skip unsupported files without repeatedly retrying them.

### What actually happens
For larger PDFs, the admin logs show:

- `BadRequestError: 400 the input length exceeds the context length`
- failure inside `OllamaService.embed`
- followed by `Failed to embed and store the extracted text.`

For Sync Storage, the logs show the same job being retried over and over with:

- `Error: Unsupported file type.`

### Relevant log details
Examples from `docker compose logs --tail=120 admin`:

- `BadRequestError: 400 the input length exceeds the context length`
- stack includes:
  - `OllamaService.embed`
  - `RagService.embedAndStoreText`
  - `RagService.embedTextAndCleanup`
  - `RagService.processAndEmbedFile`
- repeated:
  - `[ error ] [file-embeddings] Job failed: ... Error: Failed to embed and store the extracted text.`
- repeated on Sync Storage:
  - `[ error ] [file-embeddings] Job failed: 7d9aa7fc644fa8e0, Error: Unsupported file type.`

I also saw successful jobs in the same environment, so the KB pipeline is not completely broken. Some `embed-file` jobs complete, while larger PDFs fail.

### Environment details
Host Ollama is reachable and has these models available:

- `llama3.1:latest`
- `qwen2.5:3b`
- `nomic-embed-text:v1.5`

### Storage layout observation
My `/opt/project-nomad/storage` directory is not docs-only. It contains:

- `kb_uploads/`
- `kolibri/`
- `maps/`
- `ollama/`
- `qdrant/`
- `zim/`
- logs and other app data

The actual uploaded KB files are under `kb_uploads/`, but Sync Storage appears to be interacting with files outside that folder as well.

### Reproduction
[...]

Issue #670, opened 2026-04-06
[Bug]: File upload fails due to embedding context length (inconsistent chunking)


bdeluca94:
### Issue Category

AI Knowledge Base Uploads

### Bug Description

Summary

Uploading certain PDFs and TXT files to the Knowledge Base fails during the embedding stage with:

400 the input length exceeds the context length

This occurs even with relatively small files (~200KB), while larger files sometimes succeed.

Environment
Platform: WSL (Ubuntu 22.04) on Windows
Deployment: Docker (Project NOMAD)
CPU-based (no GPU currently)
Embeddings via Ollama
Qdrant used for vector storage (manually restored and network-connected)
Observed Behavior
File upload begins normally
Text extraction succeeds
Embedding job fails around ~10–30% (often ~19%)
Error in logs:
BadRequestError: 400 the input length exceeds the context length

Stack trace shows failure in embedding step:

OllamaService.embed
RagService.embedAndStoreText
Key Findings
Issue is not file size dependent
Issue is not PDF validity (confirmed via pdftotext)
Same content succeeds when:
converted to .txt
manually split into smaller chunks before upload

This strongly suggests:
👉 Files are sometimes passed to the embedding model without proper chunking

### Steps to Reproduce

Reproduction
Upload a medium/large PDF or dense TXT file (~200KB–1MB)
Observe failure during embedding phase
Check logs → context length error
Expected Behavior
Files should be automatically chunked into embedding-safe sizes
No single embedding request should exceed model context limits
Actual Behavior
Entire document or oversized chunks appear to be sent to embedding model
Causes hard failure instead of graceful chunking
Workaround
Convert PDF → TXT using pdftotext
Manually split file:
split -l 300 file.txt part_
Upload smaller chunks individually

Additional Notes
Earlier issues were caused by missing Qdrant container and networking, but those have been resolved
Current issue persists independently of infrastructure

### Expected Behavior

Suspected Cause
Chunking logic in file ingestion pipeline is:
inconsistent
or bypassed for certain file types / formats
No safeguard against exceeding embedding model context window
Suggested Fix
Enforce max token/character limit per embedding request
Ensure consistent chunking across all file types
Add fallback splitting if chunk exceeds model limits
Optionally expose chunk size config

### Actual Behavior

Occasionally accepted split files inconsistently. 

### N.O.M.A.D. Version

v1.31.0

### Operating System

Other (not yet officially supported)

### Do you have a dedicated GPU?

No

[...]

Issue #369, opened 2026-03-16
[Bug]: Knowledge Base PDF failing to process after upload.


ewcook:
### Issue Category

Installation/Setup

### Bug Description

After uploading a PDF to the knowledge Base the processing fails at %19. 

### Steps to Reproduce

Upload PDF to knowledge base

### Expected Behavior

Process the data and import for use with the AI chatbot

### Actual Behavior

Fails at %19

### N.O.M.A.D. Version

1.29.1

### Operating System

Ubuntu 24.04

### Docker Version

29.3.0

### Do you have a dedicated GPU?

Yes

### System Specifications

QEMU VM:

CPU: Intel(R) Core(TM) Ultra 7 165H X 18
RAM: 16G
NPU: Processing accelerators: Intel Corporation Meteor Lake NPU (rev 04)
Storage: 250GB

### Relevant Logs

```shell

```

### Browser Console Errors (if UI issue)

```javascript

```

### Pre-submission Checklist

- [x] I have searched for existing issues that might be related to this bug
- [x] I am running the latest version of Project N.O.M.A.D. (or have noted my version above)
- [x] I have redacted any personal or sensitive information from logs and screenshots
- [x] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

chriscrosstalk:
Can you share more details? Specifically: how large is the PDF, and what error message (if any) do you see? The default upload limit is 10MB (configurable via PR #316 once it merges). If the file is large, that could be the issue. Also, is your AI Assistant installed and running?

ewcook:
The PDF files is between 2M

2.0M Mar 15 18:26 infoscale_install_802_lin.pdf


Error from logs 

{"level":30,"time":1773834266806,"pid":1,"hostname":"26fbe4cfe65e","msg":"[EmbedFileJob] Dispatched embedding job for file: infoscale_install_802_lin.pdf-74950f961d0c.pdf"}
{"level":30,"time":1773834266810,"pid":95,"hostname":"26fbe4cfe65e","msg":"[EmbedFileJob] Starting embedding process for: infoscale_install_802_lin.pdf-74950f961d0c.pdf"}
{"level":30,"time":1773834266839,"pid":95,"hostname":"26fbe4cfe65e","msg":"[EmbedFileJob] Services ready. Processing file: infoscale_install_802_lin.pdf-74950f961d0c.pdf"}
[...]

PR #763, opened 2026-04-22
fix(rag): pass num_ctx and truncate to Ollama embed call

### LINKED ISSUE #762: [Bug]: PDF/text embedding fails with 'input length exceeds the context length' on certain content
## Summary

`OllamaService.embed()` in `admin/app/services/ollama_service.ts` calls `/api/embed` without passing `options.num_ctx` or `truncate`. When the host Ollama installs nomic-embed-text:v1.5 with the embedding model's default `num_ctx=2048` (the model's training context, per its model card), the RAG chunker can produce inputs that exceed it on dense PDFs and other content with high token density.

The native `/api/embed` then errors with `the input length exceeds the context length`. The catch block silently falls back to the OpenAI-compatible `/v1/embeddings` endpoint, which also errors and surfaces as `BadRequestError: 400` from the OpenAI SDK.

## Root cause

`admin/app/services/rag_service.ts`:
- `MAX_SAFE_TOKENS = 1600` and `TARGET_TOKENS_PER_CHUNK = 1500` are sized for nomic's documented 2K architectural context
- `CHAR_TO_TOKEN_RATIO = 2` underestimates real BPE token count for dense English text (real ratio is closer to 1.3 chars/token)
- Net: a 1500-target chunk can be ~3000 chars / ~2300 actual tokens, exceeding 2048

The chunker has a safety check that calls `truncateToTokenLimit`, but the truncation uses the same wrong char-to-token ratio, so it doesn't bring the chunk under the real limit either.

`admin/app/services/ollama_service.ts:484` then sends the chunk to `/api/embed` without overriding `num_ctx`, so Ollama uses whatever the local modelfile says. On installs where `num_ctx=2048`, the request fails.

## Reproduction

NOMAD3 happens to have nomic-embed-text:v1.5 with `num_ctx=8192` in its modelfile, so it does not reproduce locally. The bug surfaces when:

- Host Ollama install (vs the bundled nomad_ollama container)
- Different Ollama version that respects/enforces the architectural 2048 limit
- Or any environment where the embedding model's effective num_ctx is 2048

Confirmed reports (all same `BadRequestError: 400 the input length exceeds the context length` stack trace into `OllamaService.embed`):
- #756 (NC4WD, v1.31.0, host Ollama, nomic-embed-text:v1.5)
- #670 (closed without an actual fix)
- #369 (closed without an actual fix)
- @hestela mentioned hitting the same in the #369 thread

## Fix

Pass `options.num_ctx: 8192` (nomic-embed-text v1.5's RoPE-extrapolated max) and `truncate: true` (silent truncation safety net) on every `/api/embed` call. PR incoming.

This makes the embed request explicit about the context budget instead of relying on each install's local modelfile defaults, and gives Ollama permission to silently truncate any chunk that still overshoots.

chriscrosstalk:
## Summary

Closes #762. Addresses Bug 1 of #756.

`OllamaService.embed()` was calling `/api/embed` with just `{ model, input }` and no `num_ctx` override. On host Ollama installs where nomic-embed-text:v1.5 ships with `num_ctx=2048` (its training context per the model card), our chunker can produce inputs that exceed that limit on dense content. The native call errors, the catch silently falls back to the OpenAI-compatible endpoint which also errors, and the user sees `BadRequestError: 400 the input length exceeds the context length`.

This patch passes:
- `options.num_ctx: 8192` (nomic-embed-text v1.5's RoPE-extrapolated max) so the embed call doesn't depend on the local modelfile's default
- `truncate: true` so Ollama silently truncates any input that still overshoots, instead of erroring

Same root cause as #369 and #670 (both closed without an actual fix). Reported on #756 by @NC4WD with a clean stack trace and storage layout details that pinpointed the codepath.

## Test plan (NOMAD3, v1.31.1, nomad_ollama container)

Pre-patch baseline: uploaded 3 PDFs and 1 .txt of varying density/size; all succeeded because NOMAD3's modelfile already has `num_ctx=8192` so the bug doesn't trigger here. Confirms the fix won't regress installs that were already working.

Post-patch verification: re-uploaded the same 4 files via `POST /api/rag/upload`; all 4 jobs completed (`active-jobs`, `failed-jobs` both `[]`); nomad_ollama logs show all `/api/embed` calls returning 200; chunks visible in `/api/rag/files`.

| File | Size | Pre-patch result | Post-patch result |
|---|---|---|---|
| Box-Oven-Cooking.pdf | 13 KB | succeeded | succeeded |
| A-Better-Way-to-Gardening.pdf | 630 KB | succeeded | succeeded |
| Survival-and-Austere-Medicine.pdf | 4 MB (192 chunks, 24 batches) | succeeded | succeeded |
| Federalist Papers (Project Gutenberg, dense plain text) | 1.1 MB (~340 chunks) | succeeded | succeeded |

Reproducing the actual failure mode requires an Ollama install where `num_ctx=2048` is enforced. The fix is defensive against that environment without changing behavior on installs that were already working.

## Notes

- No chunker constants changed in this PR. `CHAR_TO_TOKEN_RATIO=2` is still inaccurate (real BPE is ~1.3 chars/token for English) but the runtime `num_ctx=8192` plus `truncate:true` makes that estimation error harmless. A follow-up PR could tighten the chunker for better retrieval quality but is not needed to fix the user-facing bug.
- The OpenAI-compatible fallback path (`/v1/embeddings`) is untouched. It's used for non-Ollama backends (LM Studio etc.) which have their own context handling.

### DIFF
--- a/admin/app/services/ollama_service.ts
+++ b/admin/app/services/ollama_service.ts
@@ -480,10 +480,21 @@ export class OllamaService {
     }
 
     try {
-      // Prefer Ollama native endpoint (supports batch input natively)
+      // Prefer Ollama native endpoint (supports batch input natively).
+      // Pass num_ctx explicitly so we don't depend on the embedding model's
+      // modelfile defaults. Some installs ship nomic-embed-text:v1.5 with
+      // num_ctx=2048, which our chunker (sized for ~1500 tokens) can exceed
+      // on dense content, causing "input length exceeds context length" errors.
+      // truncate:true is a runtime safety net for any chunk that still overshoots.
+      // 8192 matches nomic-embed-text:v1.5's RoPE-extrapolated max.
       const response = await axios.post(
         `${this.baseUrl}/api/embed`,
-        { model, input },
+        {
+          model,
+          input,
+          truncate: true,
+          options: { num_ctx: 8192 },
+        },
         { timeout: 60000 }
       )
       // Some backends (e.g. LM Studio) return HTTP 200 for unknown endpoints with an incompatible


jakeaturner (approved):
LGTM! Though I think this is definitely something we should make configurable in "Advanced Mode" - JT 4/27/2026

PR #767, opened 2026-04-22
Add Caddy authentication configuration instructions

### LINKED ISSUE #764: [Feature]: Add authentication instructions to the documentation
### Feature Category

Documentation

## Summary
Project N.O.M.A.D. currently does not include built-in authentication by design, which makes it simple and accessible for local use. However, when users expose their instance to a local network (or beyond), there is no official guidance on how to restrict access to sensitive routes such as system settings.

## Problem
Users who:
- expose NOMAD to a LAN
- port forward for remote access
- or share their instance with others

may unintentionally allow unrestricted access to administrative functionality.

While the README mentions that users should rely on network-level controls, there are no concrete examples provided for how to implement this in practice.

I have a homeserver that I expose to the internet, I have a homepage that links to various resources and webpages I have running.  This setup was necessary for me so that I could grant users of my site access to my nomad resources. 

### Proposed Solution (#767) 

Add a documentation section demonstrating how to use a reverse proxy (Caddy) to implement basic authentication for specific routes.

The proposed documentation:
- Shows how to install Caddy
- Explains how to generate a secure password hash
- Provides a minimal working Caddyfile example
- Demonstrates protecting `/settings/system` using `basicauth`
- Clarifies that NOMAD itself should not be directly exposed

## Why Caddy?
Caddy was chosen because:
- It is easy to configure (single file)
- It includes built-in HTTPS and authentication support
- It is beginner-friendly for homelab users
- It requires minimal setup compared to alternatives

## Scope
This does NOT:
- modify NOMAD’s codebase
- introduce built-in authentication
- change default behavior

It only adds optional documentation for users who want additional access control.
See the attached pull request for the exact change I am proposing.

## Notes
This aligns with the project's philosophy of keeping NOMAD lightweight while still providing guidance for more advanced or security-conscious users.

This is the same process I have used to configure my own home server (although my caddyfile is a bit more customized since it's part of a larger collection of services/webservers). I figured I should share my solution so that others do not have to go through the trial and error that I did to set this up. 

Happy to adjust wording or expand this section further if needed.

### Alternative Solutions

_No response_

### Use Case

As mentioned above, I have already implemented this because I needed it for my home server and it was easier than implementing it manually into the project nomad codebase on my fork.

All Users

Low - Minor convenience

Since this doesn't require an implementation within the software itself, I think this is a really good implementation of security. If there are other dangerous directories that are on the same port as the splash page, the instructions should be modified to reflect that.

### Examples or References

_No response_

Yes - I can write the code

### Additional Context

_No response_

### Pre-submission Checklist

Zacharyprime:
Added instructions for configuring authentication with Caddy, including installation steps and example configuration.

## What this PR does
This PR adds a new section to the README under "About Security" that demonstrates how to configure optional network-level authentication using Caddy.

closes #764

## Why this is useful
NOMAD is intentionally designed without authentication, which works well for local, single-user setups. However, users who expose their instance to a network may want a simple way to restrict access to sensitive routes (such as `/settings/`).

This PR provides:
- A practical, copy-pasteable Caddy setup
- Instructions for generating password hashes
- A minimal reverse proxy configuration
- Guidance on safe port exposure

## Key points
- Does NOT introduce authentication into NOMAD itself
- Does NOT change any existing functionality
- Only adds documentation for optional external configuration

## Example use case
A user hosting NOMAD on a home server can:
- expose a single port via Caddy
- protect admin routes with a username/password
- keep the main interface accessible

## Additional context
This approach keeps NOMAD simple while giving users a clear path to improve security without requiring built-in auth.

## Testing
Tested locally with:
- NOMAD running on port 8080
- Caddy reverse proxy on a separate port
- `/settings/*` correctly prompting for authentication

## Future considerations
If built-in authentication is added later, this section can either:
- remain as an advanced alternative
- or be simplified/removed depending on direction

---

Let me know if you'd like any changes or additional examples included.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ By design, Project N.O.M.A.D. is intended to be open and available without hurdl
 For now, we recommend using network-level controls to manage access if you're planning to expose your N.O.M.A.D. instance to other devices on a local network. N.O.M.A.D. is not designed to be exposed directly to the internet, and we strongly advise against doing so unless you really know what you're doing, have taken appropriate security measures, and understand the risks involved.
 
 ### How to configure authentication with Caddy (network level authentication)
-It is possible to use a caddy configuration in order to prevent users from accessing certain pages, apps or ports. Here we will show how to configure the settings menu to require a username and password to access. It is recommended not to expose your system to the internet unless you know what you are doing. Please follow these steps at your own risk.
+It is possible to use a caddy configuration in order to prevent users from accessing certain pages, apps or ports. Here we will show how to configure the settings menu to require a username and password to access. It is recommended not to expose your system to the internet unless you know what you are doing. Please follow these steps at your own risk. These instructions are intended as a helpful example and not a prescription for good network security policy.
 
 First install caddy:
 ```bash
@@ -139,10 +139,17 @@ Example Configuration File:
 
     reverse_proxy localhost:8080
 
-    @protected path /settings/*
+    @protected {
+        path /settings/*
+        path /api/settings/*
+        path /api/system/*
+        path /api/docker/*
+        path /api/jobs/*
+        path /api/services/*
+    }
+
     basicauth @protected {
         <user> <hash>
-
     }
 }
 
@@ -156,7 +163,10 @@ user: This is the username you want to login with
 
 hash: This is the hash of your password generated in the earlier step
 
-Note that this assumes project nomad is hosted on port 8080. Also, port 8080 (or whichever port you have nomad configured to) does not need to be port forwarded. The reverse proxy will automatically forward the connection.
+Notes:
++ This assumes the nomad command center is hosted on port 8080 (default)
++ Port 8080 (or whichever port is configured for the command center) does not need to be port forwarded. The reverse proxy will automatically forward the connection.
++ If you use a real domain (e.g. nomad.example.com), Caddy will automatically provision and renew TLS certificates via Let’s Encrypt.
 
 Once everything has been configured, restart the caddy service to apply changes:
 ```
--- a/README.md
+++ b/README.md
@@ -114,6 +114,55 @@ By design, Project N.O.M.A.D. is intended to be open and available without hurdl
 
 For now, we recommend using network-level controls to manage access if you're planning to expose your N.O.M.A.D. instance to other devices on a local network. N.O.M.A.D. is not designed to be exposed directly to the internet, and we strongly advise against doing so unless you really know what you're doing, have taken appropriate security measures, and understand the risks involved.
 
+### How to configure authentication with Caddy (network level authentication)
+It is possible to use a caddy configuration in order to prevent users from accessing certain pages, apps or ports. Here we will show how to configure the settings menu to require a username and password to access. It is recommended not to expose your system to the internet unless you know what you are doing. Please follow these steps at your own risk.
+
+First install caddy:
+```bash
+sudo apt update
+sudo apt install caddy
+```
+Caddy does not store your password as raw text, you must generate a hash for your password by running the following command:
+```bash
+caddy hash-password
+```
+The program will prompt you to create a password and will output the appropriate hash to put in the caddyfile.
+
+Then create the caddy configuration
+```
+sudo nano /etc/caddy/Caddyfile
+```
+
+Example Configuration File:
+```
+<domain>:<port> {
+
+    reverse_proxy localhost:8080
+
+    @protected path /settings/*
+    basicauth @protected {
+        <user> <hash>
+
+    }
+}
+
+```
+
+domain: This is the domain name of your server if you have one (leave blank or put public ip otherwise).
+
+port: This is the port you plan to port forward for access to nomad. 
+
+user: This is the username you want to login with
+
+hash: This is the hash of your password generated in the earlier step
+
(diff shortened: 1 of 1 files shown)


chriscrosstalk (changes requested):
Really useful contribution — the audience (homelab users exposing NOMAD past localhost) maps exactly onto our real user base. Glad you walked through it from your own setup rather than copy-pasting a generic snippet.

**One change needed before merge (security):** The matcher `path /settings/*` protects the admin UI routes but not `/api/*`. NOMAD's settings UI performs mutations via `/api/settings/*`, `/api/system/*`, `/api/docker/*`, `/api/jobs/*`, etc. — all of which are reachable without hitting `/settings/*` at all. Anyone reading our public API reference could bypass the basicauth with a `curl -X POST /api/settings/write ...`. A user who follows these docs thinking admin is locked down would have a false sense of security, which is worse than having no docs here.

Two ways to close the gap:

**Option A — narrow model** (keep public access to library/chat/docs, lock admin):
```
@protected {
    path /settings/*
    path /api/settings/* /api/system/* /api/docker/* /api/jobs/* /api/services/*
}
```

**Option B — fully private** (basicauth everything; simpler, arguably the more common case when exposing to the internet):
```
basicauth {
    <user> <hash>
}
```
applied at the site block level with no matcher.

Your current example is closest to Option A — if you want to go with that route, just expand the path list. Option B is a simpler default and probably what we'd recommend for anyone not sure which they want. Your call.

**One small suggestion (not a blocker):** Mention Caddy auto-HTTPS — it's the killer feature and the biggest reason to pick Caddy over nginx. Something like: *"If you use a real domain name (e.g. `nomad.example.com`) instead of an IP, Caddy will automatically provision and renew a Let's Encrypt certificate for you."*

Thanks again for taking the time on this.

Zacharyprime:
I agree with your suggestions and have made the changes and updated some of the language for clarity.

I am 90% sure the change shown should work appropriately based on my experience using caddy which is why I went ahead and made the commit. However, I do not have ssh access to my server at the moment. Please let me verify this later today or tomorrow on my actual system before proceeding.

chriscrosstalk:
Changes look good on paper — the expanded `@protected` matcher covers the right API surfaces, the Let's Encrypt note is in, and I appreciate the "helpful example, not a prescription" disclaimer. That's the right tone for this kind of docs.

Happy to wait for your end-to-end verification before approving. When you've confirmed it works against your actual NOMAD instance (specifically: test that `curl -X POST http://<domain>:<port>/api/settings/write ...` returns a 401 without credentials, since that's the specific bypass the original example left open), just drop a note here and I'll flip to approved.

Zacharyprime:
I have confirmed that this solution does work with one caveat.

The command center page probably makes some API requests in the background for preloading purposes or something. So with this configuration, you still get prompted for a login when you open the command center. But, you can just hit cancel and it will still allow you to access all of the tools. It's a minor inconvenience but does not hinder any function.

Four copies of these error messages will display but do not appear to cause any problems:
image


If you try to open the settings menus and hit cancel, it rejects and places the user into a 401 page as expected.

And of course if you type in the correct login everything functions as normal. If you hit cancel on the first prompt for the command center, you will still be prompted with another chance to login after clicking the settings menu. Entering the correct login for either prompt will function as expected.

After the user has logged in once, the login state appears to save. It doesn't appear to be the browser doing this since it doesn't get added to my saved passwords. I suspect there is some time window that caddy allows reconnection. I verified with other devices that it is always system/browser specific, so I see this as a feature.

Testing via curl requests I get this response with no password:
```bash
HTTP/2 401 
alt-svc: h3=":8081"; ma=2592000
server: Caddy
www-authenticate: Basic realm="restricted"
content-length: 0
date: Wed, 22 Apr 2026 20:02:30 GMT

```

And then this response with a user/password passed with the -u flag:
```bash
HTTP/2 404 
alt-svc: h3=":8081"; ma=2592000
content-type: text/html; charset=utf-8
date: Wed, 22 Apr 2026 20:02:41 GMT
via: 1.1 Caddy
x-request-id: eodv7532yas9c0m2g9s5ee27
content-length: 1831
#... the html content I won't share here...
```
The fact that this is a 404 response is not concerning here, it's just because I've sent it an empty request.

chriscrosstalk:
Thanks @Zacharyprime — really appreciate the polish here. The verification work especially: testing `curl -X POST /api/settings/write` to confirm the 401 closes the original bypass was exactly the right move, and you took the security pushback seriously and made the docs better.

After thinking on this, I want to keep the README focused on the supported NOMAD experience. Adding a Caddy reverse-proxy recipe to the README — even with disclaimers — reads as endorsement and pulls us toward supporting an arbitrary range of edge-case configs that aren't really NOMAD's surface area.

The better home for this: a how-to guide on projectnomad.us with prominent "this is unsupported, you take on the security responsibility" caveats — same approach we took with the WSL2 install guide. Adding it to my to-do list. When it lands, your verified config will be the foundation and you'll get credit. Closing this PR but the contribution isn't lost.

chriscrosstalk:
@Zacharyprime — closing the loop on this one. The Caddy + basicauth guide is live: **https://www.projectnomad.us/install/caddy**

Your verified config is the foundation, including the `@protected` matcher expansion that closed the original `/api/*` bypass. You're credited in the Community credit section with a link back to this PR. Thanks again for the work.

PR #771, opened 2026-04-22
fix(ZIM): prune completed downloads from Content Explorer remote list

johno10661:
## Summary

Fixes a "ghost entry" bug in the ZIM Content Explorer (`/settings/zim/remote-explorer`) where a ZIM that has just finished downloading reappears in the remote browse list after scrolling.

### Root cause

`ZimService.listRemote()` filters installed ZIMs server-side using a snapshot of the on-disk set taken at call time. Each paginated fetch is an independent call. A page fetched *during* the download returned the ZIM (it wasn't installed yet). The client accumulates that page into its list.

Client-side, `flatData` in `remote-explorer.tsx` filtered accumulated pages only against the *active* downloads list. When the download completed, the BullMQ job transitions out of `active|waiting|delayed|failed`, so `listDownloadJobs` stops returning it — the filter stops matching — and the stale accumulated entry reappears until a full page refresh.

### Fix

Client-side filter extension. Does **not** touch `listRemote()` (recently reworked in #746).

- Subscribe to the installed ZIM list via `['zim-files']` (same query key the Content Manager page already uses, so no duplicate fetch).
- `flatData` now prunes items whose filename is in the installed set *or* currently downloading.
- When a job drops off the polled downloads list (completion or cancel), invalidate `['zim-files']` so the newly-installed file is picked up on the next render — the ghost disappears immediately instead of at next user-triggered refresh.

## Repro

1. Open Content Explorer (`/settings/zim/remote-explorer`).
2. Click Download on any ZIM — it moves to the "downloading" section and is removed from the remote list.
3. While the download is in progress, scroll the remote list to trigger pagination fetches.
4. Wait for the download to complete.
5. Scroll again.

**Before:** the now-installed ZIM reappears in the remote list. Page refresh makes it disappear.  
**After:** the ZIM stays filtered out; no refresh needed.

## Test plan

- [x] Reproduce the bug on a real instance (before patch): ghost reappears on scroll after completion.
- [x] Verify fix on a real instance (after patch): ghost stays filtered out.
- [ ] Verify a plain download (no scrolling during) still works end-to-end.
- [ ] Verify infinite-scroll pagination is unbroken (no early termination, no extra fetches).
- [ ] Verify cancel flow still removes the entry from active downloads and does not re-surface a non-installed ZIM.

## Notes

- Shares the `['zim-files']` cache key with `admin/inertia/pages/settings/zim/index.tsx` (Content Manager). Opening the Explorer no longer triggers a duplicate fetch if the Manager tab has been visited recently.
- No backend changes; no migration; no new dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,8 +43,10 @@ ENV NODE_ENV=production
 WORKDIR /app
 COPY --from=production-deps /app/node_modules /app/node_modules
 COPY --from=build /app/build /app
-# Copy root package.json for version info
-COPY package.json /app/version.json
+# Generate version.json from the VERSION build-arg so the image tag is the
+# single source of truth (previously copied root package.json, which drifted
+# from the tag when semantic-release did not commit the bump back).
+RUN echo "{\"version\":\"${VERSION}\"}" > /app/version.json
 
 # Copy docs and README for access within the container
 COPY admin/docs /app/docs
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Contributions are welcome and appreciated! Please see [CONTRIBUTING.md](CONTRIBU
 - **Benchmark Leaderboard:** [benchmark.projectnomad.us](https://benchmark.projectnomad.us) - See how your hardware stacks up against other NOMAD builds
 - **Troubleshooting Guide:** [TROUBLESHOOTING.md](TROUBLESHOOTING.md) - Find solutions to common issues
 - **FAQ:** [FAQ.md](FAQ.md) - Find answers to frequently asked questions
+- **Community Add-Ons:** [admin/docs/community-add-ons.md](admin/docs/community-add-ons.md) - Third-party content packs built by the community
 
 ## License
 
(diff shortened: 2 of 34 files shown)


chriscrosstalk (dismissed):
Approve. Tight client-side fix — extends the existing `flatData` filter to also exclude items whose filename is in the installed set, and invalidates the `['zim-files']` cache when a download drops out of the polled list (covers both completion and cancellation; harmless on cancel since the file isn't installed).

Sharing the `['zim-files']` query key with the Content Manager page is the right call — no duplicate fetch, single source of truth for "what's on disk." Filename extraction via `download_url.split('/').pop()` matches the pattern already used a few lines below in the active-downloads filter.

Doesn't touch `listRemote()` (recently reworked in #746), keeps the blast radius small.

chriscrosstalk:
Thanks @johno10661 — confirmed still reproducible on `dev`: `remote-explorer.tsx` `flatData` only filters out items that are currently downloading, not ones that just finished installing, so a ZIM completed mid-pagination reappears until a full refresh. Your analysis matches the current code exactly.

Same situation as your other PR though — the branch is on a stale base and the diff is replaying a lot of already-merged work on top of the real ~45-line change. Could you rebase onto current `origin/dev`? Happy to take it once it's a clean diff. Appreciate the detailed repro.

chriscrosstalk:
Thanks for the thorough diagnosis here. Since you filed this, the accumulated-page and installed-file filtering groundwork landed on `dev` (via the #731 work), which fixed most of it. The one remaining gap was that the installed-files list was not refreshed when a download completed, so a just-installed ZIM could still ghost back in. I have put up a focused fix for exactly that in #1099, credited to you for the original diagnosis. Closing this in favor of that one, since your branch had also drifted a good bit from current `dev`. Appreciate the detailed root-cause writeup.

PR #774, opened 2026-04-22
fix(ZIM): narrow reconcile Wikipedia skip to the managed selection file

johno10661:
## Summary

Fixes a bug where a category tier (e.g. **Medicine → Comprehensive**) reverts to a lower tier (**Standard**) after every container restart, even though the underlying ZIM files are still on disk.

### Root cause

`CollectionManifestService.reconcileFromFilesystem()` (`admin/app/services/collection_manifest_service.ts`) skipped every on-disk ZIM whose filename starts with `wikipedia_en_`:

```ts
// Skip Wikipedia files (managed by WikipediaSelection model)
if (file.name.startsWith('wikipedia_en_')) continue
```

The intent was to leave the file tracked by `WikipediaSelection` alone. But the prefix is too broad — curated category tiers in `collections/kiwix-categories.json` include Wikipedia-themed ZIMs:

- `wikipedia_en_medicine_maxi_2026-01.zim` (Medicine → Comprehensive)

Anything matching `wikipedia_en_*` was skipped before it could be added to `seenZimIds`. The cleanup pass immediately after then deleted any `installed_resources` row whose `resource_id` wasn't in `seenZimIds` — so these rows got wiped on every restart.

Result: `getInstalledTierForCategory` saw the higher tier as incomplete and silently returned the lower one. In the reporter's install, Comprehensive reverted to Standard on every restart.

### Fix

Narrow the skip to the single filename actually tracked by `WikipediaSelection`:

```ts
const wikipediaSelection = await WikipediaSelection.query().first()
const managedWikipediaFilename = wikipediaSelection?.filename ?? null
...
if (managedWikipediaFilename && file.name === managedWikipediaFilename) continue
```

All other `wikipedia_en_*` files reconcile normally. No change to the rest of reconcile, no migration, no schema change.

## Repro

1. Download the Medicine → Comprehensive tier (installs `wikipedia_en_medicine_maxi_*.zim` among others).
2. Verify the Content Explorer shows Medicine as "Comprehensive".
3. `docker restart nomad-admin`.
4. **Before:** Medicine reverts to "Standard". `SELECT * FROM installed_resources WHERE resource_id='wikipedia_en_medicine_maxi'` returns 0 rows.  
   **After:** Medicine stays at "Comprehensive". The `installed_resources` row survives the restart.

## Test plan

- [x] Reproduce the bug on a real instance (pre-patch): comprehensive tier row gets wiped on every restart.
- [x] Verify fix on a real instance (post-patch): row persists across restarts; tier stays correct.
- [x] Verify the `WikipediaSelection`-managed file (e.g. `wikipedia_en_all_maxi_*`) is still excluded from `installed_resources` post-restart.
- [ ] Verify a fresh install with no `wikipedia_selections` row reconciles all `wikipedia_en_*` files normally (no skip).
- [ ] Verify downloading a new curated tier containing a `wikipedia_en_*` file, then restarting, still reports the correct tier.

## Notes

- Side effect: other `wikipedia_en_*` curated ZIMs that had been silently missing from `installed_resources` on affected installs will now get reconciled in on the next startup (17 such rows appeared on the reporter's install after the patch). This is the intended behavior — they were supposed to be tracked all along.
- Does not touch any Wikipedia selector flow. `ZimService.downloadRemoteSuccessCallback` and `onWikipediaDownloadComplete` still own the selector's installed/uninstalled state via `WikipediaSelection`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,8 +43,10 @@ ENV NODE_ENV=production
 WORKDIR /app
 COPY --from=production-deps /app/node_modules /app/node_modules
 COPY --from=build /app/build /app
-# Copy root package.json for version info
-COPY package.json /app/version.json
+# Generate version.json from the VERSION build-arg so the image tag is the
+# single source of truth (previously copied root package.json, which drifted
+# from the tag when semantic-release did not commit the bump back).
+RUN echo "{\"version\":\"${VERSION}\"}" > /app/version.json
 
 # Copy docs and README for access within the container
 COPY admin/docs /app/docs
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Contributions are welcome and appreciated! Please see [CONTRIBUTING.md](CONTRIBU
 - **Benchmark Leaderboard:** [benchmark.projectnomad.us](https://benchmark.projectnomad.us) - See how your hardware stacks up against other NOMAD builds
 - **Troubleshooting Guide:** [TROUBLESHOOTING.md](TROUBLESHOOTING.md) - Find solutions to common issues
 - **FAQ:** [FAQ.md](FAQ.md) - Find answers to frequently asked questions
+- **Community Add-Ons:** [admin/docs/community-add-ons.md](admin/docs/community-add-ons.md) - Third-party content packs built by the community
 
 ## License
 
--- a/admin/app/controllers/benchmark_controller.ts
+++ b/admin/app/controllers/benchmark_controller.ts
@@ -5,6 +5,7 @@ import { runBenchmarkValidator, submitBenchmarkValidator } from '#validators/ben
 import { RunBenchmarkJob } from '#jobs/run_benchmark_job'
 import type { BenchmarkType } from '../../types/benchmark.js'
 import { randomUUID } from 'node:crypto'
+import logger from '@adonisjs/core/services/logger'
 
 @inject()
 export default class BenchmarkController {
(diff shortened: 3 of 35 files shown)


chriscrosstalk (dismissed):
Approve. Good catch on the over-broad skip — the prefix match was wiping `installed_resources` rows for any curated `wikipedia_en_*` ZIM (Medicine → Comprehensive's `wikipedia_en_medicine_maxi`, etc.) on every reconciliation pass.

Narrowing to the single filename actually tracked by `WikipediaSelection` is the right scope. The extra `WikipediaSelection.query().first()` per reconciliation is fine — it runs once at startup, not in any hot path. Failsafe behavior is correct: if no selection row exists (`managedWikipediaFilename === null`), every Wikipedia-themed file gets reconciled normally rather than skipped, which is the desired default.

chriscrosstalk:
Thanks @johno10661 — I traced this through and confirmed the bug is still live on `dev`: `collection_manifest_service.ts` still skips every `wikipedia_en_*` file in `reconcileFromFilesystem()`, so curated tiers that ship a `wikipedia_en_*` ZIM (e.g. Medicine → Comprehensive) get their `installed_resources` rows wiped on restart. Your root-cause writeup is spot on and the fix is the right shape.

The only thing holding it up is the base — the branch is far enough behind that the PR is replaying a couple dozen already-merged commits, which buries the actual 10-line change. Could you rebase onto current `origin/dev`? Once it's a clean one-file diff we can move it forward. Thanks for the thorough report.

chriscrosstalk:
Thanks @johno10661 — this is a real bug and your fix is exactly right. Since the branch was trapped on a stale base (the diff had picked up ~35 files of already-merged commits), I've reimplemented your change on a clean branch off `dev` in #995 with credit to you. Closing this in favor of that. Appreciate the catch and the clean diagnosis.

PR #775, opened 2026-04-23
feat(docker): add compose identity labels to managed containers

maximpedraza0:
## Summary

NOMAD currently labels the containers it creates via Easy Setup with only `com.docker.compose.project=project-nomad-managed` and `io.project-nomad.managed=true`. Tools that match containers against a compose project — `docker compose ps`, Portainer, Unraid's Compose Manager — can't group these containers under the managed stack because the rest of the standard compose identity labels are missing. Users see the managed containers as orphans in the Docker tab, or as an empty stack in compose-aware UIs.

This PR sets the complete set of compose identity labels on every container `DockerService` creates, so those tools treat them as first-class members of a compose project.

## Labels added (always set)

| Label | Value |
|-------|-------|
| `com.docker.compose.service` | `service.service_name` |
| `com.docker.compose.container-number` | `"1"` |
| `com.docker.compose.oneoff` | `"False"` |
| `com.docker.compose.version` | `"2.0.0"` |
| `com.docker.compose.config-hash` | sha256 of service_name (stable per service) |
| `com.docker.compose.depends_on` | `""` (NOMAD doesn't model compose-style deps) |
| `com.docker.compose.image` | image digest from `docker.getImage().inspect()` |

All values are deterministic and computable from information NOMAD already has — no configuration needed for the baseline behaviour.

## Deployment-specific labels (opt-in via env vars)

`docker compose ps` also matches containers against the working dir and the config files of the current compose invocation. Those paths are deployment-specific, so NOMAD can't infer them. Two optional env vars let operators set them:

- `NOMAD_MANAGED_WORKING_DIR` → `com.docker.compose.project.working_dir`
- `NOMAD_MANAGED_CONFIG_FILES` → `com.docker.compose.project.config_files`

When unset (default), those two labels are simply omitted. Environments that don't need full compose-manager integration get the benefit of the other 7 labels without having to configure anything.

## Isolation between multiple NOMAD instances

While in this file, the managed project name and the network that utility containers attach to are also made configurable:

- `NOMAD_MANAGED_PROJECT` → overrides the `project-nomad-managed` default
- `NOMAD_DOCKER_NETWORK` → overrides the `project-nomad_default` default

Defaults unchanged, so existing deployments keep working as-is. Operators who want to run multiple NOMAD instances on the same host (e.g. staging alongside production) can now do so without the managed containers colliding.

## Test plan

- [ ] Install any service from Easy Setup → new container has all 7 core labels
- [ ] `docker compose ls --all` now lists `project-nomad-managed` as a project
- [ ] With `NOMAD_MANAGED_WORKING_DIR` set, `docker compose -p project-nomad-managed ps` lists the managed containers
- [ ] Existing deployments without new env vars behave exactly as before — just with richer container labels
- [ ] With `NOMAD_MANAGED_PROJECT=my-nomad-managed` set, new containers go to that project instead — two NOMAD instances on the same host don't fight over the same managed stack

### DIFF
--- a/admin/app/services/docker_service.ts
+++ b/admin/app/services/docker_service.ts
@@ -10,16 +10,19 @@ import { KiwixLibraryService } from './kiwix_library_service.js'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
 import { exec } from 'child_process'
 import { promisify } from 'util'
+import { createHash } from 'crypto'
 // import { readdir } from 'fs/promises'
 import KVStore from '#models/kv_store'
 import { BROADCAST_CHANNELS } from '../../constants/broadcast.js'
 import { KIWIX_LIBRARY_CMD } from '../../constants/kiwix.js'
+import env from '#start/env'
 
 @inject()
 export class DockerService {
   public docker: Docker
   private activeInstallations: Set<string> = new Set()
-  public static NOMAD_NETWORK = 'project-nomad_default'
+  public static NOMAD_NETWORK = env.get('NOMAD_DOCKER_NETWORK', 'project-nomad_default')
+  public static MANAGED_PROJECT = env.get('NOMAD_MANAGED_PROJECT', 'project-nomad-managed')
 
   private _servicesStatusCache: { data: { service_name: string; status: string }[]; expiresAt: number } | null = null
   private _servicesStatusInflight: Promise<{ service_name: string; status: string }[]> | null = null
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
Thanks @maximpedraza0 — careful, well-structured PR with real thought put into it (the env-var isolation for multi-instance hosts especially). Going to close it though, after thinking through the architectural fit.

The core issue: NOMAD's "managed" containers aren't actually a compose project. They're created on-demand by admin code through the Docker API, with no underlying compose file. Adding the full compose identity label set makes them visible to compose-aware tools (`docker compose ps`, Portainer, Unraid Compose Manager), but those tools will then expect compose lifecycle commands (`up`, `down`, `pull`) to work — and they won't, because there's no compose file behind them. That's a worse UX than the current "shows as orphans" state.

Specifically on `com.docker.compose.config-hash`: the PR computes it as `sha256(service_name)` (stable per service), but the real compose semantic is a hash of the resolved service definition that changes when the definition does. Compose tools use it to detect "container is out of sync, recreate needed." Faking it means tools would always think NOMAD's containers are up-to-date even when they aren't.

The right architectural home for real compose identity is the in-progress work to make NOMAD installable as a plain compose stack. At that point the labels emerge naturally from a real compose file rather than being synthesized.

Closing this one. Your PR #776 will be evaluated on its own merits separately. Thanks again for the time and the contribution quality.

PR #776, opened 2026-04-23
feat(collections): bundle specs in image + offline-resilient loading

maximpedraza0:
## Summary

Currently every call to `CollectionManifestService.getSpecWithFallback` must reach `raw.githubusercontent.com/Crosstalk-Solutions/project-nomad/refs/heads/main/collections/*.json` or fall back to a DB-cached spec. A fresh install without internet has no cache to fall back to, so `/easy-setup` can't render category or Wikipedia options until NOMAD reaches GitHub at least once. For an offline-first product, this is a noticeable gap.

Separately, contributors testing a change to one of the spec files today have to either wait for their PR to merge before their code can load the new data, or hand-seed the DB with an edited spec. Both are painful.

## Changes

1. **`Dockerfile`**: `COPY collections /app/collections` — the three spec files ship with the image.

2. **`CollectionManifestService.getLocalSpec`**: reads and validates a spec from that directory. Called from `getSpecWithFallback` as a last-resort fallback when both the remote fetch and the DB cache return nothing. Fresh-install-offline now renders Easy Setup immediately.

3. **`NOMAD_SPEC_BASE_URL` env var**: when set, replaces the hardcoded upstream base URL. The three filename mappings are now table-driven so adding a new spec type later is a one-line change. Useful for testing a fork/branch, running against a mirror, or pointing airgapped installs at an internal mirror.

Default behaviour is unchanged when the env var is unset.

## Test plan

- [ ] Fresh install, internet available → remote fetch still works, DB cache populates as before
- [ ] Fresh install offline → Easy Setup renders immediately using bundled specs
- [ ] Set `NOMAD_SPEC_BASE_URL` to `https://raw.githubusercontent.com/<fork>/project-nomad/refs/heads/<branch>/collections` → remote fetch uses that base
- [ ] Unset `NOMAD_SPEC_BASE_URL` → defaults back to upstream main

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,6 +50,12 @@ COPY package.json /app/version.json
 COPY admin/docs /app/docs
 COPY README.md /app/README.md
 
+# Bundle collection spec files (Wikipedia, Kiwix categories, Maps) so NOMAD
+# remains usable offline when upstream GitHub is unreachable (first install
+# on an airgapped host, or as a fallback when the remote spec is older than
+# what ships with the image).
+COPY collections /app/collections
+
 # Copy entrypoint script and ensure it's executable
 COPY install/entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod +x /usr/local/bin/entrypoint.sh
--- a/admin/app/services/collection_manifest_service.ts
+++ b/admin/app/services/collection_manifest_service.ts
@@ -3,6 +3,8 @@ import vine from '@vinejs/vine'
 import logger from '@adonisjs/core/services/logger'
 import { DateTime } from 'luxon'
 import { join } from 'path'
+import { readFile } from 'fs/promises'
+import env from '#start/env'
 import CollectionManifest from '#models/collection_manifest'
 import InstalledResource from '#models/installed_resource'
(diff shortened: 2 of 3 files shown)


chriscrosstalk:
Thanks @maximpedraza0 — appreciate the thinking here, but going to close this one too after walking through the install model.

The premise this PR is built on — fresh install offline needs to render Easy Setup — doesn't match NOMAD's actual install requirements. A fresh install requires internet. The installer pulls Docker images, model files, and curated content from upstream sources during initial setup; the spec fetch is one of several upstream calls happening at that point. Bundling specs into the image wouldn't unblock an offline fresh install in any meaningful way — the install would already have failed by the time Easy Setup tried to render.

We do have an offline provisioning path on the roadmap, but it's a "clone an already-installed NOMAD server" model rather than "bootstrap from scratch offline." That's planned for a later phase of the broader ecosystem work, and the implementation will look quite different from spec bundling (image + content cloning, not spec embedding).

The `NOMAD_SPEC_BASE_URL` env var alone is a small nice-to-have for fork/mirror testing but not enough to justify the broader change. Closing this one — thanks again for the contribution quality. If you've got other ideas grounded in NOMAD's actual install model, happy to look at them.

PR #780, opened 2026-04-23
feat(maps): regional map downloads via go-pmtiles extract

bgauger:
## What this adds

**Regional Map Download option**
Settings → Maps gets a new option alongside the existing global download. Users pick one or more countries (with an optional max-zoom cap) and the admin extracts just those regions from Protomaps — no 125 GB global download required. The existing "Download Global Map" path is untouched; regions are purely additive.

**Backend: go-pmtiles integration**
- `go-pmtiles v1.30.2` is pinned and bundled in the admin image at `/usr/local/bin/pmtiles` (added in the Dockerfile production stage).
- New BullMQ queue `pmtiles-extract` backed by `RunExtractPmtilesJob`. The job spawns the bundled binary with `--region <geojson>`, polls file size for progress, honors both same-process cancel (SIGTERM) and cross-process cancel (Redis signal key). Final extracts register as `installed_resource` rows. Orphan `.pmtiles` files sharing a `resource_id` prefix are pruned on completion so the storage dir doesn't accumulate partial runs.
- Deterministic job IDs (sha256 of `sourceUrl + regionFilepath + maxzoom`) so duplicate dispatches collapse.
- `POST /api/maps/extract-preflight` runs the same binary with `--dry-run` and returns `{ tiles, bytes, source }` — the UI uses this for the size estimate before the user commits.
- `POST /api/maps/extract` dispatches the job.

**Country data**
- Natural Earth `ne_50m_admin_0_countries.geojson` (3 MB) is checked in under `admin/resources/geodata/` and ships in the production build via `adonisrc.ts` `metaFiles`.
- New `CountriesService` loads and memoizes the dataset, exposes alpha-2 indexed lookup, continent groupings, and a resolver that builds the temporary geojson file passed to `pmtiles extract --region`.
- New endpoints: `GET /api/maps/countries` (alpha-2 + alpha-3 + continent + population), `GET /api/maps/country-groups` (continent-grouped for the picker UI).

**World basemap — grey-screen fix**
- On first `/maps` load, `ensureBaseAssets()` kicks off a one-time extract of a small world-level pmtiles (~15 MB, z ≤ 5) to `storage/maps/pmtiles/world.pmtiles`. The map style layers it underneath regional extracts so tiles outside any installed region render at low zoom instead of showing grey.
- In-flight promise deduplication + a `worldBasemapReady` cache means parallel cold-start requests don't double-spawn the extract.
- The world file is filtered out of `listRegions()` and blocked from deletion (protected system asset).

**Frontend**
- New `CountryPickerModal` component — searchable multi-select grouped by continent with a max-zoom slider. Preflight estimate refreshes on every selection/zoom change so users see the size update live before clicking download.
- New "Regional Map Download" card on the Settings → Maps page.

**Misc**
- `download_service.ts` refactored to share the progress/lifecycle surface across HTTP downloads and pmtiles extracts (same UI, same cancel semantics, same registration hook).
- `commands/queue/work.ts` registers the new queue worker.
- Validators enforce ISO 3166-1 alpha-2 country codes and clamp `maxzoom` to [0, 15].

## Test plan
Build and bring the stack up, then exercise the full loop with a small extract (e.g. Malta at `maxzoom=8` — under 1 MB) from the Settings → Maps UI. That covers preflight, job dispatch, progress polling, file registration, and world-basemap fallback in one pass.


chriscrosstalk (approved):
Tested end-to-end on NOMAD3 with the go-pmtiles binary installed per the Dockerfile pin. Walkthrough:

- Country picker UX (search, continent groups, multi-select, max-zoom) all work
- Live preflight estimate updates correctly with country selection changes
- Small extract validated: Malta @ z=14 (4MB) and UK @ z=15 (2.5GB) both downloaded cleanly
- World basemap fix verified — pan to a region without installed tiles at low zoom shows continent outlines, not grey
- Existing global download path untouched
- HTTP downloads regression-checked (refactored `download_service.ts` shared progress surface works)
- pmtiles binary version verified inside the running container

Found three UX issues during testing — all fixed in follow-up PR #817 (debounce/slider lock, stored-files auto-refresh, already-installed badge). That should land alongside this for a clean v1.32.0 user experience.

This is one of the most impactful community contributions to NOMAD — eliminating the 125 GB Protomaps barrier opens up the offline-maps use case to a much broader audience. Strong approval.

jakeaturner (approved):

PR #783, opened 2026-04-24
Docs/add map marker api

kennethbrewer3:
Updated the admin/docs/api-reference.md file to include the four actions that can be currently done with map markers.


jakeaturner:
Great update, thanks!

Issue #785, opened 2026-04-25
[Feature]: On map, display coordinates as the user moves around the map.


kennethbrewer3:
### Feature Category

UI/UX Improvement

### Problem Statement

I find it difficult and frustrating to add mark markers at a specific set of coordinates. I would like to see the coordinates of where my cursor is on the map.

### Proposed Solution

Add an overlay above the cursor that will display the coordinates as the user moves around the map.

### Who would benefit from this feature?

All Users

### How important is this feature to you?

High - Would significantly improve my experience

### Would you be willing to help implement this?

Yes - I can write the code

### Pre-submission Checklist

- [x] I have searched for existing feature requests that might be similar
- [x] This feature aligns with N.O.M.A.D.'s mission of offline-first knowledge and education
- [x] I understand that feature requests are not guaranteed to be implemented

kennethbrewer3:
PR: https://github.com/Crosstalk-Solutions/project-nomad/pull/786

chriscrosstalk:
Thanks @kennethbrewer3 — sensible UX improvement. PR #786 will be reviewed.

chriscrosstalk:
Closing — feature shipped in v1.32.0-rc.1 (PR #786). Live coordinates display + crosshair cursor are in. Thanks @kennethbrewer3.

PR #786, opened 2026-04-25
feat(maps): show map coordinates on mouse move

kennethbrewer3:
This is an enhancement for the map component. It fulfills the ask of the issue I opened https://github.com/Crosstalk-Solutions/project-nomad/issues/785

* As the user moves the mouse around the map, the latitude and longitude are displayed above the cursor.
* The cursor has been modified to be a crosshair to aid in adding map markers at specific locations.
* When the user is dragging the map to view other regions, the cursor becomes a hand and the coordinates are hidden.
* When the cursor is moved over on screen components like the map marker box, the map scale control, the buttons on the right hand side of the map, and the top bar, the coordinates are hidden.
* Moved the map scale control to an independent coordinate for easier maintenance.

The cursor was hidden in the screenshots unfortunately.

Screenshot 2026-04-25 at 04 24 17

Screenshot 2026-04-25 at 04 24 45


chriscrosstalk (approved):

jakeaturner (approved):

PR #789, opened 2026-04-26
ci: run admin typecheck before build

cuyua9:
## Summary
- run the existing admin `typecheck` script in the pull request build workflow
- keep the typecheck after `npm ci` and before `npm run build`, so TypeScript errors fail before the build step

## Tests
- `git diff --check`
- parsed `.github/workflows/build-admin-on-pr.yml` with Python YAML
- attempted `npm ci` in `admin`; blocked locally by macOS Command Line Tools architecture mismatch while building `@openzim/libzim` native bindings (`libxcrun.dylib` x86_64 vs arm64e), before project scripts ran

### DIFF
--- a/.github/workflows/build-admin-on-pr.yml
+++ b/.github/workflows/build-admin-on-pr.yml
@@ -20,6 +20,10 @@ jobs:
         run: npm ci
         working-directory: ./admin
 
+      - name: Run typecheck
+        run: npm run typecheck
+        working-directory: ./admin
+
       - name: Run build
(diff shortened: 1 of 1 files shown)


cuyua9:
Closing this myself after a value/scope re-check.

The change is technically reasonable, but the diff only adds one CI typecheck step and I do not have a concrete false-negative or full local validation to justify it as a useful standalone contribution. I do not want to leave low-signal CI churn open in your queue. Sorry for the noise.

Issue #796, opened 2026-04-27
Marker popup on map does not render notes field


chriscrosstalk:
Spun out from #768 (API fix in PR #770).

Once `notes` can be set on a marker via the API, the map UI's marker popup should display it. Currently the popup only renders name + coordinates and ignores `notes` even when populated directly via the DB.

Scope: marker popup component in the map viewer. Add a `notes` block (when non-empty) below the existing fields.

PR #802, opened 2026-04-28
feat(maps): add details on marker

kennethbrewer3:
PR https://github.com/Crosstalk-Solutions/project-nomad/pull/786 shares the ScaleToggle component. They can be merged in either order.

Resolves: https://github.com/Crosstalk-Solutions/project-nomad/issues/796

Map enhancements

* Moved the Scale Toggle to its own component
* Moved the Marker Panel to its own component
* Enhanced the Marker Panel to allow for sorting by marker name or the hue of the marker
* Marker Panel allows search by name
* Allows a marker to accept the notes field using markdown
* Allows a marker name, notes, and color to be changed.


voidborne-d:
Nice split — `MapMarkerFormPopup` reused for create + edit is a clean refactor, and the `notes` field flowing all the way from `useMapMarkers` → form → view popup with Markdown rendering reads well. A few things I noticed while reading through:

**1. Textarea auto-grow doesn't fire on mount when editing a marker with long notes.**

In `MapMarkerFormPopup.tsx`, the height adjustment only runs in `onChange`:

```tsx
onChange={(e) => {
  setNotes(e.target.value)
  e.currentTarget.style.height = 'auto'
  e.currentTarget.style.height = `${e.currentTarget.scrollHeight}px`
}}
```

If you open the edit popup on a marker whose notes already span 5+ lines, the textarea renders at `min-h-[64px]` and the user has to scroll inside it (or type a character) to see the full content. A `useLayoutEffect` that runs once on mount with a ref to the textarea would handle the initial render symmetrically with subsequent edits.

**2. Editing in-flight is silently discarded if the user clicks a different marker.**

The marker click handler in `MapComponent.tsx` calls `setEditingMarkerId(null)` unconditionally, which unmounts `MapMarkerFormPopup` and loses any unsaved name/notes edits. Not a bug per se — but with notes potentially being multi-paragraph Markdown, a confirmation when there are pending changes might save accidental loss. At minimum, the cancel/close paths could behave the same way.

**3. Hue-sort buckets grayscale and invalid hex at the same value as pure red.**

`getHueSortValue` returns `0` for `delta === 0` (grayscale) and `0` for non-hex strings, which is also where pure-red lands (`hue ≈ 0`). With the current `PIN_COLORS` palette this isn't visible, but if anyone adds a gray/black pin or a custom CSS color name later, the sort grouping will look surprising. Could either dock grayscale to `-1` so they cluster separately, or sort the `delta === 0` case by lightness as a tiebreaker.

**4. Markdown link safety.**

Good call passing `target="_blank" rel="noopener noreferrer"` in the custom `a` component. `react-markdown` v9 sanitizes `javascript:` URIs at parse time via `urlTransform` before they reach the component, so this is already covered — just flagging since custom-component overrides occasionally regress that protection.

Non-blocking; the structural split is the right move and the merge-with-786 ordering note is helpful. Will defer to maintainer on whether (1)–(3) are in scope here or follow-ups.

— @voidborne-d (post-author of the linked #800)

kennethbrewer3:
> Nice split — `MapMarkerFormPopup` reused for create + edit is a clean refactor, and the `notes` field flowing all the way from `useMapMarkers` → form → view popup with Markdown rendering reads well. A few things I noticed while reading through:
> 
> **1. Textarea auto-grow doesn't fire on mount when editing a marker with long notes.**
> 
> In `MapMarkerFormPopup.tsx`, the height adjustment only runs in `onChange`:
> 
> ```tsx
> onChange={(e) => {
>   setNotes(e.target.value)
>   e.currentTarget.style.height = 'auto'
>   e.currentTarget.style.height = `${e.currentTarget.scrollHeight}px`
> }}
> ```
> 
> If you open the edit popup on a marker whose notes already span 5+ lines, the textarea renders at `min-h-[64px]` and the user has to scroll inside it (or type a character) to see the full content. A `useLayoutEffect` that runs once on mount with a ref to the textarea would handle the initial render symmetrically with subsequent edits.
> 
> **2. Editing in-flight is silently discarded if the user clicks a different marker.**
> 
> The marker click handler in `MapComponent.tsx` calls `setEditingMarkerId(null)` unconditionally, which unmounts `MapMarkerFormPopup` and loses any unsaved name/notes edits. Not a bug per se — but with notes potentially being multi-paragraph Markdown, a confirmation when there are pending changes might save accidental loss. At minimum, the cancel/close paths could behave the same way.
> 
> **3. Hue-sort buckets grayscale and invalid hex at the same value as pure red.**
> 
> `getHueSortValue` returns `0` for `delta === 0` (grayscale) and `0` for non-hex strings, which is also where pure-red lands (`hue ≈ 0`). With the current `PIN_COLORS` palette this isn't visible, but if anyone adds a gray/black pin or a custom CSS color name later, the sort grouping will look surprising. Could either dock grayscale to `-1` so they cluster separately, or sort the `delta === 0` case by lightness as a tiebreaker.
> 
> **4. Markdown link safety.**
> 
> Good call passing `target="_blank" rel="noopener noreferrer"` in the custom `a` component. `react-markdown` v9 sanitizes `javascript:` URIs at parse time via `urlTransform` before they reach the component, so this is already covered — just flagging since custom-component overrides occasionally regress that protection.
> 
> Non-blocking; the structural split is the right move and the merge-with-786 ordering note is helpful. Will defer to maintainer on whether (1)–(3) are in scope here or follow-ups.
> 
> — @voidborne-d (post-author of the linked #800)

Thank you for the review. I will look at these comments and make updates. Good comments!

voidborne-d:
@kennethbrewer3 Looked through. The componentization is good — `ScaleUnitToggle` / `MapMarkerFormPopup` / `ViewMapMarkerPopup` are clean splits and the form-popup unification across create+edit avoids forking that input UI in two places. Few things I'd flag:

**Likely blocker — `admin/inertia/lib/api.ts` not updated for `notes`.** The backend controller + model + vine validator on this branch all support `notes`, but the client wrapper still has the pre-PR signatures:

```ts
async createMapMarker(data: { name: string; longitude: number; latitude: number; color?: string }) { ... }
async updateMapMarker(id: number, data: { name?: string; color?: string }) { ... }
async listMapMarkers(): /* response item type lacks `notes` */
```

The new `useMapMarkers` calls `api.createMapMarker({ name, longitude, latitude, color, notes })` with `notes` as an object-literal property — that's an excess-property error under strict TS, so I'd expect `tsc` (or whatever runs in CI) to fail. Worth adding `notes?: string | null` to all three (request bodies + the response item type the generic `client.get<…>`/`client.post<…>` declares).

Side effect even if TS happens to pass: the response types omit `notes`, so on create/update the post-action `result.notes ?? null` evaluates to `null` from a type-system perspective and the local marker state loses the notes the server just saved — until the next `listMapMarkers()` re-fetch. (At runtime axios still has the field on the JSON, but you're relying on the cast surviving.)

**Minor — `onSave` is fire-and-forget inside the form.** `MapMarkerFormPopup.handleSave` calls `onSave(...)` without `await`, and the Save button isn't disabled while the parent's `addMarker`/`updateMarker` is in flight. Network slow / failure → users can double-click and create dupes, or the popup just sits there with no feedback. Threading an `isSaving` state through (or letting the parent unmount the popup synchronously then handle persistence) tightens it up.

**Nit — security boundary on notes markdown.** With `react-markdown@^10.1.0` the default `urlTransform` blocks `javascript:` etc., and you're not enabling `rehype-raw`, so raw HTML is dropped. So this is fine as-is — flagging only because the moment someone adds `rehypeRaw` to allow inline HTML in notes, the popup becomes XSS-able. A short comment in `ViewMapMarkerPopup` ("don't add rehypeRaw without sanitization") would future-proof it.

**Cosmetic** — the multi-line destructured props in `MapMarkerFormPopup` and `MarkerPanel` ended up with this odd hanging-indent shape (`function Foo({\n                                             longitude,`). Looks like a Prettier disagreement; if the repo's formatter is configured a particular way it'll re-indent on next pass.

The hue-sort in `MarkerPanel` (`getHueSortValue` falling back to 0 for grayscale or invalid hex via `normalizeColorHex`) is nice — does the right thing for both preset ids and any free-form hex you might allow later.

LGTM once the api.ts types catch up to the controller. Happy to re-look after that change.

kennethbrewer3:
@voidborne-d  I addressed your PR comments. Extremely useful and thorough feedback. Good call on requiring confirmation before canceling the marker edit. I have also altered the color sort to sort by lightness and then by hue. That should prevent collisions between grayscale and red.

voidborne-d:
@kennethbrewer3 Awesome — thanks for taking the feedback so seriously, and the lightness-then-hue tweak for the color sort is a nice fix for the grayscale-vs-red collision. LGTM from my side.

kennethbrewer3:
> @kennethbrewer3 Awesome — thanks for taking the feedback so seriously, and the lightness-then-hue tweak for the color sort is a nice fix for the grayscale-vs-red collision. LGTM from my side.

You're very welcome. I really liked the collaboration. I appreciated your attention to detail and your explanations. I'm not a web developer so I'm learning on the fly. I develop mostly in Flutter and Android. I dabble some on the iOS side. 

Great callout about the way the colors might sort unexpectedly.

chriscrosstalk (approved):

chriscrosstalk:
Heads-up for the merge: this conflicts with #786 when merged after it (both restructure `MapComponent.tsx` along different axes; both introduce `ScaleUnitToggle`).

I tested the four-PR bundle (#786 → #802 → #806 → #813) end-to-end on NOMAD3 and pushed the resolved integration as `test/v1.32-maps-bundle`. The merge resolution is captured in commit [54cd133](https://github.com/Crosstalk-Solutions/project-nomad/commit/54cd133) — full notes in the commit body. TL;DR:

- **ScaleUnitToggle.tsx**: combined #786's `onMouseEnter` prop (essential for hide-coords-over-controls UX) with #802's `cursor-pointer` + `if (scaleUnit !== X)` guards. Kept Tailwind classes (codebase convention).
- **MapComponent.tsx**: layered #786's cursor/coords/dragging UX on top of #802's componentized marker structure (`MapMarkerFormPopup`, `ViewMapMarkerPopup`). Dropped #786's now-superseded inline marker form. Added `.maplibregl-popup` to control selectors so hide-coords UX extends to popups.

Feel free to cherry-pick from the integration branch or use it as a reference.

voidborne-d:
Nice — combining #786's `onMouseEnter` prop with #802's componentized marker structure is the right call, and extending the hide-coords selector to `.maplibregl-popup` so it covers the new `MapMarkerFormPopup` / `ViewMapMarkerPopup` is a thoughtful follow-on. The integration branch reads cleanly.

jakeaturner:
Hey @kennethbrewer3,  first off, thank you for all the work you've been putting into these PRs!

#786  has been merged into `dev` 🎉. For the remaining three (#802 , #806 , and #813 ), I started taking a look at resolving the merge conflicts myself, but I've run into a snag: because all four PRs touch many of the same files, they're stepping on each other pretty significantly. Combined with many commits per PR, it's become really difficult for me to untangle the original intent of each change without potentially breaking something.

Rather than risk losing context or introducing regressions, I'd like to ask if you could do the following:

1. Close #802, #806, and #813 (no work lost, the commits will still exist on your branches)
2. Create a single new branch off the latest `dev`
3. Cherry-pick or re-apply your changes from those three PRs onto that branch, rebased cleanly on `dev`
4. Open one new PR with all of those changes consolidated

A few tips to make future PR's smoother:
- **When PRs touch the same files, it's usually best to either wait for the earlier PR to merge before opening the next one, or just keep adding commits to the existing PR.** Parallel PRs on overlapping files almost always end up in conflict hell.
- **Try to group related changes into fewer, more meaningful commits** rather than many micro-commits. A good rule of thumb: each commit should represent a logical unit of work that you'd want to be able to revert or review on its own.

I genuinely appreciate the contributions and that you're breaking work into pieces (that's usually exactly what we want from contributors), the issue here is just that the pieces are overlapping rather than being independent.

kennethbrewer3:
> Hey @kennethbrewer3, first off, thank you for all the work you've been putting into these PRs!
> 
> #786 has been merged into `dev` 🎉. For the remaining three (#802 , #806 , and #813 ), I started taking a look at resolving the merge conflicts myself, but I've run into a snag: because all four PRs touch many of the same files, they're stepping on each other pretty significantly. Combined with many commits per PR, it's become really difficult for me to untangle the original intent of each change without potentially breaking something.
> 
> Rather than risk losing context or introducing regressions, I'd like to ask if you could do the following:
> 
> 1. Close [feat(maps): add details on marker #802](https://github.com/Crosstalk-Solutions/project-nomad/pull/802), [fix(maps): Map Marker Point Inaccurate #806](https://github.com/Crosstalk-Solutions/project-nomad/pull/806), and [feat(maps): Several user experience improvements #813](https://github.com/Crosstalk-Solutions/project-nomad/pull/813) (no work lost, the commits will still exist on your branches)
> 2. Create a single new branch off the latest `dev`
> 3. Cherry-pick or re-apply your changes from those three PRs onto that branch, rebased cleanly on `dev`
> 4. Open one new PR with all of those changes consolidated
> 
> A few tips to make future PR's smoother:
> 
> * **When PRs touch the same files, it's usually best to either wait for the earlier PR to merge before opening the next one, or just keep adding commits to the existing PR.** Parallel PRs on overlapping files almost always end up in conflict hell.
> * **Try to group related changes into fewer, more meaningful commits** rather than many micro-commits. A good rule of thumb: each commit should represent a logical unit of work that you'd want to be able to revert or review on its own.
> 
> I genuinely appreciate the contributions and that you're breaking work into pieces (that's usually exactly what we want from contributors), the issue here is just that the pieces are overlapping rather than being independent.

Excellent feedback. I will be able to handle this later this evening. In the future would it be helpful to squash commits before pushing?

PR #806, opened 2026-04-29
fix(maps): Map Marker Point Inaccurate

kennethbrewer3:
Map markers are now their own SVG that have an accurate point to the coordinate. Also the map marker accepts an icon parameter for future UI improvement. Currently the default map icon is a white circle.


chriscrosstalk (approved):

PR #823, opened 2026-05-04
security: move hardcoded HMAC secret out of benchmark_service.ts (CWE-798)

sebastionoss:
## Summary

`admin/app/services/benchmark_service.ts` contained a hardcoded 48-character HMAC secret used to sign benchmark submissions to the central repository API at `benchmark.projectnomad.us`:

```ts
const BENCHMARK_HMAC_SECRET = '778ba65d0bc0e23119e5ffce4b3716648a7d071f0a47ec3f'
```

Because this repository is public, anyone can read the secret from source (or from `git` history) and forge valid HMAC signatures for arbitrary submissions to the upstream collector. The existing comment in the code already acknowledged the risk ("a determined attacker could extract this"), so this PR finishes the job by moving the secret out of source.

- **CWE:** [CWE-798: Use of Hard-coded Credentials](https://cwe.mitre.org/data/definitions/798.html)
- **Affected file/function:** `admin/app/services/benchmark_service.ts` — `BenchmarkService.submitToRepository()`
- **Severity:** High (integrity of upstream benchmark data; spam/DoS surface against the collector)

### Data flow

1. The constant `BENCHMARK_HMAC_SECRET` is defined at module scope in `benchmark_service.ts`.
2. `submitToRepository()` calls `createHmac('sha256', BENCHMARK_HMAC_SECRET).update(timestamp + JSON.stringify(submission))` and sends the digest as `X-Benchmark-Signature` to `benchmark.projectnomad.us`.
3. The upstream API trusts that signature as proof the submission came from a genuine NOMAD instance.

Since the key is the same constant in every clone of this repo, that trust assumption is already broken.

## Fix

Three small changes, all in `admin/`:

1. **`app/services/benchmark_service.ts`** — replace the hardcoded literal with `env.get('BENCHMARK_HMAC_SECRET')`. In `submitToRepository()`, fail closed with a clear error if the secret isn't configured, so a missing env var can never silently downgrade to "unsigned" or "default-signed" requests.
2. **`start/env.ts`** — register `BENCHMARK_HMAC_SECRET` in the AdonisJS env schema as an optional string. Optional (rather than required) keeps the admin app bootable for users who don't opt into benchmark submission; the runtime check in step 1 still blocks actual sending.
3. **`.env.example`** — document the new variable, with a generation hint (`openssl rand -hex 24`). The example is left commented out so operators must explicitly opt in.

No fallback path retains the old constant. The HMAC algorithm, payload format, and headers are unchanged, so the benchmark server side requires no changes — operators just need to configure the same secret on both ends.

## Testing

- `cd admin && npm run typecheck` — passes; the new `env.get('BENCHMARK_HMAC_SECRET')` call is correctly typed as `string | undefined` and the runtime guard narrows it before use.
- Verified by reading call sites that `BENCHMARK_HMAC_SECRET` is referenced only inside `submitToRepository()` and only after the `if (!BENCHMARK_HMAC_SECRET)` guard, so an unset env var produces a thrown error rather than a malformed/empty signature.
- Confirmed via grep that no other module imported the old constant.

## Security analysis

The signing key is the only authentication on submissions to the benchmark collector. With the value embedded in a public repo:

- Any third party can craft `X-Benchmark-Signature` headers that the upstream API accepts as authentic.
- They can submit poisoned or fabricated benchmark data, skewing whatever aggregate stats or rankings the collector exposes.
- They can also use the verified channel for low-cost spam/DoS against the collector.

After the fix, the secret lives only in operator-controlled environment configuration. A new clone of this repo carries no usable signing material, and a misconfigured deployment refuses to submit rather than emitting a forgeable request.

**Important operational note for maintainers:** the leaked value `778ba65d0bc0e23119e5ffce4b3716648a7d071f0a47ec3f` is still present in `git` history and must be considered compromised. Please rotate the key on the benchmark server (`benchmark.projectnomad.us`) and distribute the new value to operators out of band; this PR cannot do that for you.

### Adversarial review

Before submitting, we tried to talk ourselves out of this finding. The main counter-arguments would be: (a) the secret is "just" anti-abuse and the collector has other defenses, or (b) being able to read a public repo already grants equivalent capability. Neither holds — read access to the source does not by itself let you forge a valid `X-Benchmark-Signature` (you still need the key material, which this PR removes from source), and the in-code comment makes clear the secret is the intended authentication boundary, not a decoration over a stronger one. Moving it to env is the minimal fix that actually restores that boundary.

cc @lewiswigmore

### DIFF
--- a/admin/.env.example
+++ b/admin/.env.example
@@ -16,3 +16,6 @@ REDIS_PORT=6379
 # On Windows dev, use an absolute path like: C:/nomad-storage
 # On Linux production, use: /opt/project-nomad/storage
 NOMAD_STORAGE_PATH=/opt/project-nomad/storage
+# HMAC secret used to sign benchmark submissions to benchmark.projectnomad.us
+# Generate with: openssl rand -hex 24
+# BENCHMARK_HMAC_SECRET=
--- a/admin/app/services/benchmark_service.ts
+++ b/admin/app/services/benchmark_service.ts
@@ -23,16 +23,17 @@ import type {
   RepositoryStats,
 } from '../../types/benchmark.js'
 import { randomUUID, createHmac } from 'node:crypto'
+import env from '#start/env'
 import { DockerService } from './docker_service.js'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
 import { BROADCAST_CHANNELS } from '../../constants/broadcast.js'
 import Dockerode from 'dockerode'
 
-// HMAC secret for signing submissions to the benchmark repository
-// This provides basic protection against casual API abuse.
-// Note: Since NOMAD is open source, a determined attacker could extract this.
-// For stronger protection, see challenge-response authentication.
-const BENCHMARK_HMAC_SECRET = '778ba65d0bc0e23119e5ffce4b3716648a7d071f0a47ec3f'
+// HMAC secret for signing submissions to the benchmark repository.
+// Must be provided via the BENCHMARK_HMAC_SECRET environment variable.
+// The benchmark server uses this to verify that submissions originate from
+// a genuine NOMAD instance.  Never commit the real secret to source control.
+const BENCHMARK_HMAC_SECRET = env.get('BENCHMARK_HMAC_SECRET')
 
 // Re-export default weights for use in service
 const SCORE_WEIGHTS = {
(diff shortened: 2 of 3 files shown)


chriscrosstalk:
Thanks @sebastiondev for the detailed writeup and the adversarial review — the analysis is correct and the fix is well-constructed. Closing here, but not because the finding isn't real.

The pragmatic issue is that any client-side change here just moves the shared secret around. Embedding the value in `install_nomad.sh` or the management compose template would put it back in distributed source; requiring operators to set `BENCHMARK_HMAC_SECRET` themselves would silently break benchmark submissions for the existing install base; and any auto-fetch mechanism at install time still hands the secret out to anyone who can run the install script.

The original code comment already pointed at this: "for stronger protection, see challenge-response authentication." The honest threat model is that HMAC-with-a-shared-secret was always a speed bump against casual abuse, not a real auth boundary — the actual defense is the server-side validation in the (now-private) benchmark repo. Closing this finding properly means a per-install enrollment + server-side challenge-response protocol, which is a separate piece of work that needs to happen on the server side first.

Really appreciate the time you put into this.

PR #837, opened 2026-05-06
fix(docker): reject failed image pulls

### LINKED ISSUE #790: Handle Docker pull errors from followProgress
Follow-up from #658.\n\n## Problem\nSeveral Docker image pulls wait on dockerode `followProgress` by passing only the resolve callback:\n\n```ts\nawait new Promise((res) => this.docker.modem.followProgress(pullStream, res))\n```\n\n`followProgress` calls back with `(err, output)`. If the pull fails, the error can be treated as the resolved value, allowing install/update flows to continue as if the image pull succeeded.\n\n## Current locations\n- `admin/app/services/docker_service.ts` during service image pulls\n- `admin/app/services/docker_service.ts` in `updateContainer`\n- `admin/app/services/benchmark_service.ts` in sysbench image setup\n\n## Suggested fix\nWrap `followProgress` with explicit reject handling:\n\n```ts\nawait new Promise((resolve, reject) => {\n  this.docker.modem.followProgress(pullStream, (err, output) => {\n    if (err) reject(err)\n    else resolve(output)\n  })\n})\n```\n\nThis should make Docker pull failures surface normally to the existing error handling paths instead of being swallowed.\n\n## Notes\nThis is one of the patch-level follow-ups split out from #658, per maintainer feedback.

juyua9:
## Summary
- add a shared Docker image pull helper that rejects `followProgress` errors
- use it for service installs, service updates, and sysbench benchmark image pulls

Closes #790

## Testing
- `npm run typecheck` (from `admin/`)
- `git diff --check`

### DIFF
--- a/admin/app/services/benchmark_service.ts
+++ b/admin/app/services/benchmark_service.ts
@@ -614,8 +614,7 @@ export class BenchmarkService {
       await this.dockerService.docker.getImage(SYSBENCH_IMAGE).inspect()
     } catch {
       this._updateStatus('starting', `Pulling sysbench image...`)
-      const pullStream = await this.dockerService.docker.pull(SYSBENCH_IMAGE)
-      await new Promise((resolve) => this.dockerService.docker.modem.followProgress(pullStream, resolve))
+      await this.dockerService.pullImage(SYSBENCH_IMAGE)
     }
   }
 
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
Thanks for flagging this @juyua9 — you're right that the pull error needed to be handled. We've implemented the complete fix in #983, which routes every pull site through a shared helper that rejects on a failed pull, including the AMD ROCm pull and the force-reinstall path. Closing this in favor of that. Appreciate the contribution.

PR #836, opened 2026-05-06
feat(maps): maps marker management

kennethbrewer3:
## Summary

This PR introduces major improvements to map marker management, UI organization, and navigation workflows. It refactors key components, enhances marker functionality, and adds new user controls for filtering, sorting, and navigation.

---

## Component Refactoring

- Extracted **Scale Selector** into its own component
- Extracted **Marker Panel** into its own component

---

### Editing & Metadata
- Markers now support:
  - Editable **name**
  - Editable **notes** (with Markdown support)
  - Custom **color**
  - Custom **icon** (foundation for future UI improvements)

### Rendering Improvements
- Markers are now rendered as **custom SVGs**
  - Improved visual accuracy (true coordinate anchoring)
  - Default icon is a white circle
  - Icon color is automatically adjusted for contrast against the marker color

---

### Search & Sorting
- Search markers by **name**
- Sort markers by:
  - Name
  - Hue (color)
  - Visibility
  - Icon (alphabetical)

### Visibility Controls
- Toggle visibility of individual markers
- Added **"Hide All"** control

### Organization (Tree View)
- Introduced a **tree view** for markers
- Group markers by:
  - First character of name
  - Hue
  - Visibility
  - Icon
- Added controls to:
  - Expand all
  - Collapse all

---

### Navigation
- Support URL parameters for map positioning: /maps?lat=38.910327&lng=-77.263447&zoom=20
- Added search input for `lat,lng`
- Added **"Fly To"** functionality:
- Moves map to specified coordinates
- Drops a temporary marker for visual feedback

### Marker Creation
- Added ability to:
- Create a marker directly from entered coordinates

---

## Database Changes

- Added migration to extend marker schema:
- Supports new fields (color, icon, notes, etc.)
- Maintains backward compatibility with existing markers

---

## UX Improvements

- Improved marker visibility management when zoomed out (reduces overlap clutter)
- Clear visual feedback when navigating to coordinates
- Better organization and scalability for large numbers of markers

---

## Notes / Future Work

- Icon system is in place for future expansion (additional icon sets, UI picker, etc.)
- Temporary navigation marker could be extended with additional behaviors (e.g., fade out, convert to saved marker)

PR #851, opened 2026-05-08
fix(map): show marker notes in popup

### LINKED ISSUE #796: Marker popup on map does not render notes field

Gujiassh:
## Summary
- preserve marker `notes` in the front-end map marker state
- render non-empty marker notes in the selected marker popup
- add a small unit test covering marker notes normalization from the API payload

## Testing
- `cd admin && npm run typecheck`

## Issue
- fixes #796

### DIFF
--- a/admin/inertia/components/maps/MapComponent.tsx
+++ b/admin/inertia/components/maps/MapComponent.tsx
@@ -166,7 +166,14 @@ export default function MapComponent() {
             onClose={() => setSelectedMarkerId(null)}
             closeOnClick={false}
           >
-            <div className="text-sm font-medium">{selectedMarker.name}</div>
+            <div className="space-y-1">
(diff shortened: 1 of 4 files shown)


kennethbrewer3:
This has been handled on 

https://github.com/Crosstalk-Solutions/project-nomad/pull/836

Gujiassh:
Closing this as superseded by #836, which already landed the marker popup notes fix. Keeping the duplicate lane out of the active queue.

PR #857, opened 2026-05-08
fix(security): enable HSTS in production

cuyua9:
## Summary
- enable the Adonis Shield HSTS header only in production
- reuse the existing `app.inProduction` runtime guard pattern used elsewhere in admin config
- keep local/development behavior unchanged

## Tests
- `git diff --check`
- Not run: `npm run typecheck` / build because installing admin dependencies is currently blocked by external native dependency downloads in this environment (`@openzim/libzim` / Node headers)

No issue filed because this is a narrow production security config fix for an existing TODO.

### DIFF
--- a/admin/.env.example
+++ b/admin/.env.example
@@ -10,6 +10,8 @@ DB_USER=root
 DB_DATABASE=nomad
 DB_PASSWORD=password
 DB_SSL=false
+# Enable only when the admin interface is served over HTTPS and you want browsers to enforce HSTS
+ENABLE_HSTS=false
 REDIS_HOST=localhost
 REDIS_PORT=6379
 # Storage path for NOMAD content (ZIM files, maps, etc.)
--- a/admin/config/shield.ts
+++ b/admin/config/shield.ts
@@ -1,4 +1,5 @@
 import { defineConfig } from '@adonisjs/shield'
+import app from '@adonisjs/core/services/app'
 
 const shieldConfig = defineConfig({
   /**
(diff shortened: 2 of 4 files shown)


jakeaturner (changes requested):
@cuyua9  thanks for the work here! We wouldn't want to enable HSTS in prod by default, however, because the vast majority of NOMAD users are just accessing directly via LAN and won't have HTTPS configured. We could, however, make this an environment variable that can be passed, but default to `false`

cuyua9:
Follow-up pushed: `ff60b576`.

I changed HSTS to be opt-in instead of enabling it for every production deployment:

- added `ENABLE_HSTS` as an optional boolean environment variable
- kept the default value false in the example env and compose config
- `admin/config/shield.ts` now enables HSTS only when the app is in production and `ENABLE_HSTS=true`

This keeps LAN/http deployments unchanged by default, while still allowing HTTPS deployments to opt in explicitly.

Validation:

```bash
git diff --check origin/dev...HEAD
```

I also attempted `npm run typecheck` under `admin/`, but this local worktree does not have the dependency/type environment installed, so it fails broadly on missing Adonis/Node modules and tsconfig base types before reaching this change. The remote Build Admin check is running on the new head now.

chriscrosstalk:
Thanks @cuyua9. The opt-in shape (env-gated, default off, scoped to production) is well-designed and wouldn't break existing users.

That said, NOMAD is intentionally a LAN HTTP appliance. Most installs will never serve over HTTPS, and HSTS is a footgun in that environment (once a browser caches it, the 180-day max-age effectively locks the user into HTTPS for that hostname/IP). The existing `// TODO: Enable HSTS in production` comment in `shield.ts` was a placeholder rather than an active intent.

For users who do want HTTPS, the right path is fronting NOMAD with a reverse proxy. We have an official guide for that at https://www.projectnomad.us/install/caddy. Caddy handles cert acquisition + renewal and can set HSTS itself, which is the cleaner separation of concerns.

Closing this PR.

PR #593, opened 2026-03-30
feat(content): custom ZIM library sources with pre-seeded mirrors

chriscrosstalk:
## Summary
- Adds custom ZIM library browsing to Content Explorer, allowing users to download from faster Kiwix mirrors or self-hosted repos
- 5 official Kiwix mirrors pre-seeded out of the box (Debian CDN, Your.org US, FAU Germany, Dotsrc Denmark, MirrorService UK)
- Browsable directory structure with breadcrumb navigation, scrollable table with sticky header
- Source selection persists across page loads

## What it does
Next to "Browse the Kiwix Library", there's now an "Add Custom Library" / "Manage Custom Libraries" button. A source dropdown lets users switch between Default (Kiwix) and any custom library. When browsing a custom library, users see folders and .zim files in a familiar file browser layout, with Download buttons that feed into the existing download pipeline.

## Why
Users reported slow downloads from the default Kiwix CDN (issue #576). Rather than the compose env var approach in PR #577, this keeps everything in the GUI where users can actually find it.

## Test plan
- [ ] Navigate to Content Explorer
- [ ] Verify 5 pre-seeded mirrors appear in the Source dropdown
- [ ] Select a mirror, browse into directories, verify ZIM files appear
- [ ] Download a ZIM from a custom mirror, verify progress and completion
- [ ] Switch back to Default (Kiwix), verify OPDS search works unchanged
- [ ] Refresh the page, verify source selection persists
- [ ] Open Manage modal, verify built-in mirrors can't be deleted
- [ ] Add a custom library, verify it appears in dropdown
- [ ] Remove a custom library, verify it's gone

Closes #576

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #859, opened 2026-05-08
fix(Downloads): treat missing Content-Type as octet-stream (#848)

### LINKED ISSUE #848: [Bug]: Downloads fail with `MIME type  is not allowed` when source response is missing a Content-Type header
### Issue Category

Docker/Container Issues

### Summary

The `downloads` queue worker rejects valid files when the source server's HTTP response doesn't include a `Content-Type` header. This makes any ZIM file hosted on `download.kiwix.org` undownloadable through NOMAD, since that server doesn't set Content-Type for `.zim` files.

### Environment

- NOMAD: latest (as of May 2026)
- Host OS: ZimaOS
- Deployment: Docker (compose)

### Steps to reproduce

1. Open the Content Manager and queue any ZIM file that NOMAD's catalog points to on `download.kiwix.org`. In my case: `devdocs_en_python_2026-05.zim` and `devdocs_en_node_2026-05.zim`.
2. The download appears in Active Downloads with a `Waiting...` status indefinitely.
3. The download never starts, never errors visibly in the UI, and never clears.

### Expected behavior

The file downloads and is registered with Kiwix, since the source URL returns a valid `200 OK` and a complete file.

### Actual behavior

The job is repeatedly picked up and immediately failed by the `downloads` worker. From `nomad_admin` logs:

```
[ info ] [downloads] Processing job: 355959c7f9730e3a of type: run-download
[ info ] [downloads] Processing job: 06c1c12d5ad0757a of type: run-download
[ error ] [downloads] Job failed: 355959c7f9730e3a, Error: MIME type  is not allowed
[ error ] [downloads] Job failed: 06c1c12d5ad0757a, Error: MIME type  is not allowed
```

Note the empty value between `MIME type` and `is not allowed` — the validator is comparing an empty string against the allowlist.

### Root cause

The Kiwix download mirror returns `200 OK` with the file and correct `Content-Length`, but does not set a `Content-Type` header at all:

```
$ curl -sI https://download.kiwix.org/zim/devdocs/devdocs_en_python_2026-05.zim
HTTP/1.1 200 OK
Date: Thu, 07 May 2026 13:23:58 GMT
Content-Length: 4300242
Connection: keep-alive
Last-Modified: Mon, 04 May 2026 14:48:40 GMT
ETag: "419dd2-650ff06c4c78f"
Accept-Ranges: bytes
Access-Control-Allow-Origin: *
Strict-Transport-Security: max-age=31536000; includeSubDomains
```

NOMAD's download worker treats the missing header as an empty MIME type and rejects the response against its allowlist, even though the URL, file extension, and file content are all valid.

### Suggested fix

Have the download worker fall back to file-extension validation when `Content-Type` is missing or empty, rather than rejecting outright. Since the file extension is already known from the catalog entry / URL, a `.zim` extension could be treated as equivalent to receiving `application/octet-stream`.

### Workaround

For anyone hitting this in the meantime: download the file manually with `curl` into the storage zim folder, then register it with Kiwix using `kiwix-manage` (using `--user 0:0` to avoid container permission issues on the library file), then restart `nomad_admin` so its reconciliation picks up the new file. The old version, if any, needs to be removed manually from both the Kiwix library (`kiwix-manage remove <book_id>`) and disk.

### Steps to Reproduce

1. Open the Content Manager and queue any ZIM file that NOMAD's catalog points to on `download.kiwix.org`. In my case: `devdocs_en_python_2026-05.zim` and `devdocs_en_node_2026-05.zim`.
2. The download appears in Active Downloads with a `Waiting...` status indefinitely.
3. The download never starts, never errors visibly in the UI, and never clears.

### Expected Behavior

The file downloads and is registered with Kiwix, since the source URL returns a valid `200 OK` and a complete file.

### Actual Behavior

The job is repeatedly picked up and immediately failed by the `downloads` worker. From `nomad_admin` logs:

```
[ info ] [downloads] Processing job: 355959c7f9730e3a of type: run-download
[ info ] [downloads] Processing job: 06c1c12d5ad0757a of type: run-download
[ error ] [downloads] Job failed: 355959c7f9730e3a, Error: MIME type  is not allowed
[ error ] [downloads] Job failed: 06c1c12d5ad0757a, Error: MIME type  is not allowed
```

Note the empty value between `MIME type` and `is not allowed` — the validator is comparing an empty string against the allowlist.

### N.O.M.A.D. Version

1.31.1

### Operating System

Other (not yet officially supported)

### Docker Version

_No response_

Yes

### GPU Model (if applicable)

RTX 5060ti

### System Specifications

_No response_

_No response_

### Relevant Logs

```shell

```

```javascript

```

### Screenshots

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

- [ ] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

bgauger:
## What

Fixes #848. ZIM (and any other binary) downloads from upstream hosts that don't set a `Content-Type` header now succeed instead of failing with `MIME type  is not allowed`.

## Why

The validator in `admin/app/utils/downloads.ts` reads `headers['content-type'] || ''`, then runs each entry of `allowedMimeTypes` through `contentType.includes(mt)`. When the header is missing, `''.includes('application/octet-stream')` is `false`, so every allowed type fails to match and the download throws.

`download.kiwix.org` and several of its mirrors don't always set Content-Type on `.zim` responses. With NOMAD pointing at the Kiwix catalog by default, this can break ZIM downloads outright. Choosifur reported the issue with `nomad_admin` logs showing the empty MIME error verbatim.

RFC 7231 §3.1.1.5 says missing Content-Type may be treated as `application/octet-stream`. That MIME is already in every binary-content allowlist we use:

- `ZIM_MIME_TYPES` (collection_update_service.ts:17, zim_service.ts:34, docker_service.ts:746-750)
- `PMTILES_MIME_TYPES` (collection_update_service.ts:18, map_service.ts:67)
- `BASE_ASSETS_MIME_TYPES` (map_service.ts:59)

So defaulting empty to octet-stream keeps existing behavior for every caller while letting Kiwix downloads through.

## How

One-line change to `admin/app/utils/downloads.ts:50` plus a comment. `headers['content-type'] || ''` becomes `headers['content-type'] || 'application/octet-stream'`.

## Verification

Logic-checked against 9 cases including:
- Mirror sets `application/octet-stream` (passes, unchanged)
- Mirror omits Content-Type entirely (now passes, was the bug)
- Empty Content-Type string (now passes)
- Wrong content type like `text/html` for PMTILES (still rejected)
- Strict allowlist without octet-stream + missing Content-Type (still rejects, no over-relaxation)
- Content-Type with charset suffix (passes, existing behavior)

Also confirmed against live Kiwix mirrors via `curl -sIL` — most set `application/octet-stream` correctly, but the redirect-by-geo logic on `download.kiwix.org` can land users on mirrors with inconsistent header behavior.

## Notes

- No new permissive matching beyond what the existing `contentType.includes(mimeType)` substring check already does.
- Doesn't change the error format or any other validator behavior — same `MIME type X is not allowed` shape, just less likely to fire spuriously.
- Independent of the upcoming custom-mirror feature (#593) — that gives users a workaround by picking a mirror that sets headers, this fixes the underlying validator so any well-formed response works.

### DIFF
--- a/admin/app/utils/downloads.ts
+++ b/admin/app/utils/downloads.ts
@@ -47,7 +47,14 @@ export async function doResumableDownload({
     timeout,
   })
 
-  const contentType = headResponse.headers['content-type'] || ''
+  // Some upstream hosts (notably download.kiwix.org for .zim files) don't set a
+  // Content-Type header at all. Per RFC 7231 §3.1.1.5, "if no Content-Type is
+  // provided" the recipient may treat it as application/octet-stream — which is
+  // already in every binary-content allowlist we use (ZIM, PMTILES, base assets).
+  // Without this default, the validator below throws `MIME type  is not allowed`
+  // and breaks all downloads from kiwix's primary host (#848).
+  const contentType =
+    headResponse.headers['content-type'] || 'application/octet-stream'
   const totalBytes = parseInt(headResponse.headers['content-length'] || '0')
   const supportsRangeRequests = headResponse.headers['accept-ranges'] === 'bytes'
 


jakeaturner (approved):
Great catch, thanks!

PR #861, opened 2026-05-09
fix(install): harden toolkit and helper script downloads

### LINKED ISSUE #807: install_nomad.sh: silent NVIDIA Container Toolkit failure + no curl retry on helper script downloads
Two independent bugs in `install_nomad.sh` that surfaced during a fresh install on Ubuntu 24.04 today. Both are small fixes; filing together since they're scoped to the same script.

## Bug 1: `gpg --dearmor` swallows TTY failure → NVIDIA Container Toolkit silently not installed

In `setup_nvidia_container_toolkit()` (around line 267), the `gpg --dearmor` step can fail when there's no controlling TTY — newer `gpg` versions try to open `/dev/tty` for status output even during non-interactive operations. When that happens, the script prints:

```
Warning: Failed to add NVIDIA container toolkit GPG key. Continuing anyway...
```

…and then continues. End result: `nvidia-container-toolkit` is **not** installed, the `nvidia` Docker runtime is **not** registered, and the verification block at the end reports:

```
✓ NVIDIA GPU detected: NVIDIA GeForce RTX 3090 Ti, 24564 MiB
○ NVIDIA Container Toolkit not installed
○ Docker NVIDIA runtime not detected
GPU acceleration not detected. The AI Assistant will run in CPU-only mode.
Tip: Your GPU is detected but Docker runtime is not configured.
Try restarting Docker: sudo systemctl restart docker
```

The "Tip" doesn't help because the toolkit was never installed. The user has an NVIDIA GPU but ends up with CPU-only inference and no clear signal that the silent downgrade happened.

**Fix:** add `--batch --yes` to the gpg call:

```bash
sudo gpg --batch --yes --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg
```

Reproduced and verified — that one change makes the toolkit install cleanly when run via SSH (no controlling TTY).

**Severity:** moderate. Silent downgrade to CPU-only is the worst failure mode for this path — affected users get poor AI performance with no clear indication anything went wrong.

## Bug 2: Helper script download has no curl retry, transient network blip = partial install

After Docker is installed, the script downloads the start/stop/update helper scripts from `raw.githubusercontent.com`. A transient DNS or network failure on that step kills the install:

```
curl: (6) Could not resolve host: raw.githubusercontent.com
Failed to download the start script. Please check the URL and try again.
```

`/opt/project-nomad/` is left half-populated, requiring `sudo rm -rf /opt/project-nomad` before re-running.

**Fix:** add `--retry 5 --retry-delay 3` to the helper-script curl calls. `raw.githubusercontent.com` has occasional CDN flakes; `curl --retry` handles them transparently and the install would have proceeded.

**Severity:** minor (flake-dependent), but the failure mode is annoying — half-installed state needs manual cleanup before retry.

## Reproduction

1. Fresh Ubuntu 24.04 with NVIDIA GPU + driver
2. Run install_nomad.sh per README
3. **Bug 1:** consistently reproduces over SSH (no TTY) — observe the GPG warning, then CPU-only verification
4. **Bug 2:** less reproducible (depends on CDN); we hit it once during the same install session

## Environment

- Ubuntu 24.04.4 LTS, kernel 6.17.0-22-generic
- NVIDIA RTX 3090 Ti, driver 595.58.03
- Ryzen Threadripper 3960X
- NOMAD v1.31.1, `install_nomad.sh` from `main`

Gujiassh:
## Summary
- add `--batch --yes` to the NVIDIA toolkit GPG dearmor step so non-TTY installs do not silently skip toolkit setup
- add curl retries to helper script downloads to reduce transient partial-install failures

## Testing
- validated the install script updates by checking the new `gpg` and `curl --retry` flags in `install/install_nomad.sh`

## Issue
- fixes #807

### DIFF
--- a/install/install_nomad.sh
+++ b/install/install_nomad.sh
@@ -249,7 +249,7 @@ setup_nvidia_container_toolkit() {
   echo -e "${YELLOW}#${RESET} Installing NVIDIA container toolkit...\\n"
   
   # Install dependencies per https://docs.ollama.com/docker - wrapped in error handling
-  if ! curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey 2>/dev/null | sudo gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg 2>/dev/null; then
+  if ! curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey 2>/dev/null | sudo gpg --batch --yes --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg 2>/dev/null; then
     echo -e "${YELLOW}#${RESET} Warning: Failed to add NVIDIA container toolkit GPG key. Continuing anyway...\\n"
     return 0
   fi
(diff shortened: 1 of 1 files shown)


chriscrosstalk (dismissed):
Thanks @Gujiassh — this is exactly the right fix and matches what we'd diagnosed in #807.

- `--batch --yes` on the gpg dearmor step correctly prevents the silent failure when there's no controlling TTY (the NVIDIA Container Toolkit was silently not installing).
- The `--retry`/`--retry-delay` on the helper-script downloads addresses the transient-failure partial installs.

Scoped tightly to those two spots, and `main` is the right base since that's the branch users fetch the install script from. LGTM — resolves #807.

jakeaturner (dismissed):

PR #862, opened 2026-05-09
fix(kiwix): lazy-load native libzim bindings

juyua9:
## Summary
- keep `@openzim/libzim` as a type-only import during module loading
- dynamically import `Archive` only inside the ZIM-specific code paths that open an archive
- apply the same lazy-loading behavior to library rebuild metadata reads and ZIM extraction

## Why
`@openzim/libzim` is a native binding. If the binding has not been built for the local platform yet, importing the package fails immediately. The Kiwix services only need that native binding when they actually open a ZIM file, so app boot/build/test paths that merely import the services should not fail before any ZIM operation runs.

This keeps the existing behavior for valid ZIM processing while avoiding an eager native-addon requirement for unrelated app paths.

## Tests
- `npm ci --ignore-scripts --no-audit --no-fund`
- `npm run typecheck`
- `npm run build`

### DIFF
--- a/admin/app/services/kiwix_library_service.ts
+++ b/admin/app/services/kiwix_library_service.ts
@@ -1,13 +1,18 @@
 import { XMLBuilder, XMLParser } from 'fast-xml-parser'
 import { readFile, writeFile, rename, readdir } from 'fs/promises'
 import { join } from 'path'
-import { Archive } from '@openzim/libzim'
+import type { Archive as ZimArchive } from '@openzim/libzim'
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
Thanks for the PR, @juyua9. Lazy-loading the native `@openzim/libzim` binding is a reasonable pattern, but in NOMAD's supported environment (the Linux Docker deployment) the binding is always present, so this is effectively a no-op in production — it doesn't change behavior for users.

The scenario it helps is running on a platform where the native binding isn't built, which today is outside our supported targets. If/when we add support for other architectures (e.g. ARM64), native-binding portability is something we'll want to handle deliberately across the whole stack rather than piecemeal, so we'd rather address it as part of that work.

Going to close this for now. Appreciate you taking the time, and thanks for thinking about portability.

Issue #710, opened 2026-04-12
[Bug]: ERROR: Failed to load the XML library file '/data/kiwix-library.xml'


boivinx7:
### Issue Category

Installation/Setup

### Bug Description

so i have this in a loop without Kiwix ever being able to boot

### Steps to Reproduce

This appens right after installing Kiwix from the apps menu

the file just as the default

`<?xml version="1.0" encoding="UTF-8"?>
<library version="20110515">
[...]

PR #863, opened 2026-05-09
fix: self-heal Kiwix library before start

euyua9:
## Summary

- add a Kiwix library XML readability guard that rebuilds `kiwix-library.xml` from the current ZIM directory when the file is missing, empty, or invalid XML
- run the guard before starting or restarting the Kiwix container, after the existing legacy glob-mode migration check
- add focused unit coverage for missing, invalid, and already-readable library XML states

## Why

Kiwix library mode depends on `/data/kiwix-library.xml`. If that file is removed, truncated, or left malformed, a normal start/restart can keep `kiwix-serve` in a crash loop even when the ZIM files are still present. This makes the library-mode path self-heal before handing control to Docker.

Related historical failure mode: #710.

## Validation

- `npm run typecheck`
- `git diff --check`

I also attempted the focused unit test command locally:

- `APP_KEY=some_random_key HOST=localhost URL=http://localhost:8080 LOG_LEVEL=info DB_HOST=localhost DB_PORT=3306 DB_USER=root DB_DATABASE=nomad REDIS_HOST=localhost REDIS_PORT=6379 npm test -- --files tests/unit/kiwix_library_service.spec.ts`

That local run is blocked in this macOS environment because `@openzim/libzim` cannot rebuild: `xcrun` is loading an incompatible-architecture Command Line Tools `libxcrun.dylib`. The TypeScript compile still passes.

### DIFF
--- a/admin/app/services/docker_service.ts
+++ b/admin/app/services/docker_service.ts
@@ -77,6 +77,7 @@ export class DockerService {
             this.invalidateServicesStatusCache()
             return { success: true, message: 'Kiwix migrated to library mode successfully.' }
           }
+          await new KiwixLibraryService().ensureReadableLibraryFile()
         }
 
         await dockerContainer.restart()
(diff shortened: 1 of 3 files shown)


euyua9:
Closing this myself after a duplicate/scope re-check. The same Kiwix library XML self-heal lane is already covered by #733, which is approved and waiting on upstream merge/checks. I opened this before catching that tracker entry, so keeping it open would only split reviewer attention.

PR #872, opened 2026-05-13
fix(RAG): unbreak multi-batch ZIM ingestion (jobId dedupe)

chriscrosstalk:
## Summary

`EmbedFileJob.dispatch()` uses a deterministic per-file jobId (`sha256(filePath).slice(0,16)`) for every batch. The parent batch's `handle()` calls `EmbedFileJob.dispatch({ batchOffset })` *before returning*, so the parent is still in `active` state and locked when the continuation tries to enqueue. BullMQ silently returns the locked parent instead of creating a new job — and in newer BullMQ versions it does so without throwing, so the existing `catch (error.message.includes('job already exists'))` branch never fires. After the parent completes, its entry stays in the `completed` ZSET (held by `removeOnComplete: { count: 50 }`), continuing to trip jobId dedupe for any subsequent re-dispatch attempts.

**Result:** every NOMAD install since 2026-02-08 (`feat: zim content embedding`) with a multi-batch ZIM (wikipedia, cooking SE, ifixit, lrnselfreliance, etc.) has only the first 50 articles indexed in qdrant. The RAG feature has been silently degraded for ~3 months — the user sees the file appear in their KB, qdrant accumulates ~50 articles' worth of vectors, and pagination quietly halts. No error surfaces anywhere.

## Fix

`dispatch()` skips the deterministic jobId for continuation batches (`batchOffset > 0`), letting BullMQ auto-generate a unique one so each batch stacks as an independent queue entry. Initial dispatches keep the deterministic jobId so re-triggering an install (UI re-click, `/api/rag/sync` rescan) remains idempotent. The existing "job already exists" branch is now gated on `!isContinuation`, since by construction continuation batches will never hit dedupe.

## Test plan

Validated on NOMAD8 (RX 6800 / Threadripper 3960X, rc.3 + this patch):

- [x] Hot-patched admin container with the rebuilt `embed_file_job.js`, restarted.
- [x] Installed `devdocs_en_python_2026-02.zim`. Without patch: capped at 372 chunks (batch 1 only). With patch: paginates through batches at offsets 50, 100, 150, 200, 250+ — qdrant points climbed past 2,700 and still going.
- [x] admin.log shows the expected sequence per batch: `Dispatched embedding job for file: X (continuation @ offset N)` followed by `Starting embedding process for: X (batch offset: N)`.
- [x] Initial-dispatch idempotency preserved: re-clicking install for the same file still hits the `job already exists` dedupe branch (verified in code, no behavior change for `batchOffset=0`).

## Why this is a v1.32.0 GA blocker

This is the 4th blocker tracked in `v1.32.0-ga-readiness.md`. Unlike the three rc.3 fixes (which are cosmetic + AMD-specific + edge-case), this one silently degrades the headline RAG feature for *every* user with a multi-batch ZIM, with no surfaced error.

### DIFF
--- a/admin/app/jobs/embed_file_job.ts
+++ b/admin/app/jobs/embed_file_job.ts
@@ -207,36 +207,58 @@ export class EmbedFileJob {
   static async dispatch(params: EmbedFileJobParams) {
     const queueService = new QueueService()
     const queue = queueService.getQueue(this.queue)
-    const jobId = this.getJobId(params.filePath)
+
+    // Continuation batches (batchOffset > 0) must NOT reuse the deterministic
+    // per-file jobId. Two BullMQ dedupe paths would otherwise silently swallow them:
+    //   1) The parent batch's handle() calls dispatch() before returning, so the
+    //      parent job is still `active` and locked — queue.add() with the same
+    //      jobId returns the locked parent rather than enqueueing the new batch.
+    //   2) After the parent completes, its entry stays in `completed` (held by
+    //      `removeOnComplete: { count: 50 }`), still tripping jobId dedupe.
+    // Letting BullMQ auto-generate a unique jobId for continuation batches stacks
+    // them as independent queue entries that each process via handle().
+    // Initial dispatches keep the deterministic jobId so re-triggering an install
+    // (UI re-click, sync rescan, etc.) is still idempotent.
+    const isContinuation = !!(params.batchOffset && params.batchOffset > 0)
+    const initialJobId = this.getJobId(params.filePath)
+
+    const jobOptions: Parameters<typeof queue.add>[2] = {
+      attempts: 30,
+      backoff: {
+        type: 'fixed',
+        delay: 60000, // Check every 60 seconds for service readiness
+      },
+      removeOnComplete: { count: 50 }, // Keep last 50 completed jobs for history
+      removeOnFail: { count: 20 }, // Keep last 20 failed jobs for debugging
+    }
+    if (!isContinuation) {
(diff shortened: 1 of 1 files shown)


chriscrosstalk:
## Heads-up: user-visible impact worth noting in release notes

Two things this fix surfaces that maintainers + users should know:

**1. Existing installs aren't auto-healed.** Any qdrant collection populated under the buggy code only contains the first 50 articles per multi-batch ZIM. After this fix merges, those users won't get the missing content automatically — they'll need to uninstall + reinstall the affected ZIMs. Worth a release-notes line, or potentially a one-time KV-flag-driven migration prompt.

**2. Pagination actually completing is a real workload increase, and on AMD ROCm it can take a box down.** Confirmed today: `nomic-embed-text:v1.5` runs at `100% GPU` on NVIDIA (RTX 5060 / `ollama ps`) but reports `size_vram: 0` on AMD with ROCm — Ollama's ROCm build doesn't currently accelerate the nomic-bert architecture used for embeddings. On NVIDIA boxes, embedding streams through fast and this fix is a pure win. On AMD boxes (and any CPU-only Ollama install), the first ingestion of a large ZIM after this fix will pin every CPU thread for hours instead of minutes. On a Threadripper 3960X with full 48-thread saturation, that was enough to starve sshd at the kernel-scheduler level until power-cycle.

This isn't a reason to delay the fix — the alternative is "RAG continues to silently degrade for everyone" which is worse — but the AMD-nomic-on-CPU behavior is worth its own upstream-Ollama investigation. Will file separately.

PR #873, opened 2026-05-13
fix(RAG): pace continuation batches when embedding is CPU-only

chriscrosstalk:
## Summary

Stacks on top of #872 (multi-batch ZIM ingestion fix).

After #872, multi-batch ZIM ingestion completes correctly — but on installs where Ollama runs the embedding model on CPU (currently every AMD ROCm install, since Ollama's ROCm build doesn't accelerate nomic-bert), the now-correct sustained 100% CPU saturation across all cores can starve other services hard enough to take the box down. Confirmed today on a Threadripper 3960X + RX 6800 NOMAD: a wikipedia-class ZIM ingestion pegged 48 threads cleanly enough that sshd lost banner-exchange responsiveness and the box ultimately required a power-cycle.

NVIDIA installs aren't affected — `nomic-embed-text:v1.5` runs at `100% GPU` on RTX 5060 (verified via `ollama ps`).

## Fix

Detect placement at runtime, pace only when needed:

1. **`OllamaService.isEmbeddingGpuAccelerated()`** — queries `/api/ps` and returns `true` if any loaded embedding model reports `size_vram > 0`. Fails closed (returns `false`) if `/api/ps` is unreachable or no embed model is loaded yet — over-pacing is safer than crashing.

2. **`EmbedFileJob.handle()`** — between batches (`hasMoreBatches: true` branch), check placement and `await new Promise(resolve => setTimeout(resolve, CPU_BATCH_DELAY_MS))` when CPU-only. `CPU_BATCH_DELAY_MS = 1000` (1s) — enough to give the OS scheduler a window for sshd/disk-collector/etc., small enough that total ingestion time isn't meaningfully affected (each batch is ~60-90s of work).

GPU-accelerated installs see zero behavior change.

## Test plan

- [ ] On a CPU-only install (or with Ollama configured to keep nomic off GPU): trigger a multi-batch ZIM ingestion. admin.log should show `Embedding is CPU-only — pacing 1000ms before dispatching next batch` between each batch. sshd remains responsive throughout.
- [ ] On a GPU install (NVIDIA): trigger the same ingestion. admin.log should NOT show the pacing log line. Total time unchanged from pre-throttle baseline.
- [ ] `/api/ps` unreachable: pacing happens (fail-closed verified by log line + total wall time increase per batch).

#872's fix is correct and necessary, but in isolation it turns a silent RAG degradation into a loud box-killer on AMD installs. This PR prevents that. Together they make multi-batch ZIM ingestion work correctly AND not require a power-cycle to recover.

Depends on #872.

### DIFF
--- a/admin/app/jobs/embed_file_job.ts
+++ b/admin/app/jobs/embed_file_job.ts
@@ -27,6 +27,12 @@ export class EmbedFileJob {
     return 'embed-file'
   }
 
+  // Delay between continuation batches when embedding runs CPU-only. Gives the OS
+  // scheduler a brief idle window so sshd / disk-collector / other services don't
+  // starve during long multi-batch ZIM ingestions. Skipped entirely when the
+  // embedding model is GPU-offloaded — see OllamaService.isEmbeddingGpuAccelerated().
+  static readonly CPU_BATCH_DELAY_MS = 1000
+
   static getJobId(filePath: string): string {
     return createHash('sha256').update(filePath).digest('hex').slice(0, 16)
   }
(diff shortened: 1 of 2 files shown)


jakeaturner (approved):
LGTM! JT 5/13/2026

PR #874, opened 2026-05-13
fix(map): Show marker notes in popups

### LINKED ISSUE #796: Marker popup on map does not render notes field

Gujiassh:
## Summary

Fixes #796.

This updates map marker normalization so the API `notes` field is preserved on client-side marker objects, then renders those notes in the selected marker popup below the marker name. The normalization logic now lives in a small pure helper module so the notes behavior has focused regression coverage without importing the React hook or API client.

## Verification

- `npm run typecheck`
- `APP_KEY=12345678901234567890123456789012 HOST=localhost URL=http://localhost LOG_LEVEL=info DB_HOST=127.0.0.1 DB_PORT=3306 DB_USER=root DB_DATABASE=project_nomad_test REDIS_HOST=127.0.0.1 REDIS_PORT=6379 node ace test unit --files tests/unit/map_markers_normalize.spec.ts`
  - The two `normalizeMapMarker` assertions passed.
  - The command then failed during application shutdown because local Redis was unavailable (`ECONNREFUSED 127.0.0.1:6379`). I attempted to start Redis via Docker, but the Redis image pull timed out against Docker Hub in this environment.

### DIFF
--- a/admin/inertia/components/maps/MapComponent.tsx
+++ b/admin/inertia/components/maps/MapComponent.tsx
@@ -166,7 +166,14 @@ export default function MapComponent() {
             onClose={() => setSelectedMarkerId(null)}
             closeOnClick={false}
           >
-            <div className="text-sm font-medium">{selectedMarker.name}</div>
+            <div className="space-y-1">
+              <div className="text-sm font-medium">{selectedMarker.name}</div>
+              {selectedMarker.notes && (
+                <div className="max-w-64 whitespace-pre-wrap text-xs text-gray-700">
+                  {selectedMarker.notes}
+                </div>
+              )}
+            </div>
           </Popup>
         )}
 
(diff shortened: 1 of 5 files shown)


chriscrosstalk:
Thanks @Gujiassh, appreciate the contribution. Same fix already shipped via PR #830 in v1.32.0-rc.2 (also closes #796), so this duplicates work that's already merged.

Heads up for future PRs: base feature branches on `dev` rather than `main`. That's where new work lands.

Closing as duplicate.

PR #877, opened 2026-05-13
fix(queue): singleton QueueService to stop ioredis connection leak

chriscrosstalk:
## Summary

Every static call site in jobs and services instantiates a fresh `QueueService` (24 call sites across 8 files). `QueueService.getQueue()` opens a BullMQ `Queue` when not cached, and each `Queue` opens two ioredis connections (one command, one blocking). Because every static call constructed a *new* `QueueService`, its internal cache was never shared — every call opened a fresh pair and none were ever closed.

In normal operation this leaked a few connections per API hit. **After PR #872 unblocked multi-batch ZIM ingestion**, every batch's `EmbedFileJob.dispatch()` call (every ~50 articles, fired every few seconds during a wikipedia-class ingestion) leaked two more. On NOMAD3 at ~1 batch/4s sustained, that's ~1,800 leaked connections/hour. Redis hit its 10,000-maxclient ceiling in ~5 hours and `nomad_admin` fell into an EPIPE flood loop. Restarting `nomad_redis` alone didn't help — admin opened 10,000+ new connections within seconds of Redis coming back. Only stopping admin actually drained the leak (10,000 → 1 connection).

## Fix

Collapse `QueueService` to a true process-wide singleton:

- Private constructor + static `getInstance()` accessor
- The existing per-queue `Map` is now shared across every dispatch / status / cleanup call, so each queue's underlying connections are opened exactly once per process
- `close()` now also clears the map so the singleton can be torn down cleanly if a graceful-shutdown hook is ever wired up

All 24 `new QueueService()` callers updated to `QueueService.getInstance()`. The private constructor enforces this at compile time — any new caller that tries `new QueueService()` fails the build.

## Test plan

Validated on NOMAD3 (RTX 5060, v1.32.0-rc.4 + this patch hot-applied):

- [x] Restarted admin with patched JS, sustained multi-batch `wikipedia_en_simple_all_nopic` ingestion through the BullMQ stalled-job auto-resume.
- [x] `connected_clients` held at **21–22 across 5 minutes** of continuous batch dispatch. Pre-fix the same scenario climbed to 10,000+ over hours.
- [x] Typecheck passes (`tsc --noEmit`).
- [x] Idle baseline: admin restart from cold = 21 clients, holds flat with no traffic.
- [ ] (Reviewer welcome to confirm) Triggering a model download or pmtiles extract should not cause connection growth either — same singleton path.

PR #872 made multi-batch ingestion correct, but in this codebase that correctness exposes a latent bug that takes the admin container down over the course of a single big ingestion run. Together with #872, this PR makes multi-batch ZIM ingestion actually usable end-to-end without operator intervention.

Stacks on `rc` after #872 (already merged into rc.4).

### DIFF
--- a/admin/app/jobs/check_service_updates_job.ts
+++ b/admin/app/jobs/check_service_updates_job.ts
@@ -95,7 +95,7 @@ export class CheckServiceUpdatesJob {
   }
 
   static async scheduleNightly() {
-    const queueService = new QueueService()
+    const queueService = QueueService.getInstance()
     const queue = queueService.getQueue(this.queue)
 
     await queue.upsertJobScheduler(
@@ -114,7 +114,7 @@ export class CheckServiceUpdatesJob {
   }
 
   static async dispatch() {
-    const queueService = new QueueService()
+    const queueService = QueueService.getInstance()
     const queue = queueService.getQueue(this.queue)
 
     const job = await queue.add(
--- a/admin/app/jobs/check_update_job.ts
+++ b/admin/app/jobs/check_update_job.ts
@@ -42,7 +42,7 @@ export class CheckUpdateJob {
   }
 
   static async scheduleNightly() {
-    const queueService = new QueueService()
+    const queueService = QueueService.getInstance()
     const queue = queueService.getQueue(this.queue)
 
     await queue.upsertJobScheduler(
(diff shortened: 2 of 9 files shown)


jakeaturner (approved):
Bravo! JT 5/13/2026

PR #879, opened 2026-05-13
fix(System): correct AMD VRAM in Graphics card + harden log probe

chriscrosstalk:
Two coordinated fixes to make the System Information page reliably show real GPU info instead of misleading `lspci` BAR0 readings or N/A.

## 1. Generalize bogus-VRAM detection to AMD

Same root cause as #835 (NVIDIA showing 32 MB), this time for AMD: `lspci` parses the first PCI memory Region (BAR0, typically 1-16 MiB on Navi cards) as `vram`. On NOMAD8 (Threadripper 3960X + Radeon RX 6800), the System Information page showed *"1 MB"* instead of *"16 GB"*. PR #850 fixed this for NVIDIA by clearing the bogus value and re-running the Ollama log probe; the check was vendor-gated to NVIDIA only.

`isBogusNvidiaVram` becomes `isBogusDgpuVram` with an `isDiscreteGpuVendor` helper matching `/nvidia|advanced micro devices|amd|ati/i`. Same 256-MiB threshold — no real discrete GPU has less than that, while Intel iGPUs (which legitimately report small shared-memory VRAM via lspci) are left untouched.

## 2. Read Ollama logs from the startup window, not tail:N

`getOllamaInferenceComputeFromLogs()` was reading the last 500 log lines and grepping for the *"inference compute"* line. That line is written once during Ollama's GPU discovery phase within seconds of startup. Under active embedding workloads we measured >1,000 log lines/min, which pushes the line past any reasonable tail within minutes — at which point the probe returns null and the UI flips to *"GPU Not Accessible"* even though Ollama is happily using the GPU (`size_vram > 0` in `/api/ps`).

Switch from `tail: 500` to `since: containerStartedAt, until: containerStartedAt + 300s`. The 5-minute window is bounded regardless of container uptime and always captures Ollama's GPU discovery output. The inference-compute line is emitted in the first few seconds, so 5 min is generous headroom.

## Test plan

Validated on NOMAD8 (RX 6800, container uptime ~10 min with sustained ingestion that generated 6,345 log lines):

| State | controllers[0] | gpuHealth | UI |
|---|---|---|---|
| Before (no fix) | `{ model: "Navi 21 ...", vram: 1 }` | `passthrough_failed` (probe didn't run on AMD) | "1 MB" |
| With #1 only (probe stale, tail:500 lost the line) | `{ model: "Navi 21 ...", vram: null }` | `passthrough_failed` | "N/A" + banner |
| With #1 + #2 | `{ model: "AMD Radeon RX 6800", vram: 16384 }` | `status: ok`, `ollamaGpuAccessible: true`, `hasRocmRuntime: true` | "16 GB", no banner |

Both branches exercise correctly: NVIDIA path unchanged (same code, just renamed identifiers), AMD path now triggers the probe and the probe reliably finds the GPU info regardless of container age.

The "1 MB" / "N/A" on a working AMD GPU is the same class of user-confidence bug as #835 was for NVIDIA. With AMD ROCm acceleration enabled by default since PR #804, more users will land in this exact state. Catching it in the same release as the AMD enablement keeps the System Information page honest about hardware that actually works.

## Notes for reviewer

- `dockerode`'s `container.logs({ since, until })` accepts UNIX timestamps in seconds. `inspect.State.StartedAt` is ISO-8601, so we convert once.
- Log volume of 1,000+ lines/min was measured during real eval ingestion, not contrived. A lighter-use install would have the line stay in tail:500 for hours, which is why this hasn't been visible before — but the failure mode is real and triggered by the headline RAG feature working well.

### DIFF
--- a/admin/app/services/system_service.ts
+++ b/admin/app/services/system_service.ts
@@ -103,17 +103,32 @@ export class SystemService {
       // line past any reasonable tail in minutes. Pinning to the startup window
       // is bounded (~5 min of logs regardless of container uptime) and never
       // ages out.
+      //
+      // Fall back to the previous tail:500 strategy if StartedAt is missing or
+      // unparseable — we can't construct a since/until window without it, but
+      // tail:500 is still useful when the container just started and the line
+      // is still recent.
       const inspect = await container.inspect()
-      const startedAtMs = new Date(inspect.State.StartedAt).getTime()
-      const startedAtSec = Math.floor(startedAtMs / 1000)
-      const startupWindowSec = startedAtSec + 300 // 5-minute window
-      const buf = (await container.logs({
+      const startedAtRaw = inspect?.State?.StartedAt
+      const startedAtMs = startedAtRaw ? new Date(startedAtRaw).getTime() : NaN
+      const hasValidStartedAt = Number.isFinite(startedAtMs) && startedAtMs > 0
+
+      const logsOpts: { stdout: true; stderr: true; follow: false; since?: number; until?: number; tail?: number } = {
         stdout: true,
         stderr: true,
-        since: startedAtSec,
-        until: startupWindowSec,
         follow: false,
-      })) as unknown as Buffer
+      }
+      if (hasValidStartedAt) {
+        const startedAtSec = Math.floor(startedAtMs / 1000)
+        logsOpts.since = startedAtSec
+        logsOpts.until = startedAtSec + 300 // 5-minute window
+      } else {
+        logger.warn(
+          `[SystemService] nomad_ollama State.StartedAt missing or invalid (${startedAtRaw ?? 'undefined'}); falling back to tail:500 for inference-compute probe`
+        )
+        logsOpts.tail = 500
+      }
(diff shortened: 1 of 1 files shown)


jakeaturner (changes requested):
@chriscrosstalk looks good but one small request: can we add a quick validation of `inspect.State.StartedAt` for missing/malformed/falsey values to prevent unexpected behavior if we pass `NaN` to `container.logs()`. I'm not sure how well `container.logs()` will handle such a value.

If invalid, we can just fallback to the previous detection strategy or silently continue through.

chriscrosstalk:
Good catch — done in `e3ed00d`. `inspect?.State?.StartedAt` now goes through a `Number.isFinite(startedAtMs) && startedAtMs > 0` gate before being used; if it fails the gate, we fall back to `tail: 500` (plus a `logger.warn`) so the probe still has a chance to find the line. Happy path verified on NOMAD8 — re-deployed + restarted, still shows `status: ok` and `vram: 16384`.

PR #880, opened 2026-05-13
fix(RAG): report ZIM ingestion progress in overall-file frame

chriscrosstalk:
Before this change, the Active Downloads / Processing Queue UI showed the ingestion progress gauge **jumping wildly** during multi-batch ZIM ingestion (5% → 88% → 27% → 5% → 56% → 36% over ~60 seconds for cooking SE).

## Root cause

Each continuation batch is a separate BullMQ job (#872), and `EmbedFileJob.handle()` reported `job.progress` in two different reference frames:

- **During-batch** (via onProgress callback): 5-95% scaled across "% through this batch's chunks"
- **End-of-batch**: overwritten to `(nextOffset / totalArticles) * 100` — "% through whole file"
- **Next continuation**: explicitly set back to 5%, climbs through per-batch range again

With GPU-accelerated ingestion completing a batch every ~4 seconds, the UI saw the jobId rotate constantly and the gauge whipsaw between the two frames.

`totalArticles` was already wired through `EmbedFileJobParams` and used at end-of-batch — but `RagService` never actually populated it, so any frame-scaling that depended on it silently fell back to the per-batch range.

## Fix

Three coordinated changes:

1. `ZIMExtractionService.extractZIMContent()` returns `{ chunks, totalArticles }` instead of raw chunks (single caller updated).
2. `RagService.processZimFile()` includes `totalArticles` in its result so the existing `EmbedFileJob.dispatch({ totalArticles: ... })` propagation actually has a value.
3. `EmbedFileJob`'s onProgress callback scales the service-reported per-batch percent into the overall-file frame when `totalArticles` is known:

   ```
   overallPercent = ((batchOffset + (percent/100) * ZIM_BATCH_SIZE) / totalArticles) * 100
   ```

   Capped at 99% to leave room for the explicit 100% at completion. Falls back to the original 5-95% range for single-batch files (PDFs/txts) where `totalArticles` is undefined.

## Test plan

Validated on NOMAD8 (RX 6800):

- **devdocs python** (~1,500 articles): monotonic across continuation jobIds: `1501@30% → 1510@33% → 1514@43% → 1518@52%`. No more reset to 5%.
- **ifixit** (~100k articles): stays near 3% for the first many batches at offset 0..3000 — correct, the file is enormous.
- **wikipedia_en_medicine** (~70k articles): stays near 0-1% for the first batches — also correct.
- Brief 0-5% blip on continuation handoff (the explicit `safeUpdateProgress(job, 5)` at batch start, before the first onProgress fires) — visible but quickly resolves to the overall-frame value. No more 5% ↔ 88% chaos.

PR #872 made multi-batch ingestion work correctly. With it working, the UI gauge is now a visible cue users rely on to understand whether their large ZIM is making progress. The pre-fix jumping pattern looks like the ingestion is broken when it isn't.

## Notes for reviewer

- `ZIMExtractionService.extractZIMContent()` return-type change is breaking for any future caller. Only `RagService.processZimFile()` calls it today; that single caller is updated.
- `ZIM_BATCH_SIZE` is imported from the existing `constants/zim_extraction.ts`.
- The interface field `totalArticles?: number` on `ProcessAndEmbedFileResponse` already existed — this PR just actually populates it.

### DIFF
--- a/admin/app/jobs/embed_file_job.ts
+++ b/admin/app/jobs/embed_file_job.ts
@@ -7,6 +7,7 @@ import { OllamaService } from '#services/ollama_service'
 import { createHash } from 'crypto'
 import logger from '@adonisjs/core/services/logger'
 import fs from 'node:fs/promises'
+import { ZIM_BATCH_SIZE } from '../../constants/zim_extraction.js'
 
 export interface EmbedFileJobParams {
   filePath: string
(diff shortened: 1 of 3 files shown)


jakeaturner (approved):
LGTM! JT 5/13/2026

Issue #883, opened 2026-05-14
RFC: Knowledge Base Ingestion UX Redesign


chriscrosstalk:
# RFC: Knowledge Base Ingestion UX Redesign

**Status:** Draft for review
**Author:** Chris
**Date:** 2026-05-13
**Target milestone:** post v1.32.0 GA, likely v1.40.0
**Companion issues:** #881, #882, plus 3 unfiled bugs documented at end

## Problem

The Knowledge Base ingestion flow today gives the user no levers and almost no information:

- **No upfront choice** about whether new content should be AI-indexed. Every ZIM, every uploaded doc, gets pushed into the embed pipeline regardless of whether the user wants it there.
- **No disk-cost preview.** A 10 GB ZIM index can occupy 1 GB or 100 GB of qdrant depending on how text-dense its contents are. Users budgeting a 250 GB NUC have no way to know which way that lands.
- **No time-cost preview.** A text-dense ZIM on a CPU-only box takes hours to days per GB. Users plan around "the download finished, NOMAD is ready" — when in reality embedding has only just begun.
- **Status is opaque.** The Processing Queue shows a percent that can flip 0→5→13→0 between batches (#880 partial). "Processing" vs "Waiting" status is read from internal job state, not BullMQ truth — leading to "all jobs processing" displays when half are actually queued. Stale UI state survives across page loads. There's no liveness signal.
- **Failures are silent.** A ZIM that produced 266 chunks of an expected 600,000 (stalled mid-extraction) is indistinguishable from one that fully completed. A video ZIM that produced 0 chunks reports "success" with no signal that AI literally cannot answer questions about its content.
- **No remediation affordances.** Stuck, partially-embedded, or unwanted-after-the-fact ZIMs can't be cancelled, retried, or un-indexed from the UI.

This RFC describes a coherent redesign that fixes all of these in one pass.

## Goals

1. **User chooses upfront** — and changes mind anytime — whether content gets AI-indexed.
2. **Cost is visible before commitment** — both disk and time.
3. **Status is honest** — what's running, how fast, whether it's stuck.
4. **Warnings appear only when relevant** — silent when fine, prominent when broken.
5. **Recovery is one click** — cancel, retry, un-index, delete are all surfaced.
6. **Onboarding sets the table** — the Easy Setup Wizard handles first-run policy without becoming busy.

[...]

PR #886, opened 2026-05-14
fix(KB): add re-embed and reset & rebuild options to fix broken embed…

jakeaturner:
Adds two new bulk actions to the Knowledge Base modal alongside the existing "Sync Storage" button:

- Re-embed All: Per-file replace. For each file currently on disk, deletes its existing Qdrant points by source match and dispatches a fresh EmbedFileJob. Files on disk are kept; any orphan points (vectors whose source file is no longer present) are preserved.
- Reset & Rebuild: Destructive. Drops the `nomad_knowledge_base` collection, recreates it, clears the Nomad-docs discovery flag, and re-queues every file on disk. Requires typing RESET to confirm.

### Backend
- New RagService methods: `reembedAll()`, `resetAndRebuild()`, plus shared helpers `_discoverKbFiles()`, `_dispatchEmbedJobsFor()`, `_deletePointsBySource()`, and
  `_hasInflightEmbedJobs()`. `scanAndSyncStorage()` refactored to use the same helpers.
- Both actions refuse with a typed message if the file-embeddings queue has any waiting/active/delayed/paused jobs, avoiding races with in-flight work.
- New routes: `POST /api/rag/re-embed-all`, `POST /api/rag/reset-and-rebuild`.

### Frontend
- Two new toolbar buttons with confirmation modals that explain the difference between the modes, call out long runtimes / sustained CPU on non-GPU systems, and warn that destructive reset removes orphan points permanently.

### Resource pacing
No new throttling required — both actions go through the existing EmbedFileJob pipeline, which already detects CPU-only embedding via `OllamaService.isEmbeddingGpuAccelerated()` and inserts the 1s inter-batch delay.

Technically more of a feat, but directly related to the RAG fixes @chriscrosstalk has been working through

### DIFF
--- a/admin/app/controllers/rag_controller.ts
+++ b/admin/app/controllers/rag_controller.ts
@@ -98,6 +98,26 @@ export default class RagController {
     }
   }
 
+  public async reembedAll({ response }: HttpContext) {
+    try {
+      const result = await this.ragService.reembedAll()
+      return response.status(200).json(result)
+    } catch (error) {
+      logger.error({ err: error }, '[RagController] Error during re-embed all')
+      return response.status(500).json({ error: 'Error during re-embed all' })
+    }
+  }
+
+  public async resetAndRebuild({ response }: HttpContext) {
+    try {
+      const result = await this.ragService.resetAndRebuild()
+      return response.status(200).json(result)
+    } catch (error) {
+      logger.error({ err: error }, '[RagController] Error during reset and rebuild')
+      return response.status(500).json({ error: 'Error during reset and rebuild' })
+    }
+  }
+
   public async health({ response }: HttpContext) {
     const result = await this.ragService.checkQdrantHealth()
     return response.status(200).json(result)
--- a/admin/app/jobs/embed_file_job.ts
+++ b/admin/app/jobs/embed_file_job.ts
@@ -240,7 +240,7 @@ export class EmbedFileJob {
     return await queue.getJob(jobId)
   }
 
-  static async dispatch(params: EmbedFileJobParams) {
+  static async dispatch(params: EmbedFileJobParams, options?: { force?: boolean }) {
     const queueService = QueueService.getInstance()
     const queue = queueService.getQueue(this.queue)
 
@@ -255,7 +255,11 @@ export class EmbedFileJob {
     // them as independent queue entries that each process via handle().
     // Initial dispatches keep the deterministic jobId so re-triggering an install
     // (UI re-click, sync rescan, etc.) is still idempotent.
+    // `force` skips the deterministic jobId for bulk callers (reembedAll /
+    // resetAndRebuild) where historical entries in :completed would otherwise
+    // silently swallow the new dispatch.
     const isContinuation = !!(params.batchOffset && params.batchOffset > 0)
+    const force = !!options?.force
     const initialJobId = this.getJobId(params.filePath)
 
     const jobOptions: Parameters<typeof queue.add>[2] = {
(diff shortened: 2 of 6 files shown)


bgauger (changes requested):
Thanks @jakeaturner — the orphan-vs-clean-slate split is clean and the `scanAndSyncStorage` refactor pulls its weight. Tested end-to-end on `pr-886` against the real Bull / Qdrant / Ollama stack and hit one blocker.

## Blocker — both bulk actions silently no-op on second use

`EmbedFileJob.dispatch()` uses a deterministic per-file jobId for idempotency (`embed_file_job.ts:269-273`). After any successful embed, that jobId sits in Bull's `:completed` set (kept by `removeOnComplete: { count: 50 }`). When `reembedAll` / `resetAndRebuild` later re-dispatch the same path, Bull recognizes the jobId and silently returns the historical job — no enqueue, no error.

Reproduction:

```
Run 1 — Re-embed All (fresh):  Qdrant 68 → 104  ✓
Run 2 — Reset & Rebuild:       Qdrant 104 → 0   (wiped, but no rebuild)  ❌
Run 3 — Re-embed All again:    Qdrant 0 → 0     ❌
```

Both runs 2 and 3 returned 200 OK with "success" messages. Queue jumped to `wait=0` instantly because Bull short-circuited. No failed jobs, no log entries distinguishing this from real work.

**Suggested fix (pick one):**

- Drain the dedupe state inside `reembedAll`/`resetAndRebuild` before `_dispatchEmbedJobsFor`:
  ```ts
  const queue = QueueService.getInstance().getQueue(EmbedFileJob.queue)
  await queue.clean(0, 1000, 'completed')
  await queue.clean(0, 1000, 'failed')
  ```
- Or add a `force` flag to `EmbedFileJob.dispatch()` that bypasses the deterministic jobId for these two bulk callers (keeps `scanAndSyncStorage` idempotent).

Related: `_dispatchEmbedJobsFor` increments `queuedCount` whenever `queue.add()` doesn't throw, but Bull doesn't throw on a dedupe hit. Even after the root fix, checking `job.processedOn === null` on the return value would let the controller report honest queued-vs-deduped counts.

## Smaller things (not blockers)

- `reembedAll` deletes points **before** dispatching. If `_dispatchEmbedJobsFor` swallows a per-file dispatch failure, the file's old points are gone but no new job exists — silently dropped from RAG. Dispatch-then-delete (or surface per-file dispatch errors) is safer.
- No mutex between `reembedAll` and `resetAndRebuild` — both check `_hasInflightEmbedJobs` once at the start and could race. Low impact in single-admin NOMAD.
- `deleteCollection` → `_ensureCollection` leaves a brief window where chat-with-RAG hits "collection not found." Worth a note in the modal copy.
- `KVStore.setValue('rag.docsEmbedded', false)` before `discoverNomadDocs(true)` is redundant since `force` bypasses the KV check.
- Destructive modal accepts case-insensitive RESET (`.toUpperCase()`) but the copy says `RESET` (caps). Pick one.

Happy to retest once the dispatch fix is in.

chriscrosstalk (approved):
Approving — this is a solid foundation for the broader KB UX work in #883 (RFC).

Quick notes for context, not blockers:

- The four extracted helpers (`_discoverKbFiles`, `_dispatchEmbedJobsFor`, `_deletePointsBySource`, `_hasInflightEmbedJobs`) are exactly the building blocks the RFC's per-card surgical operations will reuse. Re-embed All / Reset & Rebuild as bulk operations + per-card Un-index / Cancel / Retry as surgical operations cover the full repair matrix without overlap.
- The `_hasInflightEmbedJobs()` race-guard is a clean pattern. The RFC will adopt the same shape for per-card actions (guarded on that file's specific job rather than the whole queue, so users can still un-index file A while file B is mid-embed).
- The CPU-only warnings in both modals ("expect sustained high CPU usage" / "potentially hours on CPU-only systems") line up with the RFC's surfacing direction. When we add the per-card status pill, the same condition gets a visual representation (🟡 paced when on CPU) instead of a one-time modal warning.
- One thing this PR doesn't address (out of scope, but worth noting): the underlying `scanAndSyncStorage` skip-logic still treats "any chunks in qdrant" as "fully embedded," so a partially-stalled ZIM (e.g., the simple_wiki 266-of-600k case I hit yesterday) is still invisible to sync. Users can now manually recover via Re-embed All, which is a big improvement, but the proper fix lives in the RFC's `ai_ingest_state` data model.

Tested mentally against the test corpus from yesterday's NOMAD1 ingestion run; no concerns. Ship it.

chriscrosstalk:
Crosslink: bulk recovery scope from this PR + per-card surgical recovery scope from #883 (RFC) compose to the full repair matrix:

| Scope | This PR | RFC #883 |
|---|---|---|
| Re-do every file | ✓ Re-embed All | — |
| Wipe everything + start over | ✓ Reset & Rebuild | — |
| Re-do one specific file | — | per-card Retry |
| Stop one file in flight | — | per-card Cancel |
| Drop one file from AI but keep on disk | — | per-card Un-index (reuses `_deletePointsBySource`) |
| Delete one file fully | already exists | per-card Delete (existing) |
| Refuse if queue is busy | ✓ `_hasInflightEmbedJobs()` | per-card actions reuse same pattern, scoped to the specific file's job |

Helpers extracted here become foundational primitives for the RFC. No conflicts, just composition.

PR #891, opened 2026-05-14
feat(KB): ratio registry for disk + time estimates (Phase 1B of RFC #883)

chriscrosstalk:
Foundation for the cost estimates and partial-stall detection that Phase 2 of RFC #883 will surface. No consumers yet — this PR just lays the table, seed rows, and lookup helper so the subsequent UI work has estimates available without a per-ZIM benchmark at runtime.

## What lands

- **New table `kb_ratio_registry`** (`pattern`, `chunks_per_mb`, `sample_count`, `notes`).
- **Migration seeds heuristic defaults** from RFC #883's appendix:

| Pattern | chunks/MB | Notes |
|---|---|---|
| `devdocs_` | 1100 | Dense API references |
| `wikipedia_en_simple_` | 270 | Simple English Wikipedia |
| `wikipedia_en_` | 270 | General Wikipedia variants |
| `cooking.stackexchange.com_` | 200 | Stack Exchange Q&A |
| `ifixit_` | 50 | Image-heavy repair guides |
| `lrnselfreliance_`, `ted_`, `freedom-of-religion_` | 0 | Video-only ZIMs |
| `` (empty fallback) | 100 | Catch-all |

- **`KbRatioRegistry` model** with static `lookup(filename)` and `estimateChunks(filename, sizeBytes)`.
- **Pure helper `kb_ratio_lookup.ts`** does longest-prefix-match — a specific entry (`wikipedia_en_simple_`) overrides a broader one (`wikipedia_en_`). 9 unit tests cover exact match, longest-wins, broader-fallback, empty-fallback, no-match-no-fallback, zero-ratio (videos), and the byte→MB scaling in `estimateChunkCount`.
- **`sample_count` starts at 0** for heuristic seeds. Phase 4 self-calibration will increment it as observed ZIMs update each row in place.

## Not in scope

- Self-calibration (Phase 4)
- UI consumers — Warning B (partial-embed stall detection) and the storage budget meter / time-remaining estimates land in Phase 2.

## Tested

- Type-check clean
- 9 unit tests pass via `node --test`
- Migration applied on NOMAD3 via hot-patch; 9 seed rows verified in DB

## Files

| File | Note |
|---|---|
| `database/migrations/1776100000001_create_kb_ratio_registry_table.ts` | New |
| `app/models/kb_ratio_registry.ts` | New |
| `app/utils/kb_ratio_lookup.ts` | New — pure helper, unit-tested |
| `tests/unit/kb_ratio_lookup.spec.ts` | New — 9 cases |

### DIFF
--- a/admin/app/models/kb_ratio_registry.ts
+++ b/admin/app/models/kb_ratio_registry.ts
@@ -1,0 +1,51 @@
+import { DateTime } from 'luxon'
+import { BaseModel, column, SnakeCaseNamingStrategy } from '@adonisjs/lucid/orm'
+import { findChunksPerMb, estimateChunkCount } from '../utils/kb_ratio_lookup.js'
+
+/**
+ * Self-calibrating registry of `{filename-prefix → chunks_per_mb}` ratios used
+ * for disk-footprint and time-to-embed estimates surfaced in the KB panel.
+ *
+ * Migration seeds the registry with heuristic defaults from the RFC #883
+ * appendix; Phase 4 self-calibration will update rows in place as ZIMs finish
+ * ingesting and the real ratio becomes known. Lookup is longest-prefix-match
+ * (see `kb_ratio_lookup.ts`) so a specific entry (`wikipedia_en_simple_`)
+ * overrides a broader one (`wikipedia_en_`).
+ */
+export default class KbRatioRegistry extends BaseModel {
+  static table = 'kb_ratio_registry'
+  static namingStrategy = new SnakeCaseNamingStrategy()
+
+  @column({ isPrimary: true })
+  declare id: number
+
+  @column()
+  declare pattern: string
+
+  @column()
+  declare chunks_per_mb: number
+
+  @column()
+  declare sample_count: number
+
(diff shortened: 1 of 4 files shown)


jakeaturner:
Pushed a small follow-up on a type concern during review — switched `chunks_per_mb` from `DECIMAL(10,2)` to `INTEGER UNSIGNED`. With mysql2, `DECIMAL` columns come back as strings, so the model's declared number type was a lie and `ratio === 0 `checks for the video-only ZIMs would have silently failed once Phase 2 wired up consumers. Integer aligns the runtime type with the declared one, and since all seeds are whole number and the heuristic has ~±50% real-world variance anyway, we're not losing any meaningful precision. unsigned rules out negative ratios at the schema level for free.

Tests and typecheck still pass. Heads-up for NOMAD3: the hot-patched table there still has the old `DECIMAL(10,2)` column, so need to drop the table (or roll back + re-run the migration) so it picks up the corrected schema. Dev machines that haven't run this migration yet are unaffected.

PR #888, opened 2026-05-14
feat(KB): per-file ingest state machine (Phase 1 of RFC #883)

chriscrosstalk:
First slice of RFC #883 — the data plumbing that everything else in the redesign builds on. No UI changes yet; this PR earns its keep by replacing the binary "any chunks in Qdrant ⇒ embedded" check with a real state machine that captures user decisions and terminal outcomes.

## What lands

- **New table `kb_ingest_state`** keyed on `file_path` with enum state (`pending_decision | indexed | browse_only | failed | stalled`). Independent of `installed_resources` so it covers both curated downloads and manually-uploaded KB files.
- **New KV key `rag.defaultIngestPolicy`** (`Always | Manual`). Registered now but not consumed yet — JIT prompt + wizard step land in Phase 3.
- **`EmbedFileJob.handle` writes state on terminal outcomes:**
  - Success (final batch) → `indexed` + chunks count
  - `UnrecoverableError` → `failed` + error message
  - Retryable errors are left to BullMQ's existing retry path (per RFC's "surface stalled after retry exhaustion" recommendation)
- **`scanAndSyncStorage` swaps the binary check for a state-aware decision tree** (see `decideScanAction`). Existing installs auto-backfill on first scan: files with chunks in Qdrant but no state row become `indexed`; new files start as `pending_decision`. The full decision matrix is in `app/utils/kb_ingest_decision.ts` and is unit-tested for all 9 combinations.
- **`deleteFileBySource` drops the state row last**, so removed files disappear entirely instead of leaving an orphan that the next scan would re-dispatch into nothing.

## Tested on NOMAD3 via hot-patch

After a clean rc.4 refresh + hot-patch:

1. `POST /api/rag/sync` → `Scanned 8 files, queued 0 for embedding` (correct: all existing files already had chunks)
2. State table: 5 ZIMs + 2 KB uploads with chunks → all `indexed`; `lrnselfreliance_en_all` (video-only, 0 chunks historically) → `pending_decision` and was re-dispatched (Bull deduped to existing `:completed` jobId, exactly the case bgauger's #886 fix drains)
3. `DELETE /api/rag/files` for a KB upload → state row removed, disk file removed, `{"message":"File removed from knowledge base."}`

## Not in scope here

Kept the PR small and reviewable; these follow up separately:

- **Ratio registry** — separate PR. Needed for partial-stall detection (Warning B in RFC) and cost estimates, but separable.
- **#880 follow-up initial-progress anchor** — separate tiny PR.
[...]

Issue #881, opened 2026-05-13
fix(AI): OllamaService.embed() OpenAI fallback drops truncate + num_ctx, reintroduces "input length exceeds context length" errors


chriscrosstalk:
# `OllamaService.embed()` OpenAI fallback drops `truncate` + `num_ctx`, reintroduces "input length exceeds context length" errors

## Summary

`OllamaService.embed()` calls the Ollama native `/api/embed` endpoint with `truncate: true` and `options.num_ctx: 8192` (the PR #763 fix). If that call throws — for any reason — the `catch` block falls back to `openai.embeddings.create()` **without** passing those parameters. On dense source documents, the fallback path then fails with the very same `BadRequestError: 400 the input length exceeds the context length` that PR #763 was meant to eliminate.

The fallback path's `catch {}` is also bare (no error logging), so when the native endpoint fails we have no visibility into *why* — only that we ended up in the fallback.

## Where

`admin/app/services/ollama_service.ts` — `embed()` method:

```ts
try {
  const response = await axios.post(`${this.baseUrl}/api/embed`, {
    model,
    input,
    truncate: true,
    options: { num_ctx: 8192 },
  }, { timeout: 60000 })
  // ...
} catch {
  // ⚠ bare catch — no error logging, no error parameter
  logger.info('[OllamaService] /api/embed unavailable, falling back to /v1/embeddings')
  const results = await this.openai.embeddings.create({
    model,
    input,
    encoding_format: 'float',
    // ⚠ no truncate, no num_ctx — defaults to the model's loaded context_length (2048 for nomic-embed-text:v1.5)
  })
  return { embeddings: results.data.map((e) => e.embedding) }
}
```

## Repro

Observed during a CPU-only multi-ZIM ingestion run on NOMAD1 (Phoenix APU, `nomic-embed-text:v1.5` running 100% CPU, `context_length: 2048` per `/api/ps`). Across ~50 minutes of sustained ingestion of dense technical content (medlineplus, arduino.stackexchange):

- 2 occurrences of `BadRequestError: 400 the input length exceeds the context length`
- Stack trace consistently shows `at async OllamaService.embed (file:///app/app/services/ollama_service.js:359:29)` — i.e., from inside the `catch` fallback
- 0 occurrences of `[OllamaService] /api/embed unavailable` in logs, despite the fallback path being entered. (This is a separate observability bug — see companion issue on queue worker logs.)
[...]

PR #895, opened 2026-05-14
feat(KB): conditional warnings A + B on Stored Files (RFC #883 §6)

chriscrosstalk:
Surfaces two silent failure modes that the prior binary "any-chunks-in-Qdrant ⇒ embedded" check could not distinguish from healthy ingestion. Warnings render only when their condition is met (silent by default; noisy only on real problems).

## ⚠ Stacks on #891

Base is \`feat/kb-ratio-registry\` (#891) because Warning B's "expected chunks" estimate comes from \`KbRatioRegistry.estimateChunks()\`. GitHub fast-forwards to \`rc\` once #891 merges.

## Warnings

| Warning | Trigger | Example |
|---|---|---|
| **A — Zero-chunk file** | \`fileSize > 100 MB && chunksInQdrant === 0\` | Video-only / image-only ZIMs (\`lrnselfreliance_en_all\`, TED talks) that the pipeline "successfully" embeds with no extractable text — AI Assistant literally cannot reference these |
| **B — Partial-embed stall** | \`chunks > 0 && chunks < 0.5 × expectedChunks\` | The simple_wiki "266 of 600,000 chunks" case observed during NOMAD1 ingestion — previously indistinguishable from fully-completed embeds |

## What lands

- **\`app/utils/kb_warning_decision.ts\`** — pure \`decideWarnings(inputs)\` with thresholds as exported constants. **10 unit tests** cover: healthy case, both warnings, under/at/over boundary, registry-miss suppression (Warning B silent when \`expectedChunks: null\`), video-only registry entries (\`expectedChunks: 0\` correctly skips Warning B), and combined-trigger edge case.
- **\`RagService.computeFileWarnings()\`** — single Qdrant scroll tallies chunks per source, filesystem walk fills in zero-chunk files, ratio registry provides the expectation, decision function emits warnings.
- **New endpoint \`GET /api/rag/file-warnings\`** returns \`Record<source, FileWarning[]>\`. Sources with no warnings are omitted so the frontend gets clean \`warnings[source] ?? []\` defaults.
- **KB modal**: warnings render inline under each file name as amber-tinted pills. Polled every 30s alongside the existing health check.

## Not in scope

- **Warning C — Chunks skipped due to length** — PR #890 (#881 fix) prevents the silent drop at the embed boundary, so the underlying condition shouldn't fire anymore. If we still want to surface "we truncated N chunks to fit", that needs separate \`skipped_count\` tracking in EmbedFileJob — a Phase 2 follow-up.
- **Suppressing Warning B during active mid-ingestion** — the user can cross-reference Processing Queue to know it's in-flight; suppressing warnings while a job runs would mask real stalls where the job died mid-batch. Will revisit when per-card status is wired through.
- **Reading chunk counts from \`kb_ingest_state.chunks_embedded\` (#888)** — this PR uses Qdrant scroll directly so it can land independently of #888.

## Tests

- 10 new unit tests on \`decideWarnings\`, all pass
- Type-check clean
- Hot-patch + browser smoke test deferred until #891 lands (the ratio registry needs to exist in the DB for \`estimateChunks()\` to return non-null estimates — without it, only Warning A fires, still useful but Warning B stays dormant)

## Files

| File | Change |
|---|---|
| \`app/utils/kb_warning_decision.ts\` | New — pure decision function |
| \`tests/unit/kb_warning_decision.spec.ts\` | New — 10 cases |
| \`app/services/rag_service.ts\` | +86: \`computeFileWarnings()\` method |
| \`app/controllers/rag_controller.ts\` | +5: \`getFileWarnings\` handler |
| \`start/routes.ts\` | +1 route |
| \`inertia/lib/api.ts\` | +10: client helper |
| \`inertia/components/chat/KnowledgeBaseModal.tsx\` | +40: amber-pill rendering |


jakeaturner:
Pushed a follow-up for an issue with computeFileWarnings() swallowing errors into {}.

The original try/catch returning {} meant a Qdrant outage, DB failure, or filesystem error rendered as "every file is healthy" in the modal — the exact silent-failure mode this PR is meant to expose. To keep the fix surgical I went with a sentinel response shape rather than letting the error bubble:

Service now returns FileWarningsResult = { ok: boolean, warnings: Record<source, FileWarning[]> }. Same try/catch, but the catch returns { ok: false, warnings: {} } instead of a bare {}.
Controller passes the result straight through.
Modal destructures warnings for the per-row pills (unchanged behavior) and renders a single amber inline notice under the "Stored Knowledge Base Files" header when ok === false: "File warnings unavailable — couldn't read storage state. Retrying…"
Why sentinel instead of throwing: letting the error propagate hits catchInternal, which fires a generic toast every 30s while polling — noisy and unactionable. The sentinel keeps the failure visible exactly where it matters (the panel itself), self-heals silently on transient failures, and shows a stable banner on persistent ones.

The new FileWarningsResult type lives in admin/types/rag.ts alongside FileWarning. Unit tests and type-checks still passing. Per-row rendering and the decideWarnings logic untouched.

PR #899, opened 2026-05-14
feat(KB): first-chat JIT prompt for ingest policy (RFC #883 Phase 3 task 12)

chriscrosstalk:
## Summary

First-chat onboarding banner that resolves the "did this user want their content indexed?" question at the natural moment — when they open AI Chat. RFC #883 Phase 3 task 12.

When a user opens \`/chat\` with content available but no global ingest policy yet recorded, a banner appears above the chat header:

> 🧠 **Index your 23 existing files for AI Assistant?** When indexed, AI Assistant can reference them while answering your questions.  
> [ Index existing content ] [ Maybe later ]

After either button is clicked, the banner never reappears for that user. Both buttons write the same policy KV (\`rag.defaultIngestPolicy\`) that #894 manages via the KB modal toggle.

## Why this is the right design

The banner fires exactly when the decision is actionable: the user is in front of the AI Assistant, has content to index, and hasn't made a global call yet. No need for a separate settings page visit.

- **No "dismiss without deciding" X.** That would re-show the banner forever, which is worse than asking once.
- **No third "Don't ask again" button.** Either choice (Always or Manual) is implicitly "decided"; the user can change their mind any time via the KB modal toggle.
- **No size estimate in the banner.** The size estimate lives on the tier-change modal (#897). Here, the relevant question is "yes or no?" not "how much?"

## Existing users on upgrade

Existing v1.31.x users upgrading to v1.32.0 will see this banner on first chat visit post-upgrade — they get an explicit opt-in moment for content that was already on disk. That matches the RFC intent: no behind-the-scenes auto-indexing for installs that predate the policy choice.

## Backend

- **New route** \`GET /api/rag/policy-prompt-state\` returns \`{ shouldPrompt, hasContent, totalFiles }\`
- **\`RagService.getPolicyPromptState()\`** reads \`KVStore('rag.defaultIngestPolicy')\` (null when unset) and counts \`kb_ingest_state\` rows. \`shouldPrompt\` is true only when policy is null AND scanner has seen ≥ 1 file — empty-NOMAD case correctly suppressed

## Frontend

- **New \`KbPolicyPromptBanner\` component** (~120 LOC) handles the two-button decision flow with React Query optimistic loading, success/error toasts, and cascade invalidation (\`kbPolicyPromptState\`, \`ingestPolicy\`, \`embed-jobs\`, \`storedFiles\`) on \"Index existing content\" so the user immediately sees jobs appear in Processing Queue
- **Mounted in \`components/chat/index.tsx\`** as the first child of the main content column. Renders nothing when \`shouldPrompt\` is false — zero impact on existing users who already picked a policy
- **Reads \`aiAssistantName\`** from Inertia page props so banner copy uses the user's chosen assistant name ("Mongo", "AI Assistant", etc.)

## Stacks on #894

Base is \`feat/kb-policy-toggle\` because the policy KV mechanism this PR writes through is introduced there. GitHub auto-rebases to \`rc\` once #894 merges. Both can land in rc.5.

## Test plan

- [ ] After rc.5 build:
  - [ ] Fresh-state user (no policy set, content on disk) opens \`/chat\` → banner appears with correct file count
  - [ ] Click \"Index existing content\" → policy flips to Always, sync runs, banner disappears, Processing Queue lights up
  - [ ] Click \"Maybe later\" → policy flips to Manual, banner disappears
  - [ ] Reopen \`/chat\` after either → banner does NOT reappear
  - [ ] Empty NOMAD (no content) → banner does NOT appear
  - [ ] Existing user who already used the KB modal toggle → banner does NOT appear

### DIFF
--- a/admin/app/controllers/rag_controller.ts
+++ b/admin/app/controllers/rag_controller.ts
@@ -95,6 +95,11 @@ export default class RagController {
     })
   }
 
+  public async policyPromptState({ response }: HttpContext) {
+    const result = await this.ragService.getPolicyPromptState()
+    return response.status(200).json(result)
+  }
+
   public async scanAndSync({ response }: HttpContext) {
     try {
       const syncResult = await this.ragService.scanAndSyncStorage()
--- a/admin/app/services/rag_service.ts
+++ b/admin/app/services/rag_service.ts
@@ -1112,6 +1112,32 @@ export class RagService {
   }
 
   /**
+   * Compute whether the first-chat JIT prompt should fire and surface the file
+   * count the banner uses in its copy ("Index your N existing files?"). The
+   * banner appears when the user hasn't yet picked a global ingest policy
+   * (`rag.defaultIngestPolicy` unset) and the scanner has actually seen at
+   * least one embeddable file — i.e., the prompt is actionable, not theoretical
+   * on a freshly-installed empty NOMAD.
+   *
+   * Once the user picks a policy (Always or Manual) via the banner buttons or
+   * the KB modal toggle, `shouldPrompt` flips to false for good.
+   */
+  public async getPolicyPromptState(): Promise<{
+    shouldPrompt: boolean
+    hasContent: boolean
+    totalFiles: number
+  }> {
+    const policy = await KVStore.getValue('rag.defaultIngestPolicy')
+    const countRow = await KbIngestState.query().count('* as total').first()
+    const totalFiles = Number((countRow as any)?.$extras?.total ?? 0)
+    return {
+      shouldPrompt: policy === null && totalFiles > 0,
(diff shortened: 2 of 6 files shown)


jakeaturner:
Pushed two small follow-ups found in review:
1. `maybeLaterMutation` was missing an `onError` toast that `indexNowMutation` has. If the policy save failed, the user got no feedback and the banner would reappear on next chat open — looks like a broken button. Mirrored the existing error handler.
2. Added `staleTime: Infinity` to the `useQuery`. For anyone who already picked a policy, `shouldPrompt` is effectively immutable per session, and the mutations already invalidate the key on the one transition that matters. Saves a request on every `/chat` mount.

PR #900, opened 2026-05-14
feat(KB): wizard AI policy step (RFC #883 Phase 3 task 13)

chriscrosstalk:
## Summary

Adds an inline auto-index policy choice inside the Easy Setup wizard's existing AI section (Step 3 \"Content\"). Persisted to \`KVStore['rag.defaultIngestPolicy']\` on wizard submit — same key #894's KB modal toggle reads/writes — so a user who completes the wizard never sees the first-chat JIT prompt (#899) because their decision is already on record.

## Why now

Tonight's PRs (#894 policy toggle + #899 JIT prompt) cover the two main paths users currently land on a policy decision: the KB modal toggle (for users already poking around the AI panel) and the first-chat banner (for everyone else, on opening AI Chat). The wizard is the *third* path — and it's the one new users actually go through. Asking the question there means we never have to ask it again.

## UI

**Step 3 (Content)** — below the AI Models grid (only when AI is selected):

> **Auto-index new content for AI Assistant?**  
> When you add new ZIMs, documents, or curated content, should AI Assistant index them automatically so it can search them while answering your questions?  
> [ Yes, always ] [ Ask me first ]  
> *You can change this any time from the Knowledge Base panel inside AI Chat.*

Two-button radio matches #894's KB-modal toggle pattern for visual consistency.

**Step 4 (Review)** — new \"Auto-index Setting\" summary card:

> New content will be **indexed automatically** as it arrives so AI Assistant can search it.

(or \"…wait for you to opt in from the Knowledge Base panel…\" when Manual).

## Default

\`'Always'\`. New users who keep the default get the \"just works\" experience — content downloaded by the wizard becomes searchable as soon as it finishes embedding. Users who prefer explicit opt-in flip to Manual before submitting. Either way the decision sticks.

## Skipped when AI isn't selected

If the user doesn't pick the AI capability in Step 1, the radio doesn't appear in Step 3, the summary card doesn't appear in Step 4, and \`rag.defaultIngestPolicy\` stays null. The JIT prompt (#899) then handles the decision later if/when they enable AI from Settings. No silent policy lock-in for users who opted out of AI.

## handleFinish change

A new \`api.updateSetting('rag.defaultIngestPolicy', ingestPolicy)\` call runs FIRST in the finish flow, before service installs / downloads, so any content that finishes embedding during this same wizard run sees the right policy from the start. Wrapped in its own try/catch — a transient KV write failure doesn't abort the rest of the wizard.

## Stacks on #894

Base is \`feat/kb-policy-toggle\` because the policy KV mechanism this PR writes through is introduced there. Auto-rebases to \`rc\` when #894 merges.

## Pairs with #899

Wizard users decide in the wizard; non-wizard users (existing v1.31.x upgrades, users who skipped the wizard, etc.) decide at first chat. Together they cover every entry path to v1.32.0 without double-prompting.

## Test plan

- [ ] After rc.5 build:
  - [ ] Walk wizard with AI selected → radio appears in Step 3 with \"Always\" pre-selected
  - [ ] Toggle to \"Ask me first\" → Step 4 Review shows \"wait for you to opt in\" copy
  - [ ] Complete wizard → \`GET /api/system/settings?key=rag.defaultIngestPolicy\` returns the chosen value
  - [ ] Visit /chat → JIT banner does NOT appear (policy is set)
  - [ ] Walk wizard WITHOUT AI selected → no policy UI in Step 3 or Step 4
  - [ ] Complete wizard without AI → KV still null → enable AI later from Settings → JIT banner appears on next /chat visit

### DIFF
--- a/admin/inertia/pages/easy-setup/index.tsx
+++ b/admin/inertia/pages/easy-setup/index.tsx
@@ -122,6 +122,13 @@ export default function EasySetupWizard(props: {
   const [selectedServices, setSelectedServices] = useState<string[]>([])
   const [selectedMapCollections, setSelectedMapCollections] = useState<string[]>([])
   const [selectedAiModels, setSelectedAiModels] = useState<string[]>([])
+  // Auto-index policy for the AI Assistant Knowledge Base. Defaults to
+  // 'Always' so a new user who keeps the default behavior gets the "just
+  // works" experience — downloads become searchable automatically. Persisted
+  // to KVStore['rag.defaultIngestPolicy'] on wizard submit (same key #894's
+  // KB modal toggle reads/writes) so the JIT prompt at first chat sees a
+  // decided policy and doesn't ask again.
+  const [ingestPolicy, setIngestPolicy] = useState<'Always' | 'Manual'>('Always')
   const [isProcessing, setIsProcessing] = useState(false)
   const [showAdditionalTools, setShowAdditionalTools] = useState(false)
   const [remoteOllamaEnabled, setRemoteOllamaEnabled] = useState(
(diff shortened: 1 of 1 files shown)


jakeaturner:
@chriscrosstalk a note on these changes (not blocking right now): The radio, the review card, and the persistence write each use a different predicate:
- Step 3 Radio: `isAiSelected = selectedServices.includes(OLLAMA) || installedServices.some(OLLAMA)`
- Step 4 Review card: `selectedAiModels.length > 0 || remoteOllamaEnabled`
- `handleFinish` persistence: same as step 3

We should review the logic here at some point and see if we can simplify into a single predicate and hoist it into a `useMemo` for cleanest logic+state.

PR #901, opened 2026-05-14
feat(KB): guardrail modal at 50GB / 10%-free thresholds (RFC #883 Phase 3 task 14 §7)

chriscrosstalk:
## Summary

One-time confirmation step gating bulk indexing actions that would use a substantial amount of disk for embedding storage. RFC #883 §7 (Phase 3 task 14 remainder — the inline-batch-estimate half is #897).

Fires only when **policy=Always** AND the estimate trips either:
- **50 GB** absolute embedding cost (\`GUARDRAIL_ABSOLUTE_BYTES\`), OR
- **10% of free disk** space (\`GUARDRAIL_FREE_DISK_RATIO\`)

Silent under \`policy=Manual\` — the user has already opted out of automatic ingestion, so a bulk-disk warning would be a false alarm; files would just queue as \`pending_decision\` either way.

## What the user sees

When a user picks a tier with a large estimated cost and clicks Submit, the existing tier modal stays in place and the guardrail modal layers over the top:

> ⚠️ **Confirm large AI indexing operation**
>
> Indexing this batch for the AI Assistant will use approximately **62.4 GB** of disk space for embeddings, on top of the raw downloads.
>
> That's more than 10% of your remaining free disk space (200 GB free). Embedding can take several hours and is hard to interrupt cleanly once started.
>
> If you'd rather review per-item before indexing, cancel here and switch your Auto-index setting to **Manual** from the Knowledge Base panel.
>
> [ Cancel ] [ Proceed anyway ]

Cancel → stays on tier modal so the user can change tier or flip policy.  
Proceed → runs the original \`onSelectTier\` path as if the guardrail wasn't there.

## Why this scope

This is the surface that actually matters for accidental disk consumption: the Recommended → Power tier jump can add tens of GB of ZIMs with proportional embedding cost. Single-file uploads almost never trip the thresholds; the JIT prompt's \"Index now\" button is already a conscious user click; background sync isn't a user decision. TierSelectionModal is where users commit to bulk dispatch.

## Files

- **\`inertia/lib/kb_guardrail.ts\`** — pure decision helper. \`evaluateGuardrail({estimateBytes, freeBytes})\` returns \`{trips, reasons[]}\`. No I/O on the helper itself; both threshold constants live alongside the logic so the math is trivially testable
- **\`inertia/components/KbGuardrailModal.tsx\`** — confirmation dialog. Headless UI \`Transition + Dialog\`, amber header, plain-English summary, \`z-[60]\` so it layers above the tier modal underneath
- **\`inertia/components/TierSelectionModal.tsx\`** — \`handleSubmit\` now evaluates the guardrail when \`policy=Always\` and \`embedEstimate\` is available; if it trips, we stash the verdict in component state and render the modal as an overlay
- **\`tests/unit/kb_guardrail.spec.ts\`** — 9 cases covering both thresholds, the boundary, the disabled-relative-check case, and both-at-once

## Disk-free signal

Reuses the existing \`useSystemInfo\` hook + \`getPrimaryDiskInfo\` helper (same path the easy-setup wizard's StorageProjectionBar uses). Passing \`freeBytes=0\` (unknown) cleanly skips the relative-disk check, so the modal still works on hosts whose disk introspection failed — it falls back to the absolute 50 GB threshold.

## Tests

\`\`\`
✔ small batch does not trip the guardrail
✔ batch at exactly the absolute threshold trips
✔ batch over the absolute threshold trips with over_absolute reason
✔ batch over 10% of free disk trips with over_free_disk reason
✔ batch can trip BOTH thresholds simultaneously
✔ freeBytes = 0 skips the relative-disk check
✔ freeBytes = 0 still trips the absolute check at 50 GB
✔ relative-disk threshold computed from GUARDRAIL_FREE_DISK_RATIO constant
✔ batch just under both thresholds does not trip
ℹ tests 9 / pass 9 / fail 0
\`\`\`

## Stacks on #897

Base is \`feat/kb-tier-estimate-on-disk\` because the guardrail consumes that PR's estimate endpoint to compute the verdict input. Auto-rebases to \`rc\` when #897 merges.

Pairs with #894 (policy toggle) and #899 (JIT prompt). Together the three PRs cover the \"how do I avoid surprising the user with auto-indexing they didn't ask for?\" arc.

## Out of scope (deferred)

- **6-hour time threshold (RFC §7)** — needs a per-host chunks-per-second metric we don't capture yet. Add after Phase 4 self-calibration (RFC §15) lands. The two storage thresholds carry the load for v1.32.0 GA
- **Wider integration** beyond TierSelectionModal — KB-modal manual sync isn't really a bulk decision, and the JIT prompt's \"Index now\" is already explicit consent. We can add surfaces if field usage shows them needed

## Test plan

- [x] 9/9 unit tests green
- [ ] After rc.5 build:
  - [ ] Policy=Always, pick a tier ≥ 50 GB estimated embedding cost → guardrail modal appears with absolute-threshold copy
  - [ ] Policy=Always, pick a tier whose estimate is > 10% of free disk (but < 50 GB) → guardrail appears with disk-percentage copy
  - [ ] Policy=Manual, same picks → tier-modal Submit proceeds directly with NO guardrail
  - [ ] Cancel from guardrail → tier modal still open, selection preserved
  - [ ] Proceed → both modals close, tier selection processes as before

### DIFF
--- a/admin/inertia/components/KbGuardrailModal.tsx
+++ b/admin/inertia/components/KbGuardrailModal.tsx
@@ -1,0 +1,109 @@
+import { Fragment } from 'react'
+import { Dialog, Transition } from '@headlessui/react'
+import { IconAlertTriangle, IconX } from '@tabler/icons-react'
+import { formatBytes } from '~/lib/util'
+import StyledButton from './StyledButton'
+import type { GuardrailVerdict } from '~/lib/kb_guardrail'
+
+/**
+ * One-time confirmation modal for bulk indexing actions that trip the
+ * disk-usage thresholds in `lib/kb_guardrail.ts`. The caller (e.g.
+ * TierSelectionModal) decides whether to show the modal by evaluating the
+ * guardrail BEFORE submit; this component just presents the verdict and
+ * passes the user's choice back via `onConfirm` / `onCancel`.
+ */
+interface KbGuardrailModalProps {
+  isOpen: boolean
+  verdict: GuardrailVerdict
+  onConfirm: () => void
+  onCancel: () => void
+}
+
+export default function KbGuardrailModal({
+  isOpen,
+  verdict,
+  onConfirm,
+  onCancel,
+}: KbGuardrailModalProps) {
+  // The primary number to surface — every triggered reason carries the same
+  // estimateBytes, so just grab the first one. `0` is a defensive fallback
+  // for the (impossible-by-construction) "open with empty verdict" case.
+  const estimateBytes = verdict.reasons[0]?.estimateBytes ?? 0
+  const freeReason = verdict.reasons.find((r) => r.kind === 'over_free_disk')
+
+  return (
+    <Transition appear show={isOpen} as={Fragment}>
+      <Dialog as="div" className="relative z-[60]" onClose={onCancel}>
+        <Transition.Child
+          as={Fragment}
+          enter="ease-out duration-200"
+          enterFrom="opacity-0"
+          enterTo="opacity-100"
+          leave="ease-in duration-150"
+          leaveFrom="opacity-100"
+          leaveTo="opacity-0"
+        >
+          <div className="fixed inset-0 bg-black/50" />
+        </Transition.Child>
+
+        <div className="fixed inset-0 overflow-y-auto">
+          <div className="flex min-h-full items-center justify-center p-4">
+            <Transition.Child
+              as={Fragment}
+              enter="ease-out duration-200"
+              enterFrom="opacity-0 scale-95"
+              enterTo="opacity-100 scale-100"
+              leave="ease-in duration-150"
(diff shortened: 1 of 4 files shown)


jakeaturner:
Pushed two small fixes from the review:
1. Submit disabled while estimate loads. `embedEstimate` is undefined until the query resolves, so a quick Submit on a freshly-selected Power tier was bypassing the guardrail entirely. Added (`embedEstimateRequest.length > 0 && isEstimating`) to the Submit disabled predicate.
3. Guardrail moved out of the outer `<Transition>`. It was rendered as a raw conditional sibling inside the parent `Transition`, which expects `Transition.Child` descendants. Wrapped the return in a `Fragment` and rendered `<KbGuardrailModal>` outside.

PR #909, opened 2026-05-18
feat(KB): per-file ingest action + state indicator on Stored Files (RFC #883 §5)

chriscrosstalk:
## Summary
- Resolves the Manual-mode dead-end: each Stored Files row now has a state
  pill + adaptive action button (`Index` / `Retry` / `Re-embed`) alongside
  Delete.
- New `POST /api/rag/files/embed` validates the source, refuses double-
  dispatch on the same filePath, optionally clears prior Qdrant points
  (force=true), then queues an `EmbedFileJob` via the existing dispatch
  helper.
- `GET /api/rag/files` response shape changes from `string[]` to
  `Array<{source, state, chunksEmbedded}>`. State + chunk count come from
  a single `KbIngestState` join folded into the existing Qdrant union —
  no new round trips.
- Per-file Re-embed (on already-indexed files with warnings) uses a
  confirmation modal since it deletes existing vectors before queueing.

## Behavior

| Row state | Action button | Force | Confirm |
|---|---|---|---|
| `pending_decision` | Index | no | no |
| `browse_only` | Index | yes | no |
| `failed` / `stalled` | Retry | yes | no |
| `indexed` + warning chip | Re-embed | yes | yes (modal) |
| `indexed` healthy / null | none — bulk Re-embed All covers it | — | — |

State pill: `Indexed` (green) / `Not Indexed` (neutral) / `Failed` (red) /
`Stalled` (amber); none on the admin_docs collapsed row.

## Coordination notes
- **Folds in PR #907's substantive fix.** Restores
  `<StyledTable<KbFileGroup>>` and uses `record.displayName` instead of
  the unresolved `sourceToDisplayName(record.source)` that ships in rc.5
  and ReferenceErrors on modal open. Needed because the new render reads
  multiple fields off `record`. PR #907's title tooltips on the three
  bulk-action buttons are NOT folded in — that part stays with #907 (trivial
  same-change conflict if both PRs land independently).
- **RFC #883 §5 partial.** Implements per-card Index / Retry / Re-embed;
  defers Un-index and per-file Cancel to a later PR.
- **Multi-select deferred.** Most Manual-mode users ingest 1-2 files at a
  time; the existing toggle covers the bulk case. Tracking issue for an
  "Index N pending" single-click button to be filed if user feedback warrants.

## Test plan
- [ ] CI passes (backend tsc, inertia tsc on my changed files, existing
      unit tests including the updated `kb_file_grouping.spec.ts`)
- [ ] Build integration image with this branch on NOMAD3 and Playwright UAT:
  - [ ] State pill renders per state across the existing 7-row Stored Files table
  - [ ] `Index` on a pending_decision file (toggle policy to Manual, add a fresh ZIM)
        opts in cleanly and row flips to `Indexed` after embed completes
  - [ ] `Retry` on a `failed` row re-dispatches (synthetically `markFailed`
        one row to repro)
  - [ ] `Re-embed` on `irp.fas.org_en_military-medicine` (current
        partial_stall on NOMAD3) shows confirm modal, deletes prior chunks,
        re-dispatches
  - [ ] Double-clicking the action button does NOT enqueue two jobs
        (inflight refusal works)
  - [ ] Delete flow unchanged (inline confirm row + button)
- [ ] Backend unit tests for `embedSingleFile` (unknown source rejected,
      inflight refused, force=true delete-then-dispatch) in follow-up PR

### DIFF
--- a/admin/app/controllers/rag_controller.ts
+++ b/admin/app/controllers/rag_controller.ts
@@ -7,7 +7,7 @@ import app from '@adonisjs/core/services/app'
 import { randomBytes } from 'node:crypto'
 import { sanitizeFilename } from '../utils/fs.js'
 import { basename } from 'node:path'
-import { deleteFileSchema, estimateBatchSchema, getJobStatusSchema } from '#validators/rag'
+import { deleteFileSchema, embedFileSchema, estimateBatchSchema, getJobStatusSchema } from '#validators/rag'
 import logger from '@adonisjs/core/services/logger'
 
 @inject()
@@ -80,6 +80,15 @@ export default class RagController {
       return response.status(500).json({ error: result.message })
     }
     return response.status(200).json({ message: result.message })
+  }
+
+  public async embedFile({ request, response }: HttpContext) {
+    const { source, force } = await request.validateUsing(embedFileSchema)
+    const result = await this.ragService.embedSingleFile(source, force ?? false)
+    if (!result.success) {
+      return response.status(409).json({ error: result.message })
+    }
+    return response.status(202).json({ message: result.message })
   }
 
   public async getFailedJobs({ response }: HttpContext) {
--- a/admin/app/controllers/rag_controller.ts
+++ b/admin/app/controllers/rag_controller.ts
@@ -86,7 +86,13 @@ export default class RagController {
     const { source, force } = await request.validateUsing(embedFileSchema)
     const result = await this.ragService.embedSingleFile(source, force ?? false)
     if (!result.success) {
-      return response.status(409).json({ error: result.message })
+      const status = {
+        not_found: 404,
+        inflight: 409,
+        delete_failed: 500,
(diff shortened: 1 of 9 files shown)


jakeaturner:
Added a small follow-up on the per-file ingest action. The endpoint was returning 409 for every `embedSingleFile` failure, which conflated four pretty different conditions: an untracked source path (really a 404), an already-queued job (the only genuine conflict — 409), and the two server-side failures from clearing prior Qdrant points or dispatching the job (both 500s).

I added a typed result on the service — `EmbedSingleFileResult` with `code: 'not_found' | 'inflight' | 'delete_failed' | 'dispatch_failed'` — and switched the controller to map each code to its proper status. The code is also returned in the response body so the frontend can branch on it directly instead of parsing the message string. Only one call site existed  (RagController.embedFile), so the blast radius is small.

PR #912, opened 2026-05-18
fix(security): reject dotted localhost download URLs

### LINKED ISSUE #911: Download URL validator allows dotted localhost hostnames
## Summary

The remote download URL validator currently lowercases `URL.hostname`, but does not normalize a trailing root dot before applying the loopback/link-local deny patterns.

## Reproduction

In Node.js, `new URL('http://localhost./file.zim').hostname` returns `localhost.`. The current `^localhost$` pattern does not match that value, so this spelling is not rejected even though it resolves to localhost semantics.

## Expected behavior

Loopback hostnames should be normalized before the denylist check so `localhost`, `localhost.`, and mixed-case variants are rejected consistently while LAN/private mirror hosts remain allowed.

luyua9:
## Summary

- normalize parsed download URL hostnames before SSRF loopback/link-local checks
- reject `localhost.` / mixed-case dotted localhost forms consistently with `localhost`
- add focused unit coverage for loopback/link-local rejects and LAN/public allow cases

Closes #911

## Tests

- `npm run typecheck`
- `node --import ts-node-maintained/register/esm --input-type=module` direct validator smoke test for dotted localhost and LAN/public allow cases

Note: `npm test -- --files=tests/unit/validators/common.spec.ts` currently stops during app boot in this local checkout because `@openzim/libzim` cannot locate its native `zim_binding.node` binding before the selected test file runs.

### DIFF
--- a/admin/app/validators/common.ts
+++ b/admin/app/validators/common.ts
@@ -14,7 +14,7 @@ import ipaddr from 'ipaddr.js'
  */
 export function assertNotPrivateUrl(urlString: string): void {
   const parsed = new URL(urlString)
-  const hostname = parsed.hostname.toLowerCase()
+  const hostname = parsed.hostname.toLowerCase().replace(/\.+$/, '')
 
   // `URL.hostname` strips the surrounding brackets from IPv6 literals
   // (e.g. `http://[::1]/` → hostname `::1`), so IPv6 patterns must match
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
Thanks @luyua9 — good catch on the `localhost.` trailing-dot bypass (#911). That normalization is carried forward. It overlapped with #930 (both rewrite `assertNotPrivateUrl`), and since this is SSRF-critical code we've consolidated both approaches into a single in-house validator with a consolidated test suite in #985 — your trailing-dot handling is in there alongside ipaddr.js range classification. Closing in favor of that. Appreciate the fix and the tests.

PR #916, opened 2026-05-18
feat(chat): confirm-on-switch + one-chat-model-at-a-time enforcement

chriscrosstalk:
## Summary

Surfaces NOMAD's previously-silent model-stacking behavior and enforces a "one chat model in VRAM at a time" invariant (the embedding model is always exempt). Addresses the NOMAD3 testing observation that switching the dropdown in the chat header was invisibly slow on low-VRAM hardware because the prior model was never unloaded — Ollama would either evict it under memory pressure or load the new one on CPU after the runner choked.

## Behavior

Three integration points all funnel through one new helper:

- **User changes the model dropdown** in an active chat session → confirm modal *"Switch to {newModel}? Switching to {newModel} will start a new chat. Your current conversation stays available in the sidebar."* On confirm, fire `keep_alive: 0` against the previous chat model, clear active session, set the new selection. Cancel snaps the visible dropdown back to the previous value (no popup state leaks into `selectedModel`).
- **User clicks a session in the sidebar** → no popup (system-initiated). Restore the session's stored model into the dropdown and fire `unloadChatModels(targetModel)` so anything that isn't the target gets the unload hint.
- **Chat page first mount** → page-load normalization. Anything stacked from a prior session gets the unload hint with the current selected model as the target-to-preserve. Guarded by a ref so it only fires once per page lifetime.

Backend is a single new helper and a single new route:

- `OllamaService.unloadAllChatModelsExcept(targetModel: string | null)` queries `/api/ps`, filters out (a) the embedding model name (hardcoded `nomic-embed-text:v1.5` to avoid the RagService circular import) and (b) `targetModel`, then fires `POST /api/generate` with empty prompt + `keep_alive: 0` in parallel against everything else. Returns the names that were hinted. Best-effort: network or Ollama errors are logged and swallowed so callers don't fail on housekeeping.
- `POST /api/ollama/unload-chat-models` wraps it with a validator.

## In-flight inference safety

`keep_alive: 0` is a post-completion hint, not a force-kill. If Session A is mid-response on gemma when Session B fires the unload, gemma stays resident until A's request completes, then evicts. The user-visible worst case is the race where A's longer-running request re-extends the timer back to the default and the unload is no-op'd; the next transition (or page reload) gets another chance, and Ollama's own LRU catches up under memory pressure regardless. Robust in-flight tracking (counter of active streams per model) deferred to a follow-up if we see stale-state in the wild.

## What's NOT in this PR

- Auto-unload chat models before `EmbedFileJob` ingestion (separate concern, separate PR)
- "Free up VRAM" manual button on Models & Settings
- VRAM-usage display on chat header / Models & Settings
- Formal "default model" picker (the existing `chat.lastModel` persistence already covers this case implicitly)
- Dedicated unit / Playwright tests (will follow as a separate PR per established cadence)

## Coordination notes

Base `rc` rather than `dev`: `dev` is significantly behind `rc` (none of the RFC #883 / rc.6 work has been backmerged yet), so branching off `dev` would put the new code in a context missing two weeks of fixes. v1.40.0 will inherit everything from rc.6 via the eventual backmerge, so landing this on `rc` is the cleanest path.

## Test plan

- [ ] CI passes (backend tsc clean; existing inertia tsconfig errors are pre-existing, unrelated to this PR)
- [ ] Manual UAT once integration image lands on NOMAD3:
  - [ ] Open chat with model A loaded → switch dropdown to B → confirm modal appears with correct copy ("Switch to B? Switching to B will start a new chat...") → confirm → new chat created → `/api/ps` shows A unloaded
  - [ ] Same as above but click Cancel → dropdown reverts to A → original session intact → `/api/ps` unchanged
  - [ ] Open chat → click a sidebar session whose model differs from currently-loaded → dropdown updates silently (no popup) → `/api/ps` shows old model unloaded
  - [ ] Load chat page with two chat models stacked → page-load normalize fires once → `/api/ps` shows only the current session's model
  - [ ] With Session A streaming a long response on gemma, switch Session B's dropdown → confirm → Session A continues streaming normally (Ollama defers the eviction)

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -5,7 +5,7 @@ import { RagService } from '#services/rag_service'
 import Service from '#models/service'
 import KVStore from '#models/kv_store'
 import { modelNameSchema } from '#validators/download'
-import { chatSchema, getAvailableModelsSchema } from '#validators/ollama'
+import { chatSchema, getAvailableModelsSchema, unloadChatModelsSchema } from '#validators/ollama'
 import { inject } from '@adonisjs/core'
 import type { HttpContext } from '@adonisjs/core/http'
 import { RAG_CONTEXT_LIMITS, SYSTEM_PROMPTS } from '../../constants/ollama.js'
@@ -31,6 +31,19 @@ export default class OllamaController {
       limit: reqData.limit || 15,
       force: reqData.force,
     })
+  }
+
+  /**
+   * Send Ollama `keep_alive: 0` hints to every currently-loaded chat model
+   * except the embedding model and (optionally) a target model to preserve.
+   * Used by the chat UI to enforce the "one chat model at a time" invariant
+   * on model-switch, session-switch, and page-load. Best-effort: a failure
+   * here should not block the calling flow.
+   */
+  async unloadChatModels({ request, response }: HttpContext) {
+    const { targetModel } = await request.validateUsing(unloadChatModelsSchema)
+    const unloaded = await this.ollamaService.unloadAllChatModelsExcept(targetModel ?? null)
+    return response.status(200).json({ unloaded })
   }
 
   async chat({ request, response }: HttpContext) {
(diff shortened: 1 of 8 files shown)


jakeaturner:
Pushed one small follow-up here: lifted 'nomic-embed-text:v1.5' into `EMBEDDING_MODEL_NAME` in `constants/ollama.ts` and updated both `RagService` and `OllamaService` to import from there. That retires the `EMBEDDING_MODEL_EXEMPT` duplicate and its "keep in sync with RagService.EMBEDDING_MODEL" comment. `constants/ollama.ts` doesn't import any services, so there's no circular-dep risk that motivated the duplication originally.

PR #920, opened 2026-05-20
fix(AI): improve remote Ollama url validation to prevent SSRF vulnera…

jakeaturner:
Hardens the remote-Ollama URL validation in ollama_controller.ts against SSRF, while still allowing the legitimate LAN / loopback targets users actually point NOMAD at (LM Studio, llama.cpp, sibling Ollama containers, `host.docker.internal`, etc.).

0801c12 — fix(AI): improve remote Ollama url validation to prevent SSRF vulnerability
- Replaces the trivial startsWith('http') check with a new assertNotCloudMetadataUrl() helper in `admin/app/validators/common.ts`.
- The helper enforces an http: / https: scheme (blocks file:, gopher:, etc.) and blocks the cloud instance-metadata endpoints (169.254.169.254 IPv4 and fd00:ec2::254 IPv6) to prevent IAM-credential exfiltration from a misconfigured cloud VM.
- Intentionally narrower than assertNotPrivateUrl — RFC1918 / loopback / link-local hosts remain allowed because that's where legitimate self-hosted endpoints live.

44a11bf — fix(security): match IPv6 SSRF patterns against unbracketed hostnames
- Follow-up fix to the existing `assertNotPrivateUrl` patterns. `URL.hostname` strips the brackets from IPv6 literals (http://[::1]/ → ::1), so the old regexes `(/^\[::1\]$/, /^\[::\]$/, …)` never matched and the IPv6 loopback / link-local / all-zeros / IPv4-mapped guards were silently dead.
- Updated patterns to match the unbracketed form so IPv6 SSRF attempts are actually blocked.

Test plan
- npm run typecheck and npm run lint clean
- Manually verify remote Ollama config rejects http://169.254.169.254/..., file:///etc/passwd, and gopher://...
- Verify it still accepts `http://host.docker.internal:11434` and a LAN IP
- Unit-check that `assertNotPrivateUrl('http://[::1]/')` now throws (previously passed)


hestela (approved):
i think this should be fine

jakeaturner:
@hestela  Follow-up on the Copilot review note: `assertNotCloudMetadataUrl` previously did a literal string compare against `169.254.169.254` and `fd00:ec2::254`, which missed IPv4-mapped IPv6 forms `([::ffff:169.254.169.254], [::ffff:a9fe:a9fe], [0:0:0:0:0:ffff:a9fe:a9fe])` and the fully-expanded IPv6 IMDS literal.

Now switching to using the battle-tested `ipaddr.js` library to parse the hostname, unwrap IPv4-mapped IPv6 to the embedded v4, normalize, then compare against a canonical blocklist. Added `tests/unit/cloud_metadata_url.spec.ts` covering each bypass form plus allow-list cases (loopback / RFC1918 / DNS) to confirm the remote-Ollama LAN use case is unaffected.

The `assertNotPrivateUrl` validator should also be updated to follow this pattern in the next release. Created issue #922 to track this. DNS rebinding out of scope here, also addressable in next release.

admin/package.json
```diff
@@ -100,6 +100,7 @@
     "edge.js": "6.4.0",
     "fast-xml-parser": "5.5.9",
     "fuse.js": "7.1.0",
+    "ipaddr.js": "^2.4.0",
```
bgauger: do you need to remove the ^?

bgauger (approved):
Approved

PR #770, opened 2026-04-22
fix(api): accept notes, marker_type, and position on markers endpoints

jrsphoto:
Fixes #768.

The VineJS validators in `createMarker` and `updateMarker` silently dropped fields not in their schema. `notes` and `marker_type` exist on the model and are returned by GET responses but could not be set via POST or PATCH. `updateMarker` additionally did not accept `latitude`/`longitude`, so markers could not be repositioned.

Changes:
- Add `notes` and `marker_type` to both validators and model assignments.
- Add `latitude`/`longitude` to the update validator.
- Add coordinate range validation (`-90..90`, `-180..180`) on both endpoints.

Release notes updated under Unreleased > Bug Fixes.

PR #926, opened 2026-05-20
feat(maps): add notes input to map pin placement popup

chriscrosstalk:
## What

Adds a notes textarea to the map pin placement popup so users can actually enter notes when they drop a pin.

## Why

The `map_markers` backend has accepted a `notes` column since PR #770 and the popup display path was wired up to render it (commit 6328256, shipped in v1.32.0). But no commit ever added the input on the client side. The feature shipped two-thirds done: notes get stored if you write them with a curl call, and they display when present, but there's no way for a regular user to actually enter them.

Before this PR:

| Layer | State |
|---|---|
| Backend column + API accepts notes | :white_check_mark: |
| Popup displays notes when populated | :white_check_mark: |
| **Placement UI has notes input** | :x: |

## How

Three files, 32 insertions / 5 deletions:

- **`admin/inertia/lib/api.ts`** - extend the `createMapMarker` request type with optional `notes?: string | null`
- **`admin/inertia/hooks/useMapMarkers.ts`** - `addMarker(name, lng, lat, color, notes)` accepts and forwards notes. The response handler already populated `notes` into local state from `result.notes ?? null`, so the display side already worked.
- **`admin/inertia/components/maps/MapComponent.tsx`** - `markerNotes` state, a 2-row `<textarea>` styled to match the name input, placed between the name field and the color picker. Reset on map-click (new placement), Cancel, and Save. Save trims and null-coalesces empty strings to null.

## Out of scope

Edit-mode for existing markers (so users can backfill notes on already-placed pins) is **not** in this PR. The selected-marker popup at `MapComponent.tsx:236-252` is still read-only. Adding edit mode involves real UX decisions (where the edit toggle lives, two-row layout for name+notes, wiring color editing too) that don't belong in this mechanical fix. Follow-up PR if anyone asks for it.

## Verified live on NOMAD3

Hot-patched NOMAD3 with this branch on top of v1.32.0 and ran a full round-trip via Playwright:

- Placement popup renders the new textarea (rows=2, placeholder "Notes (optional)") between name input and color picker
- Typed a name + 3-line notes, clicked Save - pin appears
- `GET /api/maps/markers` returned the saved row with multi-line notes intact (newlines preserved)
- Clicked the saved pin - read-only popup shows name + all three notes lines with whitespace preserved
- Selected-marker popup confirmed still read-only (no inputs) - confirms Option B scope held

## Test plan

- [ ] Click anywhere on the map; placement popup opens with name input focused and an empty notes textarea below it
- [ ] Type a name + notes, click Save; pin appears
- [ ] Click the new pin; popup shows name + notes
- [ ] Click another spot, leave notes empty, save; pin saves with `notes: null` (verify via `GET /api/maps/markers`)
- [ ] Esc in the notes textarea closes the placement popup (consistent with the name input)
- [ ] Enter in the notes textarea adds a newline (does **not** submit) - diverges from the name input's Enter-to-save on purpose

### DIFF
--- a/admin/inertia/components/maps/MapComponent.tsx
+++ b/admin/inertia/components/maps/MapComponent.tsx
@@ -72,6 +72,7 @@ export default function MapComponent({
   const [isDraggingMap, setIsDraggingMap] = useState(false)
   const [placingMarker, setPlacingMarker] = useState<{ lng: number; lat: number } | null>(null)
   const [markerName, setMarkerName] = useState('')
+  const [markerNotes, setMarkerNotes] = useState('')
   const [markerColor, setMarkerColor] = useState<PinColorId>('orange')
   const [selectedMarkerId, setSelectedMarkerId] = useState<number | null>(null)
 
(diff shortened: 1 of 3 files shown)


kennethbrewer3:
Is this compatible with the work done on https://github.com/Crosstalk-Solutions/project-nomad/pull/836? I've added that functionality and enhanced it with markdown.

chriscrosstalk:
Confirmed the regression on current `dev`: dropping a pin opens a placement popup with a name field and color picker only, no notes input, even though the backend (#770) and the marker view-popup both support notes. This adds the notes field back to the placement form so a note can be entered when the pin is created. CI is green and the branch is conflict-free.

jakeaturner (approved):

PR #930, opened 2026-05-21
fix(security): Parse private URLs with ipaddr.js

### LINKED ISSUE #922: fix: use ipaddr.js parsing in assertNotPrivateUrl and add unit test
The `assertNotPrivateUrl` validator in `admin/app/validators/common.ts` should be updated to use `ipaddr.js` parsing to follow the precedent set by `assertNotCloudMetadataUrl` introduced in PR #920 . Unit test should also be added to ensure conformation to stated behavior.

Gujiassh:
## Summary

Fixes #922.

- replaces the regex-only `assertNotPrivateUrl` literal-IP checks with `ipaddr.js` parsing
- preserves NOMAD's LAN-appliance behavior by allowing RFC1918 addresses and DNS names
- continues blocking loopback, link-local, unspecified, and IPv4-mapped loopback/link-local literals
- adds focused unit coverage for private URL validation edge cases

## Duplicate preflight

Before implementing, I checked for overlapping work:

- no existing PR for `Gujiassh:fix/private-url-ipaddr-20260521`
- no open PR matching `assertNotPrivateUrl ipaddr`
- #922 is still open and specifically asks for this `ipaddr.js` parsing update
- the related #920 PR is already merged and only covers `assertNotCloudMetadataUrl`, which this change uses as the precedent

## Verification

```sh
cd admin
node --loader ts-node-maintained/esm --test tests/unit/private_url.spec.ts tests/unit/cloud_metadata_url.spec.ts
npm run typecheck
git diff --check
```

### DIFF
--- a/admin/app/validators/common.ts
+++ b/admin/app/validators/common.ts
@@ -14,24 +14,25 @@ import ipaddr from 'ipaddr.js'
  */
 export function assertNotPrivateUrl(urlString: string): void {
   const parsed = new URL(urlString)
-  const hostname = parsed.hostname.toLowerCase()
+  const hostname = parsed.hostname.toLowerCase().replace(/^\[|\]$/g, '')
 
-  // `URL.hostname` strips the surrounding brackets from IPv6 literals
-  // (e.g. `http://[::1]/` → hostname `::1`), so IPv6 patterns must match
-  // the unbracketed form.
-  const blockedPatterns = [
-    /^localhost$/,
-    /^127\.\d+\.\d+\.\d+$/,
-    /^0\.0\.0\.0$/,
-    /^169\.254\.\d+\.\d+$/, // Link-local / cloud metadata
-    /^::1$/, // IPv6 loopback
-    /^fe80:/i, // IPv6 link-local
-    /^::ffff:/i, // IPv4-mapped IPv6 (e.g. ::ffff:7f00:1 = 127.0.0.1)
-    /^::$/, // IPv6 all-zeros (equivalent to 0.0.0.0)
-  ]
-
-  if (blockedPatterns.some((re) => re.test(hostname))) {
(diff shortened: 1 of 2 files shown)


chriscrosstalk (approved):
Thanks @Gujiassh — this is a solid, more-robust replacement for the regex blocklist, and it matches NOMAD's intended policy: block literal loopback / link-local / unspecified addresses, allow RFC1918 and DNS names (LAN appliances need those).

Reviewed it closely for bypasses and regressions and it holds up:
- Covers everything the old regex did (127/8, 0.0.0.0, 169.254/16, ::1, fe80::/10, ::, and IPv4-mapped loopback/link-local via the toIPv4Address conversion).
- More precise on mapped addresses — the old `::ffff:` regex over-blocked mapped *public* IPs; this only blocks mapped addresses that are genuinely loopback/link-local after conversion.
- Cloud-metadata 169.254.169.254 stays blocked; bare LAN hostnames and external FQDNs pass through as DNS names exactly as before.
- Good call documenting that DNS-rebinding is a fetch-time concern outside this guard's scope.

One thing before merge: this is application code, so it should target `dev` rather than `main`. Could you retarget the branch? Once it's on `dev` this is good to go. Resolves #922.

chriscrosstalk:
Thanks again @Gujiassh — your ipaddr.js approach was the right direction and it's carried forward. While reviewing it alongside #912 we found the two were complementary but conflicting (both rewrite this same function): the ipaddr.js rewrite here dropped the trailing-root-dot handling, so `localhost.` would have slipped through. Since this is SSRF-critical code, we've consolidated both into a single in-house validator with a full test suite in #985, which keeps your range-classification approach and adds the trailing-dot normalization. Closing in favor of that. Appreciate the contribution and the thorough write-up.

PR #960, opened 2026-05-31
fix: patch vulnerabilities

yabets4:
$(cat VULNS.md)

### DIFF
--- a/install/sidecar-disk-collector/Dockerfile
+++ b/install/sidecar-disk-collector/Dockerfile
@@ -1,6 +1,9 @@
 FROM alpine:3.20
 RUN apk add --no-cache util-linux bash
+# Create non-root user for security (CWE-250: Execution with Unnecessary Privileges)
+RUN adduser -D -u 1000 nomaduser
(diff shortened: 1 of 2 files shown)


yabets4:
## [MEDIUM] CWE-250 — Docker Containers Running as Root
**CVSS:** 5.3 | **KEV:** N/A (Design weakness, not a CVE)

**Finding 1:** `install/sidecar-disk-collector/Dockerfile`
**Status:** FIXED — Added non-root user `nomaduser` with `USER` directive

**Finding 2:** `install/sidecar-updater/Dockerfile`
**Status:** FIXED — Added non-root user `nomaduser` with `USER` directive

**Details:**
- Both Docker containers were running as root by default (no USER directive)
- If an attacker compromises the container, they gain root privileges on the host (container escape risk)
- Applied fix: Created `nomaduser` (UID 1000) and added `USER nomaduser` directive before CMD

**Kill Chain:**
- Container compromise → Root access in container → Potential container escape → Host compromise
- Confidence: MEDIUM

PR #965, opened 2026-06-01
fix(rag): improve context-reliance hedging and use heading metadata a…

jakeaturner:
…t query time

Users often saw the assistant disclaim ("Sorry, I couldn't find specific context regarding X, but here's a general answer...") even when material that directly answered the query was embedded. Two compounding causes, both on the read/generation side:

1. The rag_context system prompt explicitly authorized the hedge: it told the model to fall back to general knowledge and "acknowledge the limitations," and mandated "According to the information available..." citations that pushed small models into meta-commentary preambles. Rewrote it to treat the retrieved context as the primary, authoritative source, lead with the answer, fall back to general knowledge *silently*, and never emit "couldn't find specific context" phrasing, while still allowing the model itself to make some "sanity-check" decisions if the retrieved context seems wildly incorrect or unrelated to the user's query.

2. The injected context labeled each block with a raw relevance score (e.g. "Relevance: 42.3%"). nomic-embed cosine scores for genuinely relevant passages sit ~0.4-0.6, so the number primed the model to distrust correct context. Dropped the model-visible score (it stays in the logs) and replaced it with a neutral source-title label.

Also added a conservative, score-scaled heading boost in rerankResults: when query keywords match a chunk's section/article title (ZIM metadata we already fetch), nudge its rank. Same diminishing-returns shape as the existing boosts and gated behind the existing semantic-quality threshold, so it can't promote a weak match.

Scope note: this removes the visible hedge and improves ranking of already-retrieved chunks. It does NOT fix retrieval recall — diluted 1500-token chunks that never reach dense search's top-k are unaffected. That's a follow-up (smaller chunks + BM25/RRF hybrid).

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -106,8 +106,17 @@ export default class OllamaController {
             `[RAG] Injecting ${trimmedDocs.length}/${relevantDocs.length} results (model: ${reqData.model}, maxResults: ${maxResults}, maxTokens: ${maxTokens || 'unlimited'})`
           )
 
+          // Label each context block with its source title when available (a neutral,
+          // honest provenance signal) but never the raw relevance score — nomic cosine
+          // scores for genuinely relevant passages sit ~0.4-0.6, and surfacing e.g.
+          // "42%" primes the model to distrust correct context. Scores stay in the logs
+          // above for debugging.
           const contextText = trimmedDocs
-            .map((doc, idx) => `[Context ${idx + 1}] (Relevance: ${(doc.score * 100).toFixed(1)}%)\n${doc.text}`)
+            .map((doc, idx) => {
+              const title = doc.metadata?.full_title || doc.metadata?.article_title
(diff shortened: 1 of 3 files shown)


chriscrosstalk (approved):
Reviewed all three files. Looks good to merge.

- Verified the metadata fields (`full_title`/`section_title`/`article_title`) exist on `RAGResult` and on the search return shape, so the controller label and the heading boost both resolve and typecheck.
- Confirmed the heading boost sits after the `MIN_SEMANTIC_THRESHOLD` early return, so it's genuinely quality-gated as described and can't promote a weak match. Boost magnitude is in line with the existing keyword/direct-match boosts.
- Prompt rewrite reads well and the scope note (fixes hedging + reranking, not recall) is accurate.

One tiny thing, non-blocking: the heading match is substring-based (`includes`), so a short keyword can match inside a larger heading word. It's consistent with the existing direct-match boost and capped/gated, so I'd leave it. 👍

hestela (approved):

bgauger (approved):

PR #988, opened 2026-06-07
Fix admin axios header typecheck

### LINKED ISSUE #987: Build failing on dev: axios 1.17.0 header typing breaks tsc (blocks all PRs)
## Summary

The `Build Admin` CI job is failing on `dev` during `node ace build`, which means **every PR branched off current `dev` fails CI too**. It's not any single PR's contents — it's a `dev`-wide break.

## Root cause

`admin/package.json` now pins `axios@1.17.0`. In 1.17, response header values are typed as `string | number | true | string[] | AxiosHeaders` instead of `string`, so several pre-existing spots that read headers as strings no longer compile:

- `admin/app/utils/downloads.ts:57` — `headers['content-type'] || '...'` later used with `.includes()` (TS2339)
- `admin/app/utils/downloads.ts:58` — `parseInt(headers['content-length'] || '0')` (TS2345)
- `admin/app/services/map_service.ts:303-304` — `parseInt(contentLength, 10)` (TS2345)

`collection_update_service.ts:160` uses `Number(...)` and is fine; the `accept-ranges === 'bytes'` comparison is also fine.

## Why it slipped through

`node ace build` runs a full emitting `tsc` which surfaces these. A local `tsc --noEmit` against an older cached `node_modules` (axios 1.15) passes, so it only shows on CI's fresh `npm ci` (which installs 1.17). Note: dependabot's axios PR #974 is closed-not-merged, so 1.17 reached `dev` by another route.

## Suggested fix

Coerce those header reads to string, forward-compatible across 1.15/1.17:

```ts
parseInt(String(headers['content-length'] ?? '0'))
String(headers['content-type'] ?? 'application/octet-stream')
```

Alternatively pin axios back to 1.15.0 if the bump wasn't intended. Once `dev` builds green again, the open PRs go green after they pick up the fix.

## Error output (CI)

```
app/services/map_service.ts(304,45): error TS2345: Argument of type 'string | number | true | string[] | AxiosHeaders' is not assignable to parameter of type 'string'.
app/utils/downloads.ts(58,31): error TS2345: ...
app/utils/downloads.ts(63,79): error TS2339: Property 'includes' does not exist on type 'string | number | true | string[] | AxiosHeaders'.
Cannot complete the build process as there are TypeScript errors.
```

liusy0802-ui:
## Summary
- coerce Axios response header values to strings before using parseInt/includes
- preserve the existing octet-stream fallback for missing content-type headers

## Validation
- npm run typecheck

Fixes #987

### DIFF
--- a/admin/app/services/map_service.ts
+++ b/admin/app/services/map_service.ts
@@ -300,7 +300,7 @@ export class MapService implements IMapService {
         throw new Error(`Failed to fetch file info: ${response.status} ${response.statusText}`)
       }
 
-      const contentLength = response.headers['content-length']
+      const contentLength = String(response.headers['content-length'] ?? '')
       const size = contentLength ? parseInt(contentLength, 10) : 0
 
       return { filename, size }
--- a/admin/app/utils/downloads.ts
+++ b/admin/app/utils/downloads.ts
@@ -53,9 +53,8 @@ export async function doResumableDownload({
   // already in every binary-content allowlist we use (ZIM, PMTILES, base assets).
   // Without this default, the validator below throws `MIME type  is not allowed`
   // and breaks all downloads from kiwix's primary host (#848).
-  const contentType =
-    headResponse.headers['content-type'] || 'application/octet-stream'
-  const totalBytes = parseInt(headResponse.headers['content-length'] || '0')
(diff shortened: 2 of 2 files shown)


jakeaturner:
I've just fixed this in 028072e on `dev` but appreciate the PR!

PR #991, opened 2026-06-08
feat(supply-depot): add custom launch URLs for apps

jakeaturner:
Let users override an app's "Open" link with a reverse-proxy or local-DNS address (e.g. https://jellyfin.myhomelab.net). Falls back to the default host+port when unset. Metadata-only — no container changes.

### What changed
- **New `custom_url` column** on `services` (migration). Stored separately from `ui_location` so the default link is always recoverable, and deliberately excluded from the service seeder so a curated app's override survives reseeds/upgrades.
- **`PUT /api/system/services/custom-url`** sets or clears the override. An empty value clears it; a non-empty value is normalized + validated. Hidden dependency services (e.g. Qdrant) are rejected (403).
- **`normalizeCustomUrl` helper** (frontend + backend twins): accepts a bare host or full URL, prepends `http://` when no scheme is present, and restricts to `http`/`https` — the XSS guard that keeps `javascript:`/`data:` URLs out of any `href`.
  - **`AppUrlModal`** — set/clear UI with inline validation and a live "Opens as:" preview of the resolved link.
- `getServiceLink` now takes an optional `customUrl` and prefers it over the computed default. Wired through the supply-depot cards, `/home` shortcuts, and the settings apps table.

### Scope / safety
- Metadata-only: no containers are created, recreated, or touched.
- `custom_url` is used **only** for user-facing links — never for internal service-to-service requests, so there's no SSRF surface.
- Validation runs on both client and server (defense in depth).

### Testing
- Manual: set/clear overrides on curated and custom apps; verified the link flows through to the card, `/home`, and settings.
- TODO: unit spec for `normalizeCustomUrl` (scheme allowlist, bare-host prefixing, empty-clears) — mirrors `custom_app_guard.spec.ts`.


chucktesta154k:
This fixed all apps within the supply depot but the logs button within settings that should launch to the dozzle container still has the same issue. Is there a ticket for this already?

chriscrosstalk:
Good catch, and confirmed. That link is built from just the port (`getServiceLink('9999')` in the Settings layout); Dozzle is a management container rather than a Supply Depot service row, so the custom URL override from this PR has nothing to attach to. There wasn't a ticket for it - opened #1324 to track it.

Issue #959, opened 2026-05-31
[Bug]: ollama has been doing "api/embed" for weeks now


annacegu:
### Issue Category

Installation/Setup

### Bug Description

I am trying to install qwen 2.5:3b as recommended by chat. When looking at knowledgebase it doens't show any documents loaded.  When I look at debug it has been doing "api/embed" for weeks now.

### Steps to Reproduce

1. install qwen 2.5:3b

### Expected Behavior

model installed

### Actual Behavior

model not in list of models being installed

### N.O.M.A.D. Version

1.32.0

### Operating System

Ubuntu 24.04

### Do you have a dedicated GPU?

No

### System Specifications

Project NOMAD Debug Info
========================
App Version: 1.32.0
Environment: production

System:
  OS: Ubuntu 24.04.4 LTS
  Hostname: 
  Kernel: 6.17.0-29-generic
  Architecture: x64
  Uptime: 8d 23h 36m

Hardware:
  CPU: Core™ i7-8700K (12 cores)
  RAM: 15.4 GB total, 5.2 GB used, 10.2 GB available
  GPU: CoffeeLake-S GT2 [UHD Graphics 630] (256 MB VRAM)
  Disk: 468.4 GB, 61% used, NVME

Installed Services:
  Information Library (nomad_kiwix_server): running
  Education Platform (nomad_kolibri): running
  AI Assistant (nomad_ollama): running
  Notes (nomad_flatnotes): running
  Data Tools (nomad_cyberchef): running

Internet Status: Online
Update Available: No (1.32.0 is latest)
Browser: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36 Edg/148.0.0.0

### Relevant Logs

```shell

```

### Browser Console Errors (if UI issue)

```javascript

```

### Screenshots

<img width="919" height="826" alt="Image" src="https://github.com/user-attachments/assets/830800fb-3fa8-4564-88a1-4f8a260becba" />

### Pre-submission Checklist

- [x] I have searched for existing issues that might be related to this bug
- [x] I am running the latest version of Project N.O.M.A.D. (or have noted my version above)
- [x] I have redacted any personal or sensitive information from logs and screenshots
- [x] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

chriscrosstalk:
[...]

Issue #944, opened 2026-05-26
[Bug]: Endless looping in the Queue for the AI KB upon initial installation with a Tesla P4 GPU


glabifrons:
### Issue Category

Installation/Setup

### Specs:
- i7-8700
- 16GB DDR4 RAM (soon to be 32GB)
- Tesla P4 8GB GPU
- 1TB SATA SSD
- Fresh Ubuntu Server 24.04 minimal installation with nouveau driver blocked and nvidia-driver-580-server nvidia-utils-580-server installed for GPU/CUDA use.

### Background and setup:
I just installed Project NOMAD on a system with the above specs, went through and selected the "additional content" I wanted for the knowledge base, and let it download and install everything.
I saw that it had errors at the bottom about the LLM models not downloading due to missing ollama, so I waited for the ollama docker container to be installed, then went through and re-selected the models.
The models downloaded successfully, then it started processing the knowledge base.

### Problem:
This is where it fell into a loop. For a few hours, it looped on 16 of them, repeatedly flagging them as "stalled" and retrying. Eventually it got down to 14 that are still in queue.
I'm not saying it's taking a long time to process them, it's looping over the same ones repeatedly. I watched as it cycled the entire list at least 4 times. Oddly, sometimes it retries ones that appeared to be successful (I copied, control-F, pasted to verify I wasn't misreading similar ones).

### My questions:
1. Where can I find logs to see what it's doing/what's failing?
2. How can I adjust the timeout, as it is retrying these awfully quickly (40 seconds to 2 minutes). I'm guessing it is expecting that if it sees a Tesla, it must be faster than an RTX 5090 or something. It's only a P4. Some of these are huge (like Gutenberg) and will take some time on even a really powerful GPU.
3. Why would it reprocess successful ones?
4. Why do the numbers not match up? (28 stalled vs 14 waiting)

### Steps to Reproduce

1. Install Tesla P4 (or equivalent?)
2. Install Ubuntu Server 24.04 using the entire disk.
Be sure to configure the bulk of the space in /opt as that's the target installation directory for this software and all its data.
3. Launch the installer script.
4. Select all "additional content", Comprehensive level for each.
5. Select all 3 AI model offerings.
6. Leave Auto-Index at default "Yes, always".
7. Complete Setup.
8. Click AI Assistant from Home Page
9. Click "Knowledge Base" at the bottom.
10. Observe Processing Queue repeatedly looping over the same failures, rather than waiting for them to complete.

[...]

PR #994, opened 2026-06-08
fix(AI): truncate-and-retry oversized embed chunks; stop 30x retry storm (#881)

chriscrosstalk:
## Problem

Dense source content (technical PDFs, encyclopedic prose, code/markup ZIMs) produces chunks that exceed the embedding model's context window (`nomic-embed-text:v1.5` defaults to 2048 tokens). Two paths still hit this even after the earlier input pre-cap:

1. **Older Ollama** (e.g. 0.18.1, #944) ignores the `num_ctx=8192` we send on `/api/embed`, staying at the model's 2048 default.
2. **The OpenAI-compat `/v1/embeddings` fallback** didn't pass `num_ctx`/`truncate` at all, so *any* Ollama version drops to 2048 whenever the native call errors and it lands on the fallback.

When a chunk overflowed, the 400 was swallowed and the chunk was **silently dropped from Qdrant**. Worse, the error propagated to `EmbedFileJob`, which re-embeds the **entire file** on each of its 30 BullMQ attempts: the "endless queue loop" / "api/embed for weeks" / pegged GPU reported in #944 and #959.

## Fix

- **`OllamaService.embed()`** — on a context-length error, retry once with an aggressive 2048-safe cap (`EMBED_CONTEXT_SAFE_CHARS = 2000`, safe even at ~1 char/token), so the chunk is embedded (start-of-chunk) instead of dropped. Native-path context errors now bubble up to this retry rather than falling through to the smaller-context fallback. The native+fallback attempt is split into `_embedWithFallback()`.
- **Fallback parity** — pass `truncate`/`num_ctx` on the `/v1/embeddings` call too (Ollama's OpenAI-compat shim forwards them).
- **`EmbedFileJob`** — classify `input length exceeds context length` as an `UnrecoverableError` so one permanently-oversized chunk can't trigger 30 full-file re-embeds. (Defense-in-depth for the rare case where even the truncated retry can't fit.)
- Add shared `OllamaService.isContextLengthError()`.

**Graceful degradation:** a truncated chunk loses its tail but stays in the index — strictly better than today's silent drop + retry storm. No global chunk-size change, so the happy path (modern Ollama, `num_ctx=8192`) is untouched and there's no chunk-count inflation.

## Testing
- `tsc --noEmit` clean.
- **Pending live validation:** dense-ZIM ingestion run on a CPU box (the documented repro: medlineplus / arduino.stackexchange) — confirm no `400 input length` storm, no full-file retry loop, and Qdrant `points_count` reflects truncated-and-kept chunks rather than drops.

Refs #881. Supersedes the #369/#670 symptom closures that never addressed the fallback path.

### DIFF
--- a/admin/app/jobs/embed_file_job.ts
+++ b/admin/app/jobs/embed_file_job.ts
@@ -238,23 +238,38 @@ export class EmbedFileJob {
         message: `Successfully embedded ${result.chunks} chunks`,
       }
     } catch (error) {
-      logger.error(`[EmbedFileJob] Error embedding file ${fileName}:`, error)
+      // A chunk that still exceeds the model's context after OllamaService's truncate-and-retry is
+      // permanently oversized for this install (e.g. a model whose context is smaller than our safe
+      // cap). Re-embedding the whole file 30x re-processes everything and can never succeed — that is
+      // the "endless queue loop" / "api/embed for weeks" (#881/#944/#959). Mark it unrecoverable so
+      // BullMQ stops after one pass instead of storming.
+      let normalizedError = error
+      if (!(error instanceof UnrecoverableError) && OllamaService.isContextLengthError(error)) {
+        logger.warn(
+          `[EmbedFileJob] Context-length overflow persisted for ${fileName} after truncation; not retrying.`
+        )
+        normalizedError = new UnrecoverableError(
+          error instanceof Error ? error.message : 'Embedding input exceeds the model context length'
+        )
+      }
+
+      logger.error(`[EmbedFileJob] Error embedding file ${fileName}:`, normalizedError)
 
       await job.updateData({
         ...job.data,
         status: 'failed',
         failedAt: Date.now(),
-        error: error instanceof Error ? error.message : 'Unknown error',
+        error: normalizedError instanceof Error ? normalizedError.message : 'Unknown error',
       })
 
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
## Validation (NOMAD1, branch built into cw-test image)

Exercised the exact failure and recovery against a live Ollama:

| Check | Result |
|---|---|
| `isContextLengthError` matcher (native-axios / OpenAI-BadRequest / bare-`.error` / unrelated / timeout / null) | 6/6 correct, null-safe |
| Reproduce the error: `/api/embed` `truncate:false` + 27K-char input (simulates old Ollama / the old param-less fallback) | **HTTP 400 `{"error":"the input length exceeds the context length"}`** |
| Retry behavior: same oversized input with `truncate:true` | **OK, 768-dim vector** (embedded, not dropped) |
| `/v1/embeddings` fallback accepts the added `truncate`/`num_ctx` params | **OK, 768-dim vector** (shim forwards them, no breakage) |

The reproduced 400 string matches what the matcher keys on, and the truncated retry recovers the chunk. `truncate:false` is functionally what an old 0.18.1 Ollama (#944) does and what the param-less fallback did, so the mechanism is proven without needing an old-Ollama box. `tsc --noEmit` clean.

jakeaturner (approved):

PR #1005, opened 2026-06-09
fix(seeder): skip catalog sync for installed apps on diverged host ports

johno10661:
## Problem

The catalog sync in `ServiceSeeder.run()` (added with the supply depot work) overwrites `container_config` and `ui_location` for every curated service not flagged `is_user_modified`. That flag is new in 1.33.0, so it is `false` for all rows that predate it — including apps that were deliberately installed on an alternate host port because the catalog default was already taken on that host.

On the first boot after upgrading to 1.33.0, the seeder resets such a record back to the catalog port while the actual container keeps running on the alternate port. The record is now desynced from reality:

- the app's open link points at the wrong port (broken, or worse, at whatever unrelated service occupies the catalog port), and
- a later update/reinstall recreates the container on the catalog port, which fails or collides with the service that occupies it.

**Real-world case:** a host running Consul (server RPC on 8300) had Kolibri installed on 8310. After upgrading to 1.33.0-rc.1, the auto-seed reset Kolibri's `ui_location` to 8300, so the Supply Depot card linked to Consul's RPC port and the app appeared dead.

## Fix

Before syncing a row, compare the **published host ports** of the existing record against the catalog entry. If the app is installed and its host ports diverge, skip the sync for that row — the deployed container is the source of truth.

Only host ports are compared, deliberately, so the corrections this sync exists for still propagate:

- internal container-port fixes (e.g. Meshtastic Web 80 → 8080) still sync — host port unchanged
- `ui_location` scheme changes (e.g. Vaultwarden → `https:8480`) still sync — host port unchanged
- unparseable/absent configs on either side fall through to the existing sync behavior

Fresh installs, custom apps, and user-modified rows are unaffected.

### DIFF
--- a/admin/database/migrations/1778700000007_backfill_user_modified_for_diverged_ports.ts
+++ b/admin/database/migrations/1778700000007_backfill_user_modified_for_diverged_ports.ts
@@ -1,0 +1,90 @@
+import { BaseSchema } from '@adonisjs/lucid/schema'
+
+/**
+ * Catalog default published host ports for curated services, as of the 1.33
+ * curated-catalog era (when the `is_user_modified` flag was introduced).
+ *
+ * A curated install whose live host port differs from its catalog default was
+ * deployed on an alternate port (commonly because the default was already taken
+ * on that host) before `is_user_modified` existed to record it. This is a
+ * point-in-time snapshot on purpose — it must not track later catalog changes.
+ */
+const CATALOG_DEFAULT_HOST_PORTS: Record<string, string[]> = {
+  nomad_kiwix_server: ['8090'],
+  nomad_ollama: ['11434'],
+  nomad_cyberchef: ['8100'],
+  nomad_flatnotes: ['8200'],
+  nomad_kolibri: ['8300'], // legacy (Gen 1)
+  nomad_kolibri_2: ['8310', '8311'],
+  nomad_stirling_pdf: ['8400'],
+  nomad_filebrowser: ['8410'],
+  nomad_calibreweb: ['8420'],
+  nomad_it_tools: ['8430'],
+  nomad_excalidraw: ['8440'],
+  nomad_meshtastic_web: ['8450'],
+  nomad_homebox: ['8470'],
+  nomad_vaultwarden: ['8480'],
+  nomad_jellyfin: ['8490'],
+  nomad_meshcore_web: ['8500'],
+}
+
+/** Sorted list of published host ports in a serialized container_config, or null. */
+function hostPortsOf(containerConfig: string | null): string[] | null {
+  if (!containerConfig) return null
+  try {
+    const bindings = JSON.parse(containerConfig)?.HostConfig?.PortBindings ?? {}
+    const ports = Object.values(bindings)
+      .flat()
+      .map((b: any) => b?.HostPort)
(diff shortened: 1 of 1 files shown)


chriscrosstalk:
Thanks for the detailed writeup. The diagnosis is right, and I confirmed the catalog-sync loop still resets a deliberately-diverged host port on current `dev`, so this is a real bug. The blocker is that this branch has drifted a long way from `dev` (117 files, since the whole curated catalog was reworked after you opened it), so it cannot go in as-is.

If you are up for it, could you rebase onto current `dev` and pare it down to just the seeder fix? A focused diff on `service_seeder.ts` is something we can review and merge quickly, and it is the part that matters here. One steer on approach: the lightest path is probably a one-time backfill that flags pre-1.33 curated installs whose host port differs from the catalog default as `is_user_modified`, so the existing sync loop leaves them alone. Just note that means those installs stop receiving future catalog config changes, which is a reasonable tradeoff for keeping their port.

Appreciate you tracking this one down.

chriscrosstalk:
Thanks again for confirming the diagnosis, and it holds up. The blocker is still the diff size: this branch carries 117 files and ~13k lines of unrelated churn from an old base, and we can't merge that into `dev` as-is. Two ways forward. Either rebase this down to just the seeder change (the `container_config` / `ui_location` guard) on a fresh branch off current `dev`, or if that's more trouble than it's worth, tell us and we'll reimplement the same fix in-house and credit you on the PR. Either way the fix lands, we just can't take the bundle.

johno10661:
Thanks for the detailed review and for confirming the diagnosis. I've rebased onto current `dev` and pared this down to a single self-contained migration, following your suggested approach.

`1778700000007_backfill_user_modified_for_diverged_ports.ts` is a one-time backfill: for each installed, curated (non-custom), not-yet-`is_user_modified` service whose published host port differs from the catalog default, it sets `is_user_modified`. It runs before `db:seed`, so the flag is in place before the sync loop would reset the port. The catalog defaults are a point-in-time snapshot inside the migration (so it won't drift with later catalog changes), and I included the legacy `nomad_kolibri` (8300) alongside the current curated set.

As you noted, the tradeoff is those installs stop receiving future catalog config changes — the right call for keeping a working port. The diff is now just the one migration file.

chriscrosstalk:
Thanks for cutting this down, and sorry for the slow reply. The rebase is exactly what we asked for.

We have taken a different approach to the same bug in #1300, and the diagnosis credit there is yours. Rather than flagging affected rows `is_user_modified`, the seeder now keeps an installed app's published host ports and applies the rest of the catalog change. Same protection for the port, without opting those installs out of future catalog updates, and it needs no backfill because it also covers divergence that happens later.

One thing worth knowing for future work in this area, because it cost me an hour today. `services.container_config` is a MySQL `json` column, so the driver hands back an already-parsed object even though the model types it `string | null`. That means `JSON.parse(row.container_config)` throws, and inside your migration's try/catch it returns null and hits the `continue`, so I believe the backfill would have flagged no rows at all. My own first attempt had exactly the same bug and passed every unit test before I ran it against a real database.

Closing this in favour of #1300. Genuinely useful find, and the port-table snapshot idea shaped how we thought about it.

PR #1006, opened 2026-06-09
feat(supply-depot): add uninstall for curated apps

johno10661:
## Problem

Curated catalog apps can be installed, started/stopped, and force-reinstalled — but never removed. Once installed, the only way to get an app off a device is manual `docker rm` plus editing the `services` row by hand. Custom apps already have a Delete action; curated apps have no equivalent.

## What this adds

**API:** `POST /api/system/services/uninstall` (`{ service_name, remove_image? }`)

- stops and removes the app's container
- optionally removes the backing image (same best-effort semantics as custom app delete: shared/in-use images are left alone)
- flips the record back to `installed = false` / `idle`, so the card returns to the available catalog

Host bind-mount data is **deliberately left on disk** — a later reinstall picks the app back up where it left off. This is the key contrast with force-reinstall, which clears volumes and immediately recreates the container.

Guards:
- custom apps → 403 (they're removed via the existing delete endpoint, which also drops the DB record)
- dependency services → 403
- not-installed apps → 409

**UI:** installed curated cards get an **Uninstall** action in the card menu (mirroring Delete on custom cards), with a confirm modal that explains app data stays on disk and offers the same "also remove the Docker image" checkbox as custom app delete.

## Implementation notes

- Container/image removal reuses `removeCustomAppContainer` — the lookup is by container name, which is the `service_name` for curated and custom apps alike.
- Uninstall is offered on all installed curated apps, consistent with Force Reinstall (which is offered everywhere and is more destructive, since it clears volumes). Happy to scope it down (e.g. hide on core services like Kiwix/Ollama) if you'd prefer.

## Testing

Verified on a live install (1.33.0-rc.1 base):
- installed Notes (flatnotes) from the catalog, uninstalled it via the new card action with "remove image" checked → container and image removed, record back to not-installed, card returned to Available, data dir preserved
- endpoint guards verified directly: unknown service → 404, not-installed service → 409
- `tsc --noEmit` and eslint clean on all touched files

### DIFF
--- a/admin/app/controllers/system_controller.ts
+++ b/admin/app/controllers/system_controller.ts
@@ -18,6 +18,7 @@ import {
   preflightValidator,
   serviceLogsValidator,
   subscribeToReleaseNotesValidator,
+  uninstallServiceValidator,
   updateCustomAppValidator,
   updateServiceValidator,
   setServiceAutoUpdateValidator,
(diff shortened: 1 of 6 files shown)


jakeaturner (approved):
Great contribution, thank you! Eventually, we'll probably need to support a `removeVolumes` option for `uninstallService` to cleanup orphaned data, but this PR is great and addresses a real need.

PR #1009, opened 2026-06-10
fix(queue): share one ioredis connection across BullMQ queues and workers

### LINKED ISSUE #885: bug(queue): internal embed pipeline opens ~1 fresh ioredis client per second under sustained ingestion — distinct from #877's QueueService fix
# bug(queue): internal embed pipeline opens ~1 fresh ioredis client per second under sustained ingestion — distinct from #877's QueueService fix

## Summary

After PR #877 collapsed `QueueService` to a process-wide singleton (24 call sites confirmed in the built bundle), the **embed processing pipeline itself** still opens fresh ioredis clients during sustained ZIM ingestion. Under load on NOMAD1 today (CPU-only, 4 active multi-batch ZIMs), the leak rate measured at:

- **~1 new ioredis client per second** with zero external API polling
- **~30 new clients per embed-batch completion**
- **Will hit Redis `maxclients` (10,000 default) in roughly 2–3 hours** of continuous embedding

This is a **separate root cause** from #877. The QueueService singleton is doing its job. The leak is downstream.

## What we know

Forensic evidence captured from NOMAD1 mid-leak (redis client count growing from 53 → 499 over ~10 minutes):

1. **PID distribution.** Of ~515 leaked file descriptors observed inside `nomad_admin`, **all but ~49 were held by PID 1 (`node bin/server.js`)** — the web server, not the `node ace queue:work --all` worker (PID 83).

2. **All leaked connections show Lua script activity.** Sample from `redis-cli CLIENT LIST`:
   ```
   id=5283 addr=172.19.0.11:45060 fd=76 name= age=435 idle=435
   flags=N db=0 cmd=evalsha lib-name=ioredis lib-ver=5.9.2
   ```
   Every leaked socket has `cmd=eval` or `cmd=evalsha` as its last command, then sits idle. `lib-name=ioredis` confirms these are BullMQ's underlying client, not a direct app-level `new Redis()` call.

3. **Connections are created in groups of 3 every ~3 seconds.** The age distribution shows perfectly consistent triples at increments of 3:
   ```
   3 connections aged   1s
   3 connections aged   4s
   3 connections aged   7s
   3 connections aged  10s
   ...
   ```
   The 3-second cadence aligns with the BullMQ continuation-batch dispatch interval during multi-batch ZIM processing.

4. **The leak is in the processing path, not the API.** Isolation test:
   - With **no external API polling, just embed running**: +60 clients in 60 sec (the baseline 1/sec leak).
   - With **10 calls to `/api/rag/active-jobs`**: +3 clients (0.3/call, minor).
   - With **10 calls to `/api/rag/files`**: +12 clients (1.2/call, modest).

   So ~95% of the leak is generated internally by the embed pipeline, not by API hits.

5. **Bullmq + ioredis versions:** `bullmq@5.67.2`, `ioredis@5.9.2`.

6. **`queueConfig.connection` is a plain `{host, port}` object** (not a shared ioredis instance):
   ```ts
   const queueConfig = {
     connection: {
       host: env.get('REDIS_HOST'),
       port: env.get('REDIS_PORT') ?? 6379,
     },
   }
   ```
   When BullMQ is fed a config object, it instantiates a fresh ioredis client internally per Queue / Worker / QueueEvents — and some operations may further spawn temporary clients. **This is the likely root cause:** without a shared ioredis instance, BullMQ has no pool to reuse.

## Repro

NOMAD1, hot-patched with the current `rc` branch (post-#877). 6 ZIMs queued for embedding, CPU-only `nomic-embed-text:v1.5`. Run for ~5 minutes:

```bash
# Baseline (just after admin restart)
docker exec nomad_redis redis-cli CLIENT LIST | wc -l
# ~25

# After 5 minutes of sustained embedding
docker exec nomad_redis redis-cli CLIENT LIST | wc -l
# ~300

# After 10 minutes
# ~600

# After 30 minutes
# ~1800

# Project: hits 10k around 2.5–3 hours
```

Stopping the admin container drains the connections instantly. Same fingerprint as the #877 era leak; same recovery.

## Why this matters

- PR #877 was the GA blocker because the leak was operationally fatal. The fix landed in rc.3.
- This new leak has the same operational signature (steady accumulation under sustained ingestion → admin restart required) but a different root cause.
- For long-tail ingests (Wikipedia, large curated tier changes), users will see admin become unresponsive within hours.
- The workaround in the meantime is a periodic admin restart, which loses 1–2 minutes of in-flight batch progress each time but otherwise preserves state via BullMQ persistence.

## Suggested fix direction

The textbook BullMQ pattern for production is to **pass a shared ioredis instance** (not a config object), so all Queue / Worker / QueueEvents share the same client pool:

```ts
// config/queue.ts
import IORedis from 'ioredis'
import env from '#start/env'

const sharedConnection = new IORedis({
  host: env.get('REDIS_HOST'),
  port: env.get('REDIS_PORT') ?? 6379,
  maxRetriesPerRequest: null,   // required by BullMQ when sharing
})

const queueConfig = {
  connection: sharedConnection,
}

export default queueConfig
```

Caveats from the BullMQ docs:

- `maxRetriesPerRequest: null` is mandatory for shared instances (Worker requires it).
- BullMQ may still spawn a separate **BClient** for blocking pops (1 per Worker); that's expected and doesn't leak.
- Some short-lived operations (`getJobs`, certain Lua scripts) will reuse the shared connection rather than spawning new ones.

After the fix, the expected steady-state redis client count for NOMAD's queue topology (7 queues × ~3 internal clients each) should sit at roughly 20–30, regardless of throughput. Today's run started there before drifting.

## Other clues / non-issues

- **Not the AdonisJS Queue command.** `commands/queue/work.js` creates 1 Worker per queue at startup with `connection: queueConfig.connection`. That's the source of ~14 connections at boot (7 queues × 2 each), which is fine. The growing leak is downstream.

- **Not `listActiveJobs` or `dispatch` directly.** Both go through `QueueService.getInstance().getQueue(name)`, which caches the Queue. The Queue itself is reused. The fresh-client behavior is happening *inside* BullMQ when it executes script commands against that Queue.

- **Not session-specific.** Started clean after an admin restart, took ~6 minutes to first cross the previous 25-client baseline, then climbed steadily. Replicates reliably.

## Workaround until fix lands

Periodic admin restart (every 60–90 minutes during sustained ingestion). The monitor used in today's test does this automatically when the count crosses 2000:

```bash
# After every redis check
if [ "$rc" -gt 2000 ]; then
  docker restart nomad_admin
fi
```

Loses ~1–2 minutes of in-flight batch progress each restart (BullMQ persistence handles the rest). Acceptable bridge until the shared-connection fix ships.

## Related

- **#877** — original QueueService singleton fix. Confirmed working; this is a separate root cause downstream.
- **RFC #883** — KB Ingestion UX redesign references this leak as a companion bug that must be fixed alongside the redesign.
- BullMQ docs on shared connections: https://docs.bullmq.io/guide/connections

## Severity

**Operationally fatal under sustained ingestion** — same severity class as #877. Not a v1.32.0 GA blocker (the partial fix in #877 already shipped), but should be next-up alongside the RFC #883 redesign that depends on a stable embed pipeline for accurate progress and status displays.

LorenzoGalassi:
## Summary

Closes #885

`config/queue.ts` hands BullMQ a plain `{host, port}` connection object. BullMQ treats that as a recipe: every `Queue` and `Worker` instantiates its own ioredis client from it, and script commands executed against those clients can spawn further short-lived connections. As documented in #885, under sustained ZIM ingestion this leaks ~1 client/sec on the admin process until Redis `maxclients` (10k) is exhausted in ~2–3 hours, forcing periodic admin restarts.

This PR implements the fix direction suggested in the issue — the [BullMQ-documented production pattern](https://docs.bullmq.io/guide/connections) of passing a single shared ioredis instance so all Queues and Workers reuse one client pool:

- **`admin/config/queue.ts`** — construct one shared `Redis` instance (preserving the `REDIS_DB` support from #939) and export it as `queueConfig.connection`. `maxRetriesPerRequest: null` is set, as BullMQ requires for shared connections.
- **`admin/package.json` / lockfile** — add `ioredis@5.10.1` as a direct dependency (previously only transitive via bullmq; pinned to the version already in the lockfile).
- **`admin/app/services/queue_service.ts`** — comment-only update: the singleton rationale from #877 still applies, but the per-Queue connection-count claim is no longer accurate with a shared instance.

No call sites change: `QueueService` and `commands/queue/work.ts` already read `queueConfig.connection`, so both the web process and the worker process pick up the shared instance automatically. Workers duplicate the connection once for their blocking client, which is expected and bounded (1 per worker).

## Testing

- `npm run typecheck` and `eslint` pass clean.
- Synthetic steady-state test against a throwaway `redis:7-alpine`, simulating NOMAD's topology (7 queues + 7 workers at the same concurrency settings, sustained `add` + `getJobCounts` + `getJobs` dispatch for 45 s, sampling `CLIENT LIST` every 5 s):

  | connection style | clients at t=0 | clients at t=45s |
  |---|---|---|
  | `{host, port}` object (current) | 15 | 22 (one per Queue + two per Worker) |
  | shared ioredis instance (this PR) | 9 | 9 (flat — 1 shared + 1 blocking dup per worker) |

  The post-fix steady state matches the ~20–30 client expectation from the issue once scaled to the full app. Caveat, in the interest of full transparency: the synthetic run did **not** reproduce the unbounded ~1 client/sec growth — that appears to need the real embed pipeline's dispatch pattern — so verification on a NOMAD instance under sustained multi-batch ZIM ingestion (the #885 repro: watch `docker exec nomad_redis redis-cli CLIENT LIST | wc -l` over ~10 min of embedding) would be a valuable confirmation before merge.

## Suggested release note

> Fixed a Redis connection leak during sustained Knowledge Base ingestion that could make the admin UI unresponsive after a few hours (#885).

### DIFF
--- a/admin/app/services/queue_service.ts
+++ b/admin/app/services/queue_service.ts
@@ -1,11 +1,11 @@
 import { Queue } from 'bullmq'
 import queueConfig from '#config/queue'
 
-// Process-wide singleton. Each `Queue` opens two ioredis connections (one for
-// commands, one blocking). Instantiating a fresh QueueService per dispatch /
-// status lookup leaks both, and under sustained job churn (e.g. multi-batch ZIM
-// ingestion enqueueing a continuation every few seconds) it saturates Redis's
-// maxclients within hours.
+// Process-wide singleton. Instantiating a fresh QueueService per dispatch /
+// status lookup leaks connections, and under sustained job churn (e.g.
+// multi-batch ZIM ingestion enqueueing a continuation every few seconds) it
+// saturates Redis's maxclients within hours. All queues additionally reuse the
+// single shared ioredis instance exported from #config/queue (#885).
 export class QueueService {
   private queues: Map<string, Queue> = new Map()
 
--- a/admin/config/queue.ts
+++ b/admin/config/queue.ts
@@ -1,11 +1,23 @@
 import env from '#start/env'
+import { Redis } from 'ioredis'
+
+// BullMQ treats a plain `{host, port}` connection object as a recipe: every
+// Queue / Worker instantiates its own ioredis client from it, and script
+// commands executed against those clients can spawn further short-lived
+// connections. Under sustained ZIM ingestion this leaked ~1 client/sec until
+// Redis maxclients was exhausted (#885). Passing a single shared ioredis
+// instance instead gives BullMQ a pool to reuse — Workers still duplicate it
+// once for their blocking client, which is expected and bounded.
+// `maxRetriesPerRequest: null` is mandatory for connections shared with BullMQ.
+const sharedConnection = new Redis({
+  host: env.get('REDIS_HOST'),
+  port: env.get('REDIS_PORT') ?? 6379,
+  db: env.get('REDIS_DB') ?? 0,
+  maxRetriesPerRequest: null,
+})
 
 const queueConfig = {
-  connection: {
-    host: env.get('REDIS_HOST'),
-    port: env.get('REDIS_PORT') ?? 6379,
-    db: env.get('REDIS_DB') ?? 0,
-  },
+  connection: sharedConnection,
 }
 
 export default queueConfig
--- a/admin/package-lock.json
+++ b/admin/package-lock.json
@@ -47,6 +47,7 @@
         "edge.js": "6.4.0",
         "fast-xml-parser": "5.7.0",
         "fuse.js": "7.1.0",
+        "ioredis": "5.10.1",
         "ipaddr.js": "2.4.0",
         "jszip": "3.10.1",
         "luxon": "3.7.2",
--- a/admin/package.json
+++ b/admin/package.json
@@ -100,6 +100,7 @@
     "edge.js": "6.4.0",
     "fast-xml-parser": "5.7.0",
     "fuse.js": "7.1.0",
+    "ioredis": "5.10.1",
     "ipaddr.js": "2.4.0",
     "jszip": "3.10.1",
     "luxon": "3.7.2",


jakeaturner (approved):
Solid fix, thanks for the contribution!

PR #1019, opened 2026-06-15
feat: replace legacy Kolibri image default with latest v19 image

jakeaturner:
## Summary
Replaces the ~6-year-old `treehouses/kolibri:0.12.8` with the upstream-official `learningequality/kolibri:0.19.4`, shipped as a distinct **"Education Platform (Gen 2)"** catalog entry rather than an in-place upgrade. **Note: "Gen 2" is an moniker local to NOMAD to distinguish the two _very_ distinct default Kolibri versions shipped, not an official designation from the Kolibri project.**

An in-place upgrade isn't safe: the new image uses a different repo, mounts at `/kolibri` instead of `/root/.kolibri`, and crosses 7 minor versions of Kolibri's own data schema. Nomad's normal update rail preserves the old binds and assumes the same image repo, so it can't carry this change. Existing `0.12.8` installs keep running untouched; users opt into Gen 2 and re-import their content.

## Changes
 - **New catalog entry** (`service_seeder.ts`, `service_names.ts`): `learningequality/kolibri:0.19.4` as `nomad_kolibri_2`, bind `…/kolibri-gen2:/kolibri`, ports **8310** (web) + **8311** (zip content), on a distinct host folder so it never collides with legacy data.
- **Sunset mechanism** (new migration + `is_deprecated` column on `services`): a generic deprecation flag. The migration **deletes** the legacy `nomad_kolibri` row where `installed = false` (orphaned catalog rows on existing deployments) and **flags it deprecated** where `installed = true`, so a running legacy container stays manageable. `getServices()` hides deprecated apps unless installed — once a holdout uninstalls, the legacy app leaves the catalog for good instead of re-offering the dead image.
- **UI** (`supply-depot.tsx`): a `legacy` badge on deprecated cards pointing users to the current Education Platform.
- **First-run fix** (`easy-setup/index.tsx`): the setup wizard installed the legacy service name, which no longer exists in the seeder— repointed to Gen 2.
- **Docs** (`supply-depot-apps.md`, `supply_depot_docs.ts`): a Kolibri section covering fresh install, channel re-import, and the legacy upgrade path, with its `#kolibri` anchor wired up.

## Testing
- `npm run typecheck` passes; new migration is recognized as pending.
- install Gen 2 and confirm it starts on 8310 and coexists with a legacy 8300 install; verify the migration's delete/flag behavior across installed vs. uninstalled legacy rows: passes

### DIFF
--- a/admin/app/models/service.ts
+++ b/admin/app/models/service.ts
@@ -85,6 +85,16 @@ export default class Service extends BaseModel {
   @column()
   declare category: string | null
 
+  // When true the service is sunset: hidden from the install catalog unless it is already
+  // installed (see SystemService.getServices). Lets a deprecated app stay manageable for users who
+  // still run it while keeping new users from installing it.
+  @column({
+    serialize(value) {
+      return Boolean(value)
+    },
+  })
+  declare is_deprecated: boolean
+
   @column()
   declare source_repo: string | null
 
--- a/admin/app/services/system_service.ts
+++ b/admin/app/services/system_service.ts
@@ -333,9 +333,15 @@ export class SystemService {
         'auto_update_enabled',
         'is_custom',
         'is_user_modified',
+        'is_deprecated',
         'category'
       )
       .where('is_dependency_service', false)
+      // Deprecated/sunset apps stay visible only while still installed, so the user can manage and
+      // uninstall them — they never reappear in the install catalog once removed.
+      .where((q) => {
+        q.where('is_deprecated', false).orWhere('installed', true)
+      })
     if (installedOnly) {
       query.where('installed', true)
     }
@@ -367,6 +373,7 @@ export class SystemService {
         auto_update_enabled: service.auto_update_enabled,
         is_custom: service.is_custom,
         is_user_modified: service.is_user_modified,
+        is_deprecated: service.is_deprecated,
         category: service.category,
       })
     }
--- a/admin/constants/service_names.ts
+++ b/admin/constants/service_names.ts
@@ -5,6 +5,7 @@ export const SERVICE_NAMES = {
   CYBERCHEF: 'nomad_cyberchef',
   FLATNOTES: 'nomad_flatnotes',
   KOLIBRI: 'nomad_kolibri',
+  KOLIBRI_GEN2: 'nomad_kolibri_2',
   // Supply Depot — curated catalog (ports 8400–8499)
   STIRLING_PDF: 'nomad_stirling_pdf',
   FILEBROWSER: 'nomad_filebrowser',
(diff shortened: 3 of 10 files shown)


chriscrosstalk:
Tested this on a v1.33.0-rc.1 box that already had the legacy Kolibri installed with ~69 GB of content. The Gen-2-as-separate-catalog-entry design and the deprecate-legacy migration both behave exactly as intended: coexistence on 8300 vs 8310/8311, the Legacy badge, install via the catalog, and HTML5/zip content serving on 8311 all check out, and the legacy install and its data were never touched. Nice approach.

One thing worth weighing on the existing-user upgrade path before GA.

The docs frame it as "installs fresh, re-import your channels," which reads as a full re-download. That's true for user accounts and learner progress (the main-DB schema jump genuinely can't be carried safely), but content looks like a separable problem: Kolibri's own local disk import can register existing channel files with no internet. I validated this against the real legacy data on the test box:

- `kolibri manage importchannel disk <channel_id> <legacy_home>` imported a legacy `treehouses/0.12.8` Khan Academy channel DB directly into the 0.19.4 install (cross-version import succeeded).
- `kolibri manage importcontent --node_ids <ids> disk <channel_id> <legacy_home>` pulled content files straight from the old `storage/kolibri` folder with no download, and the resulting perseus exercise served correctly through the 8311 zip server.

So the heavy part, the gigabytes already on disk, can move across locally. Only accounts and progress need rebuilding.

Proposal: an optional "Migrate content from legacy Kolibri" step on the Gen 2 install (or a button on the Legacy card) that enumerates channels from `storage/kolibri/content/databases/*.sqlite3`, runs importchannel/importcontent disk for each against the legacy home, and leaves accounts/progress as the only manual step. The blobs could be hardlinked rather than copied (same filesystem) to keep it near-instant and avoid temporary 2x disk usage.

Cross-version import worked cleanly on the channels I tried; worth confirming across the wider channel set. Can share the full test transcript if useful.

jakeaturner:
@chriscrosstalk great findings! Love the hardlink migration approach -- let me see what I can work up

jakeaturner:
@chriscrosstalk After a deep dive, I realized that the hardlink approach just isn't stable enough to be feasible for most installs and also has some explicit cases that simply won't work (e.g. across filesystems if NOMAD data stored on different device).

Also considered spinning up an ephemeral container to run the `kolibri manage importchannel` commands, but realized this is exactly what the local network import option inside Kolibri is for -- and almost certainly much more stable and polished. So, added documentation with the exact steps for users to run this import and a "How to migrate content {to|from} {Gen 1|Gen 2}" link on both of the Education Platform app cards.

Not the perfect solution, but gives users more explicit control and avoids NOMAD needing to maintain fragile spaghetti logic that would still need to temporarily duplicate disk space anyways

PR #1032, opened 2026-06-22
Fix: update Wikipedia ZIM collection

gheja:
The current collection in the repo is outdated, those versions of the ZIMs are no longer hosted.

Updated it to the latest available versions as of today.


chriscrosstalk:
Thanks for this. You're right that those dated ZIM URLs had rotated out and were 404ing. We've got the same refresh already staged in #1046, which also fixes the CI validator that's supposed to catch this rot automatically, so I'm going to close this in favor of that one. Appreciate you flagging it.

PR #1040, opened 2026-06-23
Add offline FDA drug reference (labels, interaction view, conditions, remedies)

caweis:
## What this adds

An offline FDA drug reference for the admin app. Three things on top of one new MySQL table:

- Drug label search and detail over the openFDA structured labels. Search by name or indication; the detail page renders each label's sections (boxed warning, indications, dosage, warnings, interactions, contraindications) in clinical order.
- A side-by-side label-interaction view. Pick up to 5 drugs and see each one's own FDA-labeled interaction text next to each other. This is not a drug-interaction checker. It never claims drug A interacts with drug B; it renders each label's Section-7 text verbatim, with a standing disclaimer saying exactly that. A pairwise (TWOSIDES-style) checker is deliberately out of scope.
- A condition cross-reference ("when to use what"). Look up a symptom or condition and see which OTC labels list it in their indications, alongside curated self-care and natural-remedy notes.

## Data and licensing

The labels come from the openFDA bulk export (api.fda.gov), U.S. public domain under CC0 1.0. Nothing is bundled in the repo or image: a two-step BullMQ job downloads the export at runtime (about 1.7 GB transient, reclaimed after ingest) and streams it into the `drug_labels` table. The curated condition and remedy text is paraphrased from public-domain federal sources (NCCIH, CDC, NIH/MedlinePlus). Every page carries the openFDA CC0 attribution and a "not affiliated with the FDA" footer.

## A note on the curated content

The condition and remedy entries are hand-written paraphrases of those federal sources. They make affirmative self-care statements (herb dosing, the "never give honey to infants under 1 year" caution, and similar). They ship with a mandatory SafetyBanner (informational only, not medical advice, not an FDA endorsement, call emergency services in an emergency) and have not had a formal clinician review. Whether to expose that part to users, gate it, or drop it is your call. It's kept behind the banner and isolated, so it's easy to pull if you'd rather ship label-search only.

## How it fits

It reuses what's already here: BullMQ and Redis for the jobs, Lucid/MySQL, the kv_store schema, VineJS validators, and the Inertia/React page conventions. Migrations post-date the current max, so no renumber. Two new runtime deps, both pure JS and arch-neutral: `yauzl` (unzip) and `stream-json` (streaming parse). No native build, no platform-specific code, so it runs the same on amd64 and arm64.

## Testing

Four standalone suites (interaction-text parsing, ingest status, conditions, natural-remedies JSON sync) plus unit specs for parsing and label mapping. tsc is clean on the feature code. I haven't run the migration or a full boot against MySQL/Redis locally; that part is for CI here.

## Open questions

- Pure helpers in `util/` or `app/utils/`? I see both in the tree.
- The curated condition/remedy content: keep it, gate it behind clinician review, or drop it for label-search only?
- Does CI bring up MySQL and Redis? If so I'll add integration tests for the job handlers; if not, the pure-logic suites are the gate.
- Home tile: I added two (Drug Reference, "When to use what") with a placeholder icon and order; adjust to fit your dashboard.


chriscrosstalk:
This is a great addition and squarely on-mission. We'd like to take it, with one direction change before we merge: we want it as **opt-in curated content**, not a built-in feature that ships enabled for everyone. Could you rework it toward the scope below?

**Where it lives:**
- Surface it inside the existing **Medicine** curated collection, in the **Standard** tier (`medicine-standard`). Selecting that tier triggers the download + ingest, the same way picking a tier installs its ZIMs today.
- Keep it integrated rather than Dockerized. The admin app staying the runtime is the right model here, no separate container.
- Gate the UI behind an install flag: no home-screen card or nav entry until the data is installed. Uninstall drops the table and removes the on-disk data.
- For freshness, wire the manifest `export_date` check into the content auto-update path so the dataset updates alongside ZIMs and maps instead of going stale.

**Download/ingest UI:**
- The Active Downloads indicator isn't ZIM-specific, it aggregates BullMQ jobs (ZIMs, map extracts, model pulls). Hook the **download** phase in as a new filetype reporting the same `{ percent, downloadedBytes, totalBytes }` shape, with aggregate percent across all the partitions. The job has to sit on a queue the download service reads and honor the existing cancel/remove buttons.
- Keep the heavy **ingest** phase in its own status surface (the `IngestStatus` UI you already built), the same way the Knowledge Base separates ingest from downloads.
- One UX must-have: make the handoff from download-complete to indexing visible (an "Indexing…" state on the Medicine card), so a still-indexing install doesn't look broken.

The main new plumbing is teaching the tier installer to handle a DB-ingested resource type, not just ZIMs. Happy to talk that through. Curious why you went integrated vs containerized too, since it lines up well with where we want this to land.

Once it's reworked to this shape we'll pull it onto a test box and run it. Thanks for this.

caweis:
Thanks for the clear direction. Reworking it into opt-in curated content makes sense and we can do that. Here's the concrete plan with a few snippets, plus a couple of things worth settling before we build.

**Integrated vs containerized:** it already is integrated, no container. The feature is an in-app page over a `drug_labels` table plus background download and ingest jobs, run inside the admin app's worker, the same way the ZIM and map ingest already run. So "integrated, admin app is the runtime" is the starting point, not something we change.

**Tier-installer plumbing (the part you offered to talk through):** today `ZimService.downloadCategoryTier` assumes every tier resource is a single ZIM URL and dispatches a `RunDownloadJob` for each. We'd add an optional `type` discriminator to `SpecResource`, defaulting to `'zim'` so every existing manifest entry and the whole ZIM/map path stay untouched:

```ts
// types/collections.ts
export type SpecResource = {
  id: string; version: string; title: string
  description: string; url: string; size_mb: number
  type?: 'zim' | 'dataset'   // defaults to 'zim'
}
```

```ts
// validators/curated_collections.ts  (Vine strips unknown keys, so it has to accept it)
export const specResourceValidator = vine.object({
  // ...existing fields...
  type: vine.enum(['zim', 'dataset']).optional(),
})
```

Then the installer branches on the type. ZIM keeps the current path; a `dataset` resource dispatches our existing download-then-ingest chain:

```ts
// services/zim_service.ts  (inside the existing downloadCategoryTier loop)
for (const resource of toDownload) {
  if (resource.type === 'dataset') {
    // we'd extend dispatch() to carry resource.id/version so an
    // InstalledResource row can be written when ingest finishes
    await DownloadDrugDataJob.dispatch()
    continue
  }
  // unchanged: RunDownloadJob.dispatch({ url: resource.url, filetype: 'zim', ... })
}
```

Install state is one `InstalledResource` row (a new `'dataset'` resource_type) written when ingest finishes, so the tier's installed badge and the existing status math pick it up with no special case. The Medicine/Standard manifest entry is one resource:

```jsonc
// collections/kiwix-categories.json  (medicine-standard tier, resources[])
{
  "id": "openfda-drug-labels",
  "type": "dataset",
  "version": "2026-06",
  "title": "FDA Drug Reference",
  "description": "Offline openFDA drug labels, interaction view, conditions, remedies",
  "url": "https://api.fda.gov/download.json",
  "size_mb": 1700
}
```

One coordination note: the app fetches `kiwix-categories.json` from this repo's `main`, so that entry has to land in your repo for the tier to trigger the dataset. Easiest is to add it on merge, or point the spec URL at the PR branch while you test.

**Multi-part download:** the FDA set is many openFDA partition zips, but the download job already fans those out internally and computes an aggregate percent. We'd model it as one logical dataset resource, feed the existing Active Downloads aggregator as a new `drug-data` filetype showing a single card, and keep the heavy ingest in its own status surface with an "Indexing…" handoff on the card, the way the Knowledge Base separates indexing from downloads. The manifest's `export_date` wires into a daily freshness check that reuses the auto-update apply path. Uninstall truncates the table and clears the on-disk parts, which auto-hides the tile.

A few questions before we start:
1. Tier home: is `medicine-standard` where you want it, given the dataset is ~1.7 GB, or would you rather it have its own tier?
2. Install state: good with a new `'dataset'` value on `InstalledResource.resource_type` (keeps the status math uniform), or keep the drug state separate and special-case it?
3. `export_date` format: do you have a sample of `api.fda.gov/download.json` handy? We want to confirm a plain string compare sorts chronologically before relying on it.

Happy to put up the discriminator plus the installer branch first as a small slice if that's easier to review than the whole rework at once.

V/r, Chris

chriscrosstalk:
Thanks for this, and for the care in the risk framing. Rendering label text verbatim, explicitly not positioning it as an interaction checker, keeping a standing disclaimer, and using CC0 openFDA data downloaded at runtime rather than bundled is exactly the right posture. We want to take this forward, including the condition and remedy cross-reference, because that kind of self-care information is genuinely valuable in an offline scenario where someone may have nothing else to go on.

A few things to get in place before it ships:

1. **Make it optional, off by default.** Right now it registers routes and a table directly into the app. It should be gated behind a Settings feature toggle so users opt in, rather than being present by default. It does not belong in the Supply Depot, since that is for containerized apps and this is an in-app feature, so a settings toggle is the right home for now.
2. **Add the opt-in to the Easy Setup Wizard by folding it into an existing step** where users already choose what offline content to include, rather than adding a dedicated step just for this. New users should be able to turn it on during setup, alongside the other reference libraries.
3. **Clinician review for the affirmative self-care and remedy content, with an unmistakable, persistent disclaimer.** We want to keep this part, but it needs a clinician pass before release, and the "informational only, not medical advice, take it with a grain of salt" framing has to be impossible to miss wherever affirmative guidance appears, not just a one-time banner. In an emergency, real medical care comes first, and the UI should say so.

The verbatim label search and interaction view are already in good shape on the guardrail front. We are keeping this open to work through those items. Really appreciate how thoughtfully you handled the sensitive parts.

chriscrosstalk:
One correction to my note above on where this lives. Rather than a standalone Settings toggle, we want it delivered as **opt-in curated content in the existing Medicine collection** (the `medicine-standard` tier), consistent with the earlier direction on this PR. That is the better fit: it reuses the tier installer, gets content auto-update freshness for free (wire the `api.fda.gov/download.json` `export_date` into the content updater), and surfaces naturally in the setup wizard's content selection alongside the other reference libraries, so no dedicated step is needed. It stays integrated in the admin app (not a Supply Depot Docker app), and installing the tier is the opt-in: no home card or nav entry until it is installed, and uninstalling drops the table and its data.

Everything else stands, and the two requirements from my previous comment are the key additions: the clinician review for the affirmative self-care and remedy content, and the unmistakable, persistent "informational only, not medical advice" disclaimer wherever affirmative guidance appears.

caweis:
Thanks for the thorough read on this, and I'm glad the risk framing came across the way I'd hoped.

Some of what you're describing overlaps with the opt-in-tier rework from the last round, so rather than have you dig through the diff, here's where each piece sits now and one thing worth knowing before you test it.

Installing the `medicine-standard` tier is the opt-in. The tier carries an `openfda-drug-labels` resource marked `type: "dataset"` in `kiwix-categories.json`, and the existing tier installer routes it to the drug download and ingest pipeline, so it stays in the admin app rather than the Supply Depot. The one catch: the app pulls the collection spec from your `main` at runtime, so the tier won't appear until that entry is on your `main`. You can point the spec URL at the PR branch if you want to try it before then.

Nothing shows up until it's installed. Both medical tiles are gated on a server-side flag that only flips once the dataset is present, and uninstall walks it back the way a ZIM delete does. It stops both drug queues, deletes the downloaded files, truncates `drug_labels` (I left the empty schema in place so a reinstall skips the migration), clears the KV markers, and drops the `installed_resources` row. Once that's done the tiles hide themselves.

On the progress side, the download phase reports into the Active Downloads indicator as its own item with a working cancel, and the Medicine card shows an indexing state during the handoff to ingest, the way you laid out earlier.

Freshness now runs inside the content auto-update cycle. The hourly content-update job checks the openFDA `download.json` `export_date` against what's installed and re-runs the download and ingest when a newer one lands, under the same `contentAutoUpdate.*` master switch and window as ZIMs and maps. So if content auto-update is off the dataset holds too, and it refreshes alongside the rest of the content when it's on.

The setup wizard picks it up on its own. Because it's a tier, it lands in the wizard's content selection next to the other libraries with no separate step, subject to the same note about the spec coming from your `main`.

The disclaimer is the part that genuinely needed doing. You were right that one banner at the top doesn't carry far enough once someone is scrolling through remedies. I've put an amber "informational only, not medical advice, and seek real medical care in an emergency" note at the head of each affirmative remedy section, on both Drug Reference and "When to use what", so it stays with the content instead of sitting once at the top (`22eda30`).

That leaves the clinician review. Reading your note again, I take that as a content pass by a clinician before the affirmative material ships, separate from the disclaimer, so I'd rather set it up properly than wave it through. The affirmative entries (18 herbal, about 20 home-care) are all paraphrased from NCCIH, CDC, MedlinePlus, and FDA pages and centralized in two data files with each entry's source URL. I've put together a short review packet from them: every item's text next to its source, with an approve / revise / remove column, so whoever does the pass can move through it quickly. I've also strengthened the on-screen wording beyond the standing disclaimer, so each affirmative section now tells the reader to check with a clinician before using anything, and before combining a remedy with a medication they already take.

A few ways we could land it, your call: you (or we) line up the clinician content-pass before this part ships; or we ship with the affirmative content gated off (or left out) until it's signed off; or, if you're comfortable with it, we ship it behind the disclaimer plus that explicit consult-a-clinician wording as the interim bar, with the formal review to follow. The verbatim label search and the condition-to-OTC matching don't depend on any of it, so none of these hold up the rest.

Thanks again for working through the tricky parts of this with me.

chriscrosstalk:
This is excellent work, and the rework hit every point. Opt-in tier, server-side gating, per-section disclaimers, freshness in the auto-update cycle, clean uninstall. Thank you for the care on the risk framing throughout.

On landing it: let's split by risk. The verbatim FDA label search and the condition-to-OTC matching are grounded in regulated label text, so those ship in the next release as-is. The affirmative herbal and home-care remedy content is the part where we'd be authoring guidance, so we want that gated off by default until a clinician has done the content pass on your review packet. Once it's signed off we flip it on in a follow-up.

Concretely, that means one more thing on your side: an affirmative-content flag that's independent of the tier install, defaulting off, so installing `medicine-standard` lights up label search and condition matching but not the remedy sections. Everything else looks ready.

On our end we'll take the review packet forward for the clinician pass. Really appreciate how you've handled this one.

caweis:
The affirmative-content flag is in (`ba36e46`). It's a `drugReference.remediesEnabled` setting, off by default and separate from the tier install. So installing `medicine-standard` brings up the FDA label search and the condition-to-OTC matching, but the herbal and home-care remedy sections stay hidden. When it's off, the server sends no remedy data on any path that could carry it: the drug-reference page, the condition detail, or the situation search. The "Natural" filter is hidden too. There's no user-facing switch, so it stays off until you flip it after the clinician pass. Label search and condition matching never read the flag, so the regulated part is ready to move whenever you are.

The review packet for the clinician pass is here: https://gist.github.com/caweis/37d6651f2853a4a30e59c8c95fc1adc2. It lists all 38 entries (18 herbal, 20 home-care) with their sources, plus a fillable CSV for the reviewer's verdicts and notes.

chriscrosstalk:
Verified the gate end to end and it's exactly right. `remediesEnabled` is registered as a boolean so the check is sound, it defaults off, and every server path that could emit remedy data (the drug-reference page, the condition detail, and the situation search) withholds it while the gate is closed, so nothing un-reviewed is serialized to the client. Label search and condition matching stay independent of the flag. Nicely done.

We've got the review packet too. The clinician content-pass runs on our side, and we'll flip the flag on in a follow-up once it's signed off.

On our end this is approved for the next release with the remedy sections gated off. Thanks for the real care you put into this one.

caweis:
Appreciate you verifying it end to end, glad it held up to a close read. The clinician pass and flipping the flag afterward sound right, and the packet maps every item to its source to keep that quick. Good to see it landing in the release.

jakeaturner:
@caweis this is an awesome addition, thanks for all the work! One thing I think is worth considering in the future (not a blocker right now) is if we actually change up the Dockerfile to bundle the `{conditions|natural_remedies|home_remedies}.json` files and read/cache directly from the FS at runtime. Again, not a concern right now, but could be a good switch up for next release just for maintainability's sake to avoid the need to keep the `.json` and `.ts` files in sync. I don't foresee those JSON files ever growing large enough to have a meaningful detriment to the image's size (famous last words XD)

caweis:
Thanks Jake, and appreciate you merging it. Fair call on the JSON. Right now the `.ts` is the runtime source of truth because the image only copies `admin/` and ships the compiled `build/`, so the repo-root `collections/*.json` never lands in the container. That's the reason for the split, and the header on those files flags the same tradeoff you're raising. Dropping the duplication is worth doing, so I'll put up a follow-up PR that makes the JSON the single source with nothing to keep in sync. I'll keep the data guaranteed-present at runtime while I'm at it, and lay out the approach in the PR so you can steer it.

PR #1058, opened 2026-07-01
fix: undefined functions in shell scripts + typos

fix2015:
The uninstall and update scripts had some issues from being copy-pasted from the install script:

- \`uninstall_nomad.sh\` references color variables (\`GREEN\`, \`RED\`, etc.) and calls \`header_red\` but none of these were defined — so error messages just silently fail or show "command not found". Added the color definitions and function.
- \`update_nomad.sh\` had the same missing \`header_red\` function, plus the success message still says "installation completed" with a broken \`\${nomad_dir}\` variable reference (should be the hardcoded path).

Also fixed a couple typos:
- "comma-seperated" → "comma-separated" in the AI prompt template (this one goes to the LLM every time)
- "reserve the right accept" → "reserve the right to accept" in CONTRIBUTING.md


chriscrosstalk:
Good catch on all of these. I confirmed it against the live scripts: `header_red` was called but never defined in both the uninstall and update scripts, so they were printing "command not found" instead of the banner, and the update script's success message pointed at an undefined path. We reimplemented the fix in-house (matching the canonical `install_nomad.sh` definitions) in #1098, with credit to you. Closing this in favor of that. Thanks for flagging it.

PR #1059, opened 2026-07-01
fix(downloads): add retry button and resource download link for failed downloads

jarvisxyz:
## Problem

When a download fails, the UI shows only an alert icon, the filename, and a dismiss (X) button. The user has no way to retry the download or navigate to the resource's download page — they have to manually re-add the download and find the source URL elsewhere.

## Solution

Failed downloads now show two action buttons below the error message:
- **Retry** — re-dispatches the original download job (file or model) via a new `POST /api/downloads/jobs/:jobId/retry` endpoint
- **Download page** — an external link to the resource's original download URL (shown only when the download URL is an HTTP(S) URL)

### Changes

| File | Change |
|---|---|
| `admin/app/services/download_service.ts` | Added `retryFailedJob()` — looks up the job in both download queues, removes it, and re-dispatches with the original params |
| `admin/app/controllers/downloads_controller.ts` | Added `retryJob()` controller action |
| `admin/start/routes.ts` | Added `POST /api/downloads/jobs/:jobId/retry` route |
| `admin/inertia/lib/api.ts` | Added `retryDownloadJob()` API client method |
| `admin/inertia/components/ActiveDownloads.tsx` | Failed-state UI now includes Retry button (with loading spinner) and Download page external link |

### Before
Before

### After
After

## Testing

- [ ] Unit test: `retryFailedJob` re-dispatches a file download with correct params
- [ ] Unit test: `retryFailedJob` re-dispatches a model download with correct params
- [ ] Unit test: `retryFailedJob` returns error for unknown job ID
- [ ] E2E: failed download card shows Retry and Download page buttons
- [ ] E2E: clicking Retry re-dispatches the job and card transitions to downloading state
- [ ] E2E: Download page link opens original URL in new tab
- [ ] E2E: Download page link hidden when URL is not HTTP(S)


chriscrosstalk:
This is a solid fill for a real dead-end: right now a failed download just sits there with no way forward. The Retry-plus-Download-page treatment is the right UX, and the full-stack slice is clean (service, controller, route, api client, component all lined up). Good call gating the Download page link to HTTP(S) URLs, and thanks for the before/after shots.

A couple of notes for whoever picks it up on our side: the retry re-dispatch removes and re-queues the job, so it's worth confirming it doesn't fight the existing BullMQ retry/attempts handling on model downloads. And the PR's test checklist is still unchecked, so it'll get a browser test across the failed-download states before merge.

jakeaturner (approved):
One small belt-and-suspenders improvement here would be to confirm inside `retryFailedJob` that the original job is in fact in a failed state, but in practice that would require an end user manually making an API call to the retry endpoint, with an existing jobId not through the UI. Not a blocker here, just something worth following-up later. Great addition, thanks for the work here!

PR #1060, opened 2026-07-01
feat: Expandable rows in Kiwix Library browser

jarvisxyz:
## Summary

The "Browse the Kiwix Library" table in Content Explorer previously showed each ZIM file's description truncated to a single line, making it difficult to assess content before downloading. This PR adds click-to-expand functionality so each row reveals the full description and additional metadata.

## Changes

- **`admin/types/zim.ts`** — Extended `RemoteZimFileEntry` with optional metadata fields: `language`, `publisher`, `category`, `tags`, `article_count`, `media_count`, `issued`
- **`admin/app/services/zim_service.ts`** — Updated `listRemote()` to map these additional fields from the raw Kiwix API response (previously discarded)
- **`admin/inertia/pages/settings/zim/remote-explorer.tsx`** — Removed `@tanstack/react-virtual` virtualization (12 items/page — not needed and incompatible with variable-height expanded rows) and added `expandable` prop to `StyledTable` with a rich `expandedRowRender` showing full description, metadata grid, tags, and file name

## Before / After

**Before** — descriptions truncated to a single line, no expandable detail:

Before

**After** — click a row to expand and see full description + metadata:

After

## Test plan

- [ ] Navigate to Settings → Content Explorer → Browse the Kiwix Library
- [ ] Verify rows render as before (collapsed state)
- [ ] Click a row — verify it expands to show full description and metadata
- [ ] Click again — verify it collapses
- [ ] Verify multiple rows can be expanded simultaneously
- [ ] Verify download button still works from both collapsed and expanded states
- [ ] Verify infinite scroll still loads more items

### DIFF
--- a/admin/app/services/zim_service.ts
+++ b/admin/app/services/zim_service.ts
@@ -164,6 +164,13 @@ export class ZimService {
           download_url,
           author: entry.author.name,
           file_name,
+          language: entry.language,
+          category: entry.category,
+          tags: entry.tags,
+          article_count: entry.articleCount,
+          media_count: entry.mediaCount,
+          publisher: entry.publisher?.name,
+          issued: entry['dc:issued'],
         })
       }
 
--- a/admin/inertia/pages/settings/zim/remote-explorer.tsx
+++ b/admin/inertia/pages/settings/zim/remote-explorer.tsx
@@ -7,7 +7,6 @@ import {
 } from '@tanstack/react-query'
 import api from '~/lib/api'
 import { useCallback, useEffect, useMemo, useRef, useState } from 'react'
-import { useVirtualizer } from '@tanstack/react-virtual'
 import StyledTable from '~/components/StyledTable'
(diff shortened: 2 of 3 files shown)


chriscrosstalk:
Good improvement. Being able to expand a row and read the full description plus metadata before committing to a multi-GB download is exactly the kind of "know before you pull" the Content Explorer needed, and surfacing the Kiwix fields we were already fetching but discarding is a clean way to get there. Reusing `StyledTable`'s expandable support keeps it consistent with the rest of the tables.

One thing for our side to check before merge: the PR removes the react-virtual virtualization, but the browse list also grows via infinite scroll, so we'll want to confirm long libraries still render smoothly without it. It'll get a browser test on the Content Explorer as part of that.

jakeaturner (approved):
Great addition! Made one small fix to guard against bad dates breaking the page rendering. Thanks for the contribution!

PR #1063, opened 2026-07-03
feat(rag): add subject/collection organization to knowledge base

### LINKED ISSUE #1062: Add subject/collection organization to the Knowledge Base (RAG)
**Problem**

The Knowledge Base currently stores everything in a single flat Qdrant collection (`nomad_knowledge_base`). For users with a large, topically diverse KB (survival guides, medical references, electronics docs, farming manuals, etc.), mixing unrelated subjects together increases retrieval noise and the chance of the AI Assistant pulling in irrelevant context, which can worsen hallucination on multi-topic instances.

**Proposed solution**

Rather than splitting into separate physical Qdrant collections (more operational overhead, migrations, etc.), tag each file/chunk with a `collection` field in its Qdrant payload (indexed for fast filtering) plus a matching column on `KbIngestState`. This is backward compatible — untagged content is simply null/uncategorized.

Changes:
- A small fixed set of categories (food, water, shelter, electricity, survival, farming, computer, medical, boating) defined in `admin/constants/kb_collections.ts` — easy to extend later, open to feedback on making this user-configurable instead of fixed
- Upload-time category dropdown in the Knowledge Base modal
- Per-file collection reassignment (move files between categories after upload) via a new `/api/rag/update-collection` endpoint
- A "Search in" collection filter on the chat page, scoping `searchSimilarDocuments` to a single collection when set
- A `GET /api/rag/collections` endpoint (Qdrant facet query) to populate filter dropdowns with only collections actually in use

**Bonus fix included in the same branch**

While working on this I found `.docx` files were being classified as plain text and run through raw-text extraction — since .docx is a ZIP/XML format, this produced garbage content in the KB. Added a proper mammoth-based extractor and a dedicated `docx` case in the file-type switch.

**Status**

Implemented and running on a personal instance (Kubuntu 26.04, Docker, RTX 5060 Ti). Existing test suite updated and passing. Happy to adjust category list, filtering approach, or split this into smaller PRs per feedback before a full review.

just-jbc:
Closes #1062

## Summary

Adds subject/collection organization to the Knowledge Base (see linked issue for full rationale), plus a `.docx` extraction fix found along the way.

## Changes

- `admin/constants/kb_collections.ts`: fixed category list (food, water, shelter, electricity, survival, farming, computer, medical, boating)
- New migration adding a nullable `collection` column to `kb_ingest_state`
- `RagService`: collection propagated through embedding, stored as an indexed Qdrant payload field, and used as an optional filter in `searchSimilarDocuments`; new `getKnowledgeCollections()` (Qdrant facet) and `updateFileCollection()` methods
- `RagController`: `GET /api/rag/collections`, `POST /api/rag/update-collection`, and `collection` accepted on upload
- `OllamaController`: `collection` accepted on `/api/ollama/chat` to scope RAG search
- Knowledge Base modal: upload-time category dropdown, per-file collection reassignment, "Search in" filter on the Stored Files table
- Chat page: "Search in" collection filter next to the model selector
- `.docx` fix: added a dedicated `docx` file-type case using `mammoth` for real text extraction, instead of falling through to raw-text extraction

## Testing

- Ran against a live instance (Kubuntu 26.04, Docker, RTX 5060 Ti, 44 ZIM files + uploaded docs)
- Verified upload with category selection, per-file reassignment, and collection-filtered chat search all work end-to-end
- Updated `tests/unit/kb_file_grouping.spec.ts` fixtures for the new required field; existing test suite passes
- Verified `.docx` upload now extracts clean text instead of raw XML

## Notes

Open to feedback on the fixed category list (could be made user-configurable) or splitting this into smaller PRs if preferred.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,5 @@ admin/public/assets
 
 # Admin specific development files
 admin/storage
+nomad_admin_config.json
+nomad_admin_hostconfig.json
--- a/.gitignore
+++ b/.gitignore
@@ -41,5 +41,3 @@ admin/public/assets
 
 # Admin specific development files
 admin/storage
-nomad_admin_config.json
-nomad_admin_hostconfig.json
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -77,10 +77,12 @@ export default class OllamaController {
 
       logger.debug(`[OllamaController] Rewritten query for RAG: "${rewrittenQuery}"`)
       if (rewrittenQuery) {
+        const collectionFilter: string | null = request.input('collection', null)
         const relevantDocs = await this.ragService.searchSimilarDocuments(
           rewrittenQuery,
           5, // Top 5 most relevant chunks
-          0.3 // Minimum similarity score of 0.3
+          0.3, // Minimum similarity score of 0.3
+          collectionFilter ?? undefined
         )
 
         logger.debug(`[RAG] Retrieved ${relevantDocs.length} relevant documents for query: "${rewrittenQuery}"`)
@@ -456,3 +458,4 @@ export default class OllamaController {
     }
   }
 }
+
--- a/admin/app/controllers/rag_controller.ts
+++ b/admin/app/controllers/rag_controller.ts
@@ -94,6 +94,35 @@ export default class RagController {
     return response.status(200).json({ message: result.message })
   }
 
+  public async renameKnowledgeCollection({ request, response }: HttpContext) {
+    const oldName: string | null = request.input('oldName', null)
+    const newName: string | null = request.input('newName', null)
+
+    if (!oldName || !newName) {
+      return response.status(400).json({ error: 'oldName and newName are required.' })
+    }
+
+    const result = await this.ragService.renameKnowledgeCollection(oldName, newName)
+    if (!result.success) {
+      return response.status(500).json({ error: result.message })
+    }
+    return response.status(200).json({ message: result.message })
+  }
+
+  public async deleteKnowledgeCollection({ request, response }: HttpContext) {
+    const name: string | null = request.input('name', null)
+
+    if (!name) {
+      return response.status(400).json({ error: 'name is required.' })
+    }
+
+    const result = await this.ragService.deleteKnowledgeCollection(name)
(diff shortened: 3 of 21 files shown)


just-jbc:
Pushed an update — collections can now be created on the fly (typing a new name when assigning a file), renamed, and removed via a new "Manage Collections" screen. Removing a collection reassigns its files back to Uncategorized rather than deleting anything.

chriscrosstalk:
This is a genuinely useful feature and the implementation is solid, so we'd like to take it forward with one change to the tagging model. Instead of a fixed list *or* fully free-form, we want both: keep a curated set of preset tags as starting points, and also let users create their own tags on top of them. Presets give people a sensible default to reach for, and user-created tags mean nobody hits a dead end when their content doesn't fit the presets.

Two parts to that:

1. **Broaden the presets.** The current list leans survival-specific (food, water, shelter, boating). Since NOMAD's Knowledge Base is general-purpose, the starter set should cover more ground so it's useful to home-lab, reference, and everyday users too, not just prepping.
2. **Make the tag control creatable.** Your data model already supports this: collection is stored as a free-form payload string and `getKnowledgeCollections()` already returns the tags actually in use. So it's mainly swapping the upload dropdown for a creatable combobox (pick a preset, pick an existing user tag, or type a new one, autocompleted from that facet endpoint) and adding input sanitization now that tags are user-entered (trim, normalize case, length cap, dedupe).

One more thing: the `.docx` extraction fix is a good catch but separable from this feature. Splitting it into its own PR would let it merge right away instead of waiting on the collections work. Happy to review that one quickly.

just-jbc:
Split the .docx extraction fix out into its own PR (#1100) so it can merge independently — this branch is now purely the collections work. Also broadened the preset tags and swapped the plain dropdown for a creatable combobox with server-side sanitization per your feedback above.

just-jbc:
Cleaned up some unrelated diff noise before this goes further: regenerated package-lock.json/package.json (an old mammoth reference and npm-version churn had snuck in), restored a few pre-existing comments that got accidentally stripped, and fixed an indentation bug in kb_ingest_state.ts. Diff should now be scoped to just the collections feature.

jakeaturner (approved):
Great addition, thanks @just-jbc !

PR #1078, opened 2026-07-10
fix(AI): stream thinking from /v1 reasoning field + abort on client disconnect (#1065)

### LINKED ISSUE #1065: [Bug]: AI chat streams empty deltas indefinitely with remote Ollama 0.31.x (thinking models): /v1 emits `reasoning` field and ignores `think` param
### Issue Category

AI Assistant (Ollama)

### Bug Description

When NOMAD is configured with a **remote Ollama** endpoint (Settings → remote Ollama URL)
running a recent Ollama build (tested 0.31.1), chatting with any thinking-capable model
(e.g. `qwen3:8b`, `qwen3.5:4b`) makes the chat UI hang forever: the SSE stream emits an
endless sequence of chunks where both `content` and `thinking` are empty strings, and
`done: true` never arrives.

There are two independent defects in `app/services/ollama_service.ts`, both in how the
OpenAI-compatible endpoint (`/v1/chat/completions`) is used:

1. **Thinking field name mismatch.** `chatStream()`'s normalizer reads
   `delta?.thinking`, but current Ollama emits thinking tokens on the OpenAI-compat
   endpoint as **`delta.reasoning`** (non-stream: `message.reasoning`). All thinking
   output is silently dropped, so the entire thinking phase surfaces as empty chunks.
   The non-stream `chat()` has the same mismatch (`choice.message.thinking`).

2. **The `think` request param is silently ignored by `/v1`.** The controller forces
   `think: true` for every thinking-capable model, and `chatStream()` forwards `think`
   as a top-level param to `chat.completions.create()`. Ollama's OpenAI-compat endpoint
   ignores this non-standard param entirely — verified with direct curl: both
   `"think": true` and `"think": false` produce identical behavior (thinking enabled,
   emitted as `reasoning`). The only control the `/v1` endpoint honors is
   **`reasoning_effort: "none"`** (verified: content streams immediately, no reasoning
   deltas).

Combined effect: thinking is always on, invisible to the user, and unbounded (no token
cap). Models like qwen3.5:4b routinely "think" for several thousand tokens before a
two-word answer, so the UI looks frozen. Worse, with Ollama's default
`OLLAMA_NUM_PARALLEL=1`, the runaway generation keeps decoding server-side even after
the client disconnects, occupying the single slot — every subsequent chat/RAG request
queues behind it and times out, so the whole AI assistant appears dead until
`ollama stop <model>` is run.

My environment is an (unofficial) macOS host, but both defects are in the Node service
code and are OS-independent: any officially supported Debian/Ubuntu install using the
built-in remote-Ollama feature against Ollama ≥ 0.31 reproduces this.

### Steps to Reproduce

1. Install Ollama 0.31.x natively on any host and pull a thinking-capable model:
   `ollama pull qwen3.5:4b`
2. In NOMAD: Settings → AI → configure remote Ollama endpoint pointing at that host
   (`/api/ollama/configure-remote`); status shows configured+connected.
3. Open the chat UI (or curl the API directly) and send any short prompt:

   curl -N -X POST http://<nomad>:8080/api/ollama/chat \
     -H 'Content-Type: application/json' \
     -d '{"model":"qwen3.5:4b","messages":[{"role":"user","content":"Reply with exactly two words: NOMAD works"}],"stream":true}'

4. Observe an endless stream of `data: {"message":{"content":"","thinking":""},"done":false}`.
5. Independently verify the two root causes against Ollama directly:

   # (a) thinking arrives as `reasoning` on the OpenAI-compat endpoint:
   curl -N http://<ollama>:11434/v1/chat/completions -d '{"model":"qwen3.5:4b","messages":[{"role":"user","content":"hi"}],"stream":true,"max_tokens":30}'
   # → deltas carry {"content":"","reasoning":"..."}

   # (b) `think:false` is ignored, `reasoning_effort:"none"` works:
   curl -N http://<ollama>:11434/v1/chat/completions -d '{"model":"qwen3.5:4b","messages":[{"role":"user","content":"hi"}],"stream":true,"think":false,"max_tokens":30}'
   # → still reasoning deltas
   curl -N http://<ollama>:11434/v1/chat/completions -d '{"model":"qwen3.5:4b","messages":[{"role":"user","content":"hi"}],"stream":true,"reasoning_effort":"none","max_tokens":30}'
   # → content streams immediately, no reasoning

### Expected Behavior

Chat responses stream normally; thinking output (if enabled) is visible in the UI as it
streams; a short prompt completes in seconds.

### Actual Behavior

SSE stream emits only empty chunks (`content: ""`, `thinking: ""`) for minutes;
`done: true` never arrives within any reasonable timeout. The abandoned generation keeps
running server-side and blocks Ollama's single parallel slot, so all subsequent AI
requests (chat, RAG query rewriting) hang too. `ollama ps` shows the model busy; the
llama-server log shows thousands of tokens decoded with no visible output in NOMAD.

### N.O.M.A.D. Version

1.33.0 (ghcr.io/crosstalk-solutions/project-nomad:latest, pulled 2026-07-03)

### Operating System

Other (not yet officially supported)

### Docker Version

Docker Desktop for Mac, engine 29.6.1, Compose v5.3.0

No

### GPU Model (if applicable)

Apple M2 (integrated; Ollama runs natively on the host with Metal, outside Docker)

### System Specifications

CPU: Apple M2 (MacBook Air)
RAM: 16 GB
Available Disk Space: ~120 GB
GPU: Apple M2 integrated (used by native Ollama via Metal)
Ollama: 0.31.1, native install, reached from containers via host.docker.internal:11434
Note: management stack runs in Docker Desktop (linux/amd64 images under Rosetta);
the bug is in the admin app's JS and does not depend on the host OS.

All management containers healthy (nomad_admin, nomad_mysql, nomad_redis, nomad_dozzle,
nomad_updater), plus nomad_kiwix_server and nomad_qdrant running. Remote Ollama status:
{"configured":true,"connected":true}. Model list, RAG embedding (nomic-embed-text via
the same remote Ollama), and everything else works — only chat is affected.

### Relevant Logs

```shell
# What NOMAD's /api/ollama/chat SSE stream emits, indefinitely (4800+ chunks over 120s):
data: {"message":{"content":"","thinking":""},"done":false}
data: {"message":{"content":"","thinking":""},"done":false}
...

# What Ollama 0.31.1 actually sends on /v1/chat/completions for the same request:
data: {"id":"chatcmpl-785","object":"chat.completion.chunk","model":"qwen3.5:4b",
  "choices":[{"index":0,"delta":{"role":"assistant","content":"","reasoning":"Thinking"},"finish_reason":null}]}

# llama-server (remote Ollama) during the "hang" — generation is running the whole time:
slot print_timing: id  0 | task 7945 | n_decoded =   2993, tg = 16.34 t/s
time=... level=INFO source=llama_server.go:1450 msg="aborting completion request due to client closing the connection"
```

```javascript

```

### Screenshots

_No response_

### Additional Context

Suggested fix (verified locally by patching the built JS inside the container):

1. In `chatStream()`'s normalizer and in `chat()`, read both field names:
   `delta?.thinking ?? delta?.reasoning ?? ''` /
   `choice.message.thinking ?? choice.message.reasoning ?? undefined`.
   This alone makes thinking stream visibly again.

2. When thinking should be off, send `reasoning_effort: "none"` instead of omitting
   `think` (which `/v1` ignores). With both patches applied, a short prompt completes
   in ~9s including model load.

3. Consider exposing a user-facing toggle for thinking instead of forcing
   `think: true` for every thinking-capable model — unbounded hidden thinking plus
   `OLLAMA_NUM_PARALLEL=1` turns one slow reply into an apparent total AI outage.

Happy to provide more logs or test a fix branch.

### Pre-submission Checklist

chriscrosstalk:
## Problem

With a thinking-capable model (e.g. `qwen3`), chat hangs forever: the SSE stream emits nothing but empty `content`/`thinking` chunks and never reaches `done`. Reported in #1065 against a remote Ollama 0.31.x, but the cause is in NOMAD's Node service and affects the OpenAI-compat (`/v1`) path used for **both local and remote** Ollama.

Two defects:

1. **Field mismatch.** `chatStream()` and `chat()` read `delta.thinking` / `message.thinking`, but Ollama's `/v1` emits thinking tokens as **`reasoning`**. All thinking output was silently dropped.
2. **No abort on client disconnect.** When the user closed the chat, the upstream generation kept decoding server-side. With Ollama's default `OLLAMA_NUM_PARALLEL=1`, that abandoned request held the only slot, so every later chat/RAG request queued behind it and the assistant appeared dead until the model was manually stopped.

## Fix

1. Read `thinking ?? reasoning` in both the streaming and non-streaming paths (the inline `<think>`-tag parser for other backends is untouched).
2. Wire an `AbortController` to the response `close` event in the controller and thread the signal into the OpenAI SDK request, so a client disconnect aborts the upstream generation.

## Testing

Verified end-to-end on NOMAD2 with `qwen3:0.6b` (reports the `thinking` capability; emits `reasoning` on `/v1`, so the bug reproduces even on 0.30.10):

**Field fix**
- Before: `data: {"message":{"content":"","thinking":""},"done":false}` endlessly, never done.
- After: thinking streams visibly (`"thinking":"Okay"`, `" let"`, `"'s"` ...) and the stream reaches `"done":true`.

**Abort on disconnect**
- Started a long generation, disconnected the client at 3s.
- Ollama's `n_decoded` froze (100 → 100 over the next 4s) and the server logged `srv stop: cancel task` / `slot release`.
- A subsequent request completed normally, confirming the slot was freed.

## Follow-up

Thinking is still force-on for capable models here. A user-facing **per-model thinking toggle (default off)**, shown in the chat window only for models that report the `thinking` capability, is a planned follow-up PR.

Closes #1065

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -164,14 +164,32 @@ export default class OllamaController {
 
       if (reqData.stream) {
         logger.debug(`[OllamaController] Initiating streaming response for model: "${reqData.model}" with think: ${think}`)
-        // Headers already flushed above
-        const stream = await this.ollamaService.chatStream({ ...ollamaRequest, think, numCtx })
+        // Headers already flushed above.
+        // Abort the upstream generation if the client disconnects — otherwise an abandoned
+        // request keeps decoding server-side and, with Ollama's default OLLAMA_NUM_PARALLEL=1,
+        // blocks every later chat/RAG request until the model is manually stopped (#1065).
+        const abortController = new AbortController()
+        response.response.on('close', () => abortController.abort())
+        const stream = await this.ollamaService.chatStream({
+          ...ollamaRequest,
+          think,
+          numCtx,
+          signal: abortController.signal,
+        })
         let fullContent = ''
-        for await (const chunk of stream) {
-          if (chunk.message?.content) {
-            fullContent += chunk.message.content
+        try {
+          for await (const chunk of stream) {
+            if (chunk.message?.content) {
+              fullContent += chunk.message.content
+            }
+            response.response.write(`data: ${JSON.stringify(chunk)}\n\n`)
           }
-          response.response.write(`data: ${JSON.stringify(chunk)}\n\n`)
+        } catch (err) {
+          if (abortController.signal.aborted) {
+            logger.debug('[OllamaController] Client disconnected; aborted upstream Ollama generation')
+            return
+          }
+          throw err
         }
         response.response.end()
 
--- a/admin/app/services/ollama_service.ts
+++ b/admin/app/services/ollama_service.ts
@@ -45,6 +45,9 @@ type ChatInput = {
   think?: boolean | 'medium'
   stream?: boolean
   numCtx?: number
+  // Aborts the upstream request when the client disconnects, so an abandoned generation
+  // doesn't keep decoding server-side and block Ollama's single parallel slot (#1065).
+  signal?: AbortSignal
 }
 
 @inject()
@@ -333,13 +336,15 @@ export class OllamaService {
       params.num_ctx = chatRequest.numCtx
     }
 
-    const response = await this.openai.chat.completions.create(params)
+    const response = await this.openai.chat.completions.create(params, { signal: chatRequest.signal })
     const choice = response.choices[0]
 
     return {
       message: {
         content: choice.message.content ?? '',
-        thinking: (choice.message as any).thinking ?? undefined,
+        // Ollama's OpenAI-compat endpoint (/v1) emits thinking as `reasoning`; its native
+        // shape uses `thinking`. Read both so thinking is never silently dropped (#1065).
+        thinking: (choice.message as any).thinking ?? (choice.message as any).reasoning ?? undefined,
       },
       done: true,
       model: response.model,
@@ -364,7 +369,9 @@ export class OllamaService {
       params.num_ctx = chatRequest.numCtx
     }
 
-    const stream = (await this.openai.chat.completions.create(params)) as unknown as Stream<ChatCompletionChunk>
+    const stream = (await this.openai.chat.completions.create(params, {
+      signal: chatRequest.signal,
+    })) as unknown as Stream<ChatCompletionChunk>
 
     // Returns how many trailing chars of `text` could be the start of `tag`
     function partialTagSuffix(tag: string, text: string): number {
@@ -383,7 +390,8 @@ export class OllamaService {
 
       for await (const chunk of stream) {
         const delta = chunk.choices[0]?.delta
-        const nativeThinking: string = (delta as any)?.thinking ?? ''
+        // /v1 emits thinking as `reasoning`; native Ollama uses `thinking`. Read both (#1065).
+        const nativeThinking: string = (delta as any)?.thinking ?? (delta as any)?.reasoning ?? ''
         const rawContent: string = delta?.content ?? ''
 
         // Parse <think> tags out of the content stream


serratusCZ:
Tested `fix/remote-ollama-thinking-stream` on my original repro environment: unofficial macOS (Apple Silicon, Docker Desktop w/ Rosetta), remote Ollama 0.32.0, `qwen3.5:4b` via `host.docker.internal`.

## Defect 1 (field mismatch) — confirmed fixed

**Before** (`:latest`), every chunk was empty for the full 60s window:
data: {"message":{"content":"","thinking":""},"done":false}
data: {"message":{"content":"","thinking":""},"done":false}

**After** (this branch), `thinking` streams real content:
data: {"message":{"content":"","thinking":"Thinking"},"done":false}
data: {"message":{"content":"","thinking":" Process"},"done":false}
data: {"message":{"content":"","thinking":":"},"done":false}
data: {"message":{"content":"","thinking":"\n\n1"},"done":false}
data: {"message":{"content":"","thinking":".  **"},"done":false}
data: {"message":{"content":"","thinking":"An"},"done":false}
data: {"message":{"content":"","thinking":"alyze"},"done":false}
data: {"message":{"content":"","thinking":" the"},"done":false}

> Note: my 60s test window wasn't long enough to see `content` / `done:true` for this prompt — `qwen3.5:4b` was still deep in an unbounded thinking phase at cutoff. Consistent with thinking still being forced-on pending #1079, not a regression from this PR.

## Defect 2 (abort on disconnect) — not independently confirmed

I only checked `ollama ps` after disconnecting, which just confirms the model stayed loaded — it doesn't show whether the abandoned generation actually stopped decoding. Before/after outputs were identical by this measure, so I can't add evidence beyond what's already in the PR description (decode-counter freeze + `cancel task` in the server log). Happy to re-test with better instrumentation if useful.

---

Environment note for the record: this all runs through `host.docker.internal` from a Docker container to a native macOS Ollama process — that path was untested in the original verification (done on "NOMAD2"), so it's good confirmation defect 1's fix holds across that boundary too.

jakeaturner (dismissed):

jakeaturner (approved):

PR #1091, opened 2026-07-13
fix(content): resolve current ZIM URL before download

### LINKED ISSUE #1045: Resolve ZIM download URLs from the live Kiwix catalog instead of static dated URLs
## Problem

Fresh ZIM downloads fail with `404` when Kiwix rotates a dated filename. The Content Explorer / curated download list is populated from static catalog JSON (`collections/wikipedia.json`, `collections/kiwix-categories.json`) fetched from `main` via `CollectionManifestService`. Each entry hardcodes a dated URL (e.g. `wikipedia_en_all_mini_2025-12.zim`). Kiwix removes old dated files when it publishes new ones, so any entry whose date has rotated 404s on download until a human edits the JSON.

Surfaced in #1015: `wikipedia_en_all_mini_2025-12.zim` was already a hard 404. Patched manually by refreshing the URLs (PR #1046), but that recurs at every rotation.

## Root cause

Two systems, only one is self-healing:

- **Installed-content auto-update (#996)** already queries the **live** Kiwix OPDS catalog via `KiwixCatalogService` (`browse.library.kiwix.org/catalog/v2/entries`) to find newer versions. Self-healing. Good.
- **The browse/fresh-download path** trusts the static dated `resource.url` from the committed JSON. Not self-healing. Goes stale and 404s.

So we already have live-catalog resolution in the codebase; it just isn't wired into the download path.

## Proposed fix

At download time, resolve the current URL for a catalog entry via `KiwixCatalogService.getLatestZim(resourceId)` (the same lookup the updater uses) rather than using the committed dated URL directly. The static JSON keeps the resource id, display name, and approximate size; the actual URL is resolved live, so a stale committed date self-heals to whatever Kiwix currently serves. Fall back to the static URL if the live lookup fails (offline / catalog unreachable).

## Scope

- `collections/wikipedia.json` and `collections/kiwix-categories.json` (medicine and other ZIM categories carry the same static-dated-URL risk).
- `maps.json` is a separate mechanism (project-nomad-maps repo) and out of scope here.

## Stopgap

Manual URL refresh in PR #1046 unblocks users now; this issue is the durable fix.

NgoQuocViet2001:
## Summary
- Resolve curated ZIM downloads against the live Kiwix catalog before dispatch.
- Keep static manifest values as a fallback when no current catalog entry is available.
- Preserve newer manifest metadata if the catalog temporarily lags.
- Compare `YYYY-M` / `YYYY-MM` versions numerically so October is newer than February.
- Add unit coverage for live, fallback, missing-size, older-catalog, and non-padded-month cases.

## Problem
Curated manifests pin dated ZIM URLs. Kiwix rotates filenames and removes old files, so fresh downloads can return 404 until the JSON is manually refreshed.

## Solution
Reuse the existing bounded `KiwixCatalogService.getLatestForResources()` lookup used by the updater, then dispatch with the resolved URL, version, and size. Per-resource catalog failures produce no map entry, so the static manifest remains the fallback.

Version comparison parses the catalog's year/month values numerically, avoiding lexicographic ordering errors such as `2026-10 < 2026-2`.

Closes #1045

## Verification
- `npx tsx --test tests/unit/zim_download_resolution.spec.ts` — 5/5 passed
- `npm run typecheck`
- `npm run build`
- `npx prettier --check app/utils/zim_download_resolution.ts tests/unit/zim_download_resolution.spec.ts`
- `npx eslint app/utils/zim_download_resolution.ts tests/unit/zim_download_resolution.spec.ts`
- `git diff --check`

## Local environment note
A full Adonis test boot on Windows is blocked by `@openzim/libzim`, which has no Windows native binding. The focused unit suite runs directly without booting the native service; the repository's Linux CI remains authoritative for the native suite.

### DIFF
--- a/admin/app/services/zim_service.ts
+++ b/admin/app/services/zim_service.ts
@@ -29,10 +29,12 @@ import CollectionManifest from '#models/collection_manifest'
 import { RunDownloadJob } from '#jobs/run_download_job'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
 import { CollectionManifestService } from './collection_manifest_service.js'
+import { KiwixCatalogService } from './kiwix_catalog_service.js'
 import { KiwixLibraryService } from './kiwix_library_service.js'
 import type { CategoryWithStatus } from '../../types/collections.js'
 import CustomLibrarySource from '#models/custom_library_source'
 import { assertNotPrivateUrl } from '#validators/common'
+import { resolveZimDownload } from '../utils/zim_download_resolution.js'
 
 const ZIM_MIME_TYPES = ['application/x-zim', 'application/x-openzim', 'application/octet-stream']
 const WIKIPEDIA_OPTIONS_URL = 'https://raw.githubusercontent.com/Crosstalk-Solutions/project-nomad/refs/heads/main/collections/wikipedia.json'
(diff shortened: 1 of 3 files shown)


chriscrosstalk:
Reviewed this against the current `KiwixCatalogService` and it's the right approach: reusing the bounded `getLatestForResources()` lookup, resolving at dispatch, and falling back to the static manifest per-resource when the catalog lags or the box is offline. The version compare matches the `YYYY-MM` convention `pickNewestZim` already uses, the offline fallback path is correct, and the unit coverage hits the cases that matter. This is exactly what #1045 was asking for. Nice work, and thanks for the tests.

NgoQuocViet2001:
@chriscrosstalk thanks for the detailed review — appreciate it!

jakeaturner (changes requested):
@NgoQuocViet2001 this is a great addition, thank you! Only one small caveat that will need to be addressed before merge: because the `version` field is a string, plain comparison will break in some cases, for example, `'2026-10' < '2026-2'`. We'll need to either attempt to parse the version to a date object first, or break it up into something like a tuple and parse as ints (e.g. `[2026, 10]`) so we can avoid those edge-cases.

NgoQuocViet2001:
@jakeaturner Thanks for catching this. Fixed in c69f4c0: ZIM year/month versions are now parsed and compared numerically, so `2026-10` correctly wins over `2026-2`. I also added the exact regression case; the focused suite passes 5/5, along with typecheck, lint, formatting, and the production build.

jakeaturner (approved):
Awesome!

PR #1100, opened 2026-07-14
fix(rag): add proper .docx text extraction via mammoth

just-jbc:
## Summary

`.docx` files are currently classified as plain text (`determineFileType` in `admin/app/utils/fs.ts`) and routed through raw-text extraction. Since `.docx` is a ZIP archive containing XML, this produces garbage content in the Knowledge Base — XML tags and binary noise instead of the actual document text.

## Changes

- Split `docx` out of the generic `text` file type into its own case
- Added a `mammoth`-based extractor (`processDocxFile`) that parses the document XML properly and returns clean plain text

## Testing

Verified `.docx` uploads now extract readable text instead of raw XML. Existing test suite passes; build is clean.

## Note

Split out of #1063 per review feedback there, so this can merge independently.

### DIFF
--- a/admin/app/services/rag_service.ts
+++ b/admin/app/services/rag_service.ts
@@ -10,6 +10,7 @@ import { createWorker } from 'tesseract.js'
 import { fromBuffer } from 'pdf2pic'
 import JSZip from 'jszip'
 import * as cheerio from 'cheerio'
+import mammoth from 'mammoth'
 import { OllamaService } from './ollama_service.js'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
 import { removeStopwords } from 'stopword'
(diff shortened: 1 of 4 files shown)


chriscrosstalk:
Thanks for splitting this out cleanly. The approach is right: `.docx` is a ZIP/XML format, so routing it out of the plain-text path into a mammoth extractor is the correct fix, and the new type is handled consistently with the existing PDF/EPUB extractors.

One request before merge: the `package-lock.json` in this PR carries a lot of unrelated churn (removed `libc` fields and added `dev` flags across native optional packages). That's from an npm version mismatch rather than anything intentional. Could you regenerate the lockfile with the npm version pinned in the repo so the diff only shows the `mammoth` additions? That keeps the dependency change auditable.

Once that's tidied, our dev team will give it a build/upload test and get it merged.

just-jbc:
Regenerated the lockfile — diff is now just the mammoth additions (12 packages) plus the one dependency entry in package.json. No unrelated churn.

jakeaturner (approved):
Another great add, thanks!

Issue #1030, opened 2026-06-22
bug(maps): world basemap is fetched on-demand from the internet — blank map offline on installs that never cached it


chriscrosstalk:
### Summary
On a NOMAD that has never extracted the world basemap while online, `/maps` renders blank/grey when offline. This defeats offline-first for any install whose regional extracts don't cover the current view. Reproduced in the field on the NOMAD Deck (portable build) during an offline demo.

### Root cause
The base map only renders if `world.pmtiles` already exists on disk:

- `generateSourcesArray()` only emits the world source when `worldBasemapReady === true` (`map_service.ts:513`). With no `world.pmtiles` and no regional extract covering the view, the style has zero sources → blank map.
- `world.pmtiles` is produced lazily by `_setupWorldBasemap()` (`map_service.ts:425`), which runs `pmtiles extract` against a **remote** URL from `getGlobalMapInfo()` → `build-metadata.protomaps.dev` + `build.protomaps.com` (`map_service.ts:50-51, 599-616`). Internet-only.
- The base-asset bundle (`downloadBaseAssets()`) ships styles/sprites/fonts but **explicitly excludes** `world.pmtiles` (`map_service.ts:89`).

Net: the basemap is silently downloaded on first online `/maps` visit. Installs that go straight offline never get it.

### Repro
1. Fresh install (or any box without `storage/maps/pmtiles/world.pmtiles`).
2. No internet.
3. Open `/maps` → blank/grey map.

(Online, the same first visit fetches ~15 MB and the map appears, which masks the bug.)

### Impact
Core offline-first promise fails on portable/field units — the exact scenario the product is built for.

### Proposed fix
Make the world basemap present after install instead of lazily fetched:
- **Preferred:** include `world.pmtiles` in the base-assets tarball so `downloadBaseAssets()` provisions it at install time (when internet is expected), and drop the `!== WORLD_BASEMAP_FILENAME` exclusion.
- Alternative: run `_setupWorldBasemap()` during install/easy-setup with a clear progress step, so the one network dependency happens before the unit ever goes offline.
- Either way, when offline with no basemap, surface a user-facing notice ("Base map not downloaded — connect to the internet once on `/maps`") instead of a silent grey screen.

### Field workaround (until fixed)
Boot online once and open `/maps`, or copy `world.pmtiles` from a provisioned unit into `storage/maps/pmtiles/`.

PR #1104, opened 2026-07-14
fix(maps): warn when world basemap missing instead of silent grey map offline

not-knope:
## Summary

Fixes the offline-first regression described in #1030: on a NOMAD that never extracted the world basemap while online, `/maps` renders blank/grey when offline (any view not covered by a downloaded regional extract).

Root cause (unchanged behavior, just no longer silent): `world.pmtiles` is lazily extracted from the internet on the first online `/maps` visit and is not shipped in the base-assets tarball, so an install that goes straight offline never gets it — and the UI gave no indication why the map was grey.

This PR doesn't change how the basemap is fetched; it makes its absence **visible and actionable** instead of a silent grey screen, and lets the user provision it deliberately while online.

## Changes

- **`MapService.checkWorldBasemapExists()`** — checks `storage/maps/pmtiles/world.pmtiles` directly on disk (path-traversal guarded), rather than trusting the in-process `worldBasemapReady` flag.
- **`MapService.provisionWorldBasemap()`** — ensures base assets, then extracts the world basemap on demand; returns whether it's now present.
- **`POST /maps/setup-world-basemap`** (+ `MapsController.setupWorldBasemap`, `api.setupWorldBasemap()`) — explicit "download the base map" action; returns a clear error when offline.
- **`worldBasemapExists`** is threaded through the `/maps` and `/settings/maps` Inertia props.
- **UI notices:**
  - `/maps`: shows a warning explaining the base map isn't downloaded and how to fix it, instead of a blank/grey map.
  - `/settings/maps`: shows a "World base map not downloaded" alert with a **Download Base Map** (~15 MB) button.

## Test plan

- [ ] Fresh install with no `storage/maps/pmtiles/world.pmtiles`, offline → `/maps` shows the notice (not a silent grey map); `/settings/maps` shows the download alert.
- [ ] Online, click **Download Base Map** → `world.pmtiles` is provisioned, notices disappear, map renders worldwide at low zoom.
- [ ] Offline, click **Download Base Map** → clear error notification, no crash.
- [ ] With `world.pmtiles` already present → no notices shown; existing behavior unchanged.

Refs #1030

### DIFF
--- a/admin/app/controllers/maps_controller.ts
+++ b/admin/app/controllers/maps_controller.ts
@@ -19,10 +19,14 @@ export default class MapsController {
 
   async index({ inertia }: HttpContext) {
     const baseAssetsCheck = await this.mapService.ensureBaseAssets()
-    const regionFiles = await this.mapService.listRegions()
+    const [regionFiles, worldBasemapExists] = await Promise.all([
+      this.mapService.listRegions(),
+      this.mapService.checkWorldBasemapExists(),
+    ])
     return inertia.render('maps', {
       maps: {
         baseAssetsExist: baseAssetsCheck,
+        worldBasemapExists,
         regionFiles: regionFiles.files,
       },
     })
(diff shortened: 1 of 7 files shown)


chriscrosstalk:
This is the right fix for #1030. Warning when the world basemap is missing instead of silently rendering a grey map is exactly the offline-first behavior we want, and failing loud beats a blank screen every time. Retargeted the base to `dev`. We'll take it for the next release.

jakeaturner (approved):
Awesome, thank you!

PR #1109, opened 2026-07-16
fix(gpu): detect Ollama CPU fallback from logs so auto-remediation fires

### LINKED ISSUE #1108: Auto GPU-passthrough remediation never fires when Ollama is installed CPU-only (nvidia-smi probe false-negative)
## Problem
On hosts where the NVIDIA container runtime is registered with Docker **only after** the AI Assistant (Ollama) was first installed, `nomad_ollama` ends up created CPU-only. `DockerService` attaches a GPU `DeviceRequest` only when `'nvidia'` is present in `docker.info().Runtimes` at install time, so a runtime added later leaves Ollama on the CPU.

The `GpuPassthroughRemediationProvider` is supposed to recover this automatically on admin boot, but it never fires.

## Root cause
The provider detects a broken passthrough by exec'ing `nvidia-smi` inside `nomad_ollama`:

```ts
const passthroughBroken =
  /Failed to initialize NVML|Unknown Error|TIMEOUT/i.test(output) ||
  !/[A-Za-z]/.test(output)
```

A CPU-only Ollama container does **not** ship `nvidia-smi`, so the exec fails with a string like `OCI runtime exec failed: ... nvidia-smi: executable file not found`. That string contains letters, so `!/[A-Za-z]/.test(output)` is `false` → `passthroughBroken` is `false` → the provider logs `NVIDIA passthrough healthy — no action needed` and skips. It cannot detect a missing GPU because the tool it probes with is itself missing.

## Repro
1. Install NOMAD on an NVIDIA box **without** the nvidia container runtime registered yet (toolkit installed, but `nvidia-ctk runtime configure` not run / no `/etc/docker/daemon.json`).
2. Install the AI Assistant → `nomad_ollama` is created CPU-only (correct, given no runtime at install time).
3. Register the runtime: `sudo nvidia-ctk runtime configure --runtime=docker && sudo systemctl restart docker`.
4. Restart the admin container.

**Expected:** admin detects Ollama is on CPU despite the runtime now being present, and reinstalls `nomad_ollama` with the GPU.
**Actual:** admin logs `NVIDIA passthrough healthy — no action needed`; Ollama stays on CPU. Models run at a fraction of a token/sec and the host sits at high CPU load. Only the manual **Fix: Reinstall AI Assistant** button recovers it.

## Impact
Likely affects many users on real NVIDIA hardware who install NOMAD before (or without) registering the nvidia runtime — a very common ordering. The silent CPU fallback is hard to diagnose (the UI/Settings reports the GPU as healthy).

## Proposed fix
Detect from Ollama's own `inference compute` startup log line (ground truth: `library=CUDA` vs `library=cpu`) instead of the `nvidia-smi` exec — the same source of truth `SystemService.getOllamaInferenceComputeFromLogs` already uses for the GPU-health display. Add a cooldown on `gpu.autoRemediatedAt` to avoid a reinstall loop for GPUs Ollama can't accelerate. PR incoming.

fpolica91:
Closes #1108.

## What & why
Fixes a false negative in `GpuPassthroughRemediationProvider`. When `nomad_ollama` is created CPU-only — most commonly because the NVIDIA container runtime was registered with Docker only **after** the AI Assistant was first installed (`DockerService` attaches a GPU `DeviceRequest` only when `'nvidia'` is in `docker.info().Runtimes` at install time) — the provider's `nvidia-smi` exec probe misread the missing-binary error as *"passthrough healthy"* and never auto-reinstalled, leaving Ollama silently on the CPU.

## Changes
- **`admin/app/services/ollama_compute.ts`** (new): pure `classifyOllamaComputeBackend(logText) → 'gpu' | 'cpu' | 'unknown'`, parsed from Ollama's `inference compute` startup line. No Docker dependency, fully unit-testable.
- **`admin/providers/gpu_passthrough_remediation_provider.ts`**: replace the `nvidia-smi` exec probe with the log-based classifier. One signal now covers both the *"created CPU-only"* case and the original *"DeviceRequests present but the toolkit binding tore"* case. Adds a 1-hour cooldown keyed on `gpu.autoRemediatedAt` so a GPU Ollama genuinely can't accelerate doesn't trigger a reinstall on every admin boot.
- **`admin/tests/unit/ollama_compute.spec.ts`** (new): 6 tests.

## Why logs instead of nvidia-smi
The `inference compute` line is what Ollama **actually loaded** — it works whether or not `nvidia-smi` ships in the image, and it's already the source of truth `SystemService.getOllamaInferenceComputeFromLogs` uses for the GPU-health display, so this aligns the provider with that path.

## Verification
- `tsc --noEmit` — clean.
- `node ace build` — succeeds.
- New unit suite — 6/6 pass (`CUDA→gpu`, `cpu-only→cpu`, `GPU + CPU fallback line→gpu`, `no line→unknown`, `empty→unknown`, `ROCm/Vulkan→gpu`).
- Reproduced the original failure on an NVIDIA GB10 host (Ollama 0.32.0): before any fix the provider logged *"healthy"* and left GLM-4.7 on the CPU (~0.06 tok/s); after a GPU-attached reinstall it runs on the GB10 at ~70 tok/s.

## Notes
I left the pre-existing lint findings in the provider file untouched — `eslint .` isn't clean across the repo and there's no lint CI, so I kept this diff focused on the bug. Happy to adjust scope, naming, or the cooldown window if you'd prefer different values.

### DIFF
--- a/admin/app/services/ollama_compute.ts
+++ b/admin/app/services/ollama_compute.ts
@@ -1,0 +1,41 @@
+/**
+ * Pure helpers for classifying which compute backend Ollama actually loaded,
+ * parsed from its container logs. Extracted so both SystemService (GPU health
+ * display) and GpuPassthroughRemediationProvider (auto-reinstall decision) share
+ * one source of truth, and so the logic is unit-testable without Docker.
+ *
+ * Ollama writes one `inference compute` log line per detected device a few seconds
+ * after startup, e.g.:
+ *   GPU:  ... msg="inference compute" id=0 library=CUDA compute=12.1 name=CUDA0 description="NVIDIA GB10" ...
+ *   CPU:  ... msg="inference compute" id=cpu library=cpu compute="" name=cpu ...
+ */
+
+export type OllamaComputeBackend = 'gpu' | 'cpu' | 'unknown'
+
+const INFERENCE_LINE_MARKER = 'msg="inference compute"'
+const GPU_LIBRARY_RE = /library=(CUDA|ROCm|Vulkan)\b/
+
+/**
+ * Classify Ollama's compute backend from raw container log text.
+ *
+ * - 'gpu'     if any `inference compute` line reports a GPU backend
+ *             (CUDA/ROCm/Vulkan). Ollama may emit a CPU line alongside a GPU
+ *             line (CPU is always listed as a fallback device), so a GPU line
+ *             anywhere means the GPU is usable.
+ * - 'cpu'     if `inference compute` lines exist but none report a GPU backend —
+ *             i.e. Ollama fell back to / was created with CPU only.
+ * - 'unknown' if no `inference compute` line is present (container too fresh,
+ *             logs rotated, or Ollama not started).
+ */
+export function classifyOllamaComputeBackend(logText: string): OllamaComputeBackend {
+  const lines = logText.split('\n').filter((line) => line.includes(INFERENCE_LINE_MARKER))
+  if (lines.length === 0) return 'unknown'
+
+  if (lines.some((line) => GPU_LIBRARY_RE.test(line))) return 'gpu'
+
+  // Lines exist but none matched a GPU library. A `library=cpu` line is a
+  // definitive CPU fallback; anything else (unrecognized library) stays unknown.
+  if (lines.some((line) => /library=cpu\b/.test(line))) return 'cpu'
+
+  return 'unknown'
+}
--- a/admin/providers/gpu_passthrough_remediation_provider.ts
+++ b/admin/providers/gpu_passthrough_remediation_provider.ts
@@ -1,22 +1,33 @@
 import logger from '@adonisjs/core/services/logger'
 import type { ApplicationService } from '@adonisjs/core/types'
+import type Docker from 'dockerode'
 
 /**
  * Auto-remediates NVIDIA GPU passthrough loss after admin / host restart.
  *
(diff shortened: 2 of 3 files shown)


chriscrosstalk:
Reviewed and approved, pending one boot-cycle test on our end.

The root cause is well spotted. A CPU-only container does not ship `nvidia-smi`, so the probe got a command-not-found error and read it as healthy, which meant the repair never fired for the case it was written for.

Three things I specifically checked, since this decides whether to recreate a container automatically:

- You read the startup time window rather than tailing. That matters, because the line is emitted once at boot and is far back in the log on a long-running box.
- A log read failure classifies as unknown, and unknown does nothing.
- The cooldown reads a key the existing code already writes, so it genuinely engages.

I also ran `classifyOllamaComputeBackend` against a real RTX 5060 startup line from one of our boxes and it returns `gpu`, so a healthy install is correctly left alone.

chriscrosstalk:
Boot test done, and this behaves correctly on real hardware.

Ran it on a box with a working RTX 5060 (the case that must be left alone), since the risk here is a false positive recreating the container on every boot:

```
[GpuPassthroughRemediationProvider] Ollama is using a GPU backend, no action needed.
```

The container id and StartedAt were byte-identical before and after the admin restart, so nothing was recreated. Typecheck is clean and all 6 unit tests pass.

Approved from our side. Thanks for the careful work on this one, particularly reading the startup window instead of tailing the log.

jakeaturner (approved):

PR #1120, opened 2026-07-18
feat(rag): add explicit opt-in for ZIM files in the knowledge base

### LINKED ISSUE #1119: [Feature]: ZIM/Kiwix files are silently embedded into the AI knowledge base with no opt-in or way to exclude them
### Feature Category

### Problem Statement

ZIM/Kiwix files are currently auto-discovered and embedded into the AI knowledge base as soon as they exist in storage — RagService._discoverKbFiles() walks both KB_UPLOADS_PATH and ZIM_STORAGE_PATH, and ZimService.registerLocalUpload() force-dispatches an embed job immediately after every ZIM upload. There's no way to exclude a ZIM from AI search short of deleting the file entirely. Browsing a ZIM's content in Kiwix and consenting to have that content embedded into the AI's searchable knowledge base are two different decisions, but the current design treats them as the same action — a privacy/consent gap for anyone who wants to keep a large offline library browsable without all of it automatically becoming part of what the AI assistant can retrieve and surface in chat responses.

### Proposed Solution

- Stop auto-discovering ZIM_STORAGE_PATH during KB sync — ZIM files stay fully available for Kiwix browsing but are no longer swept into RAG automatically.
- Add an explicit "Add to Knowledge Base" action per ZIM file in the Content Manager, making inclusion a deliberate per-file choice.
- Add a lightweight active/inactive toggle for files already added to the KB, so a file can be excluded from AI search and re-included later instantly, without deleting or re-embedding it.
- Hard-deleting an indexed ZIM also cleans up its knowledge-base entry (previously could leave an orphaned "Indexed" row behind until a full Reset & Rebuild).

### Alternative Solutions

Considered leaving auto-discovery in place and just adding a way to exclude files after the fact — rejected because it still defaults to "everything you browse is searchable" instead of defaulting to consent. Also considered a global on/off switch for ZIM auto-embedding rather than per-file control, but that doesn't let a user include some ZIMs (e.g. reference docs) while excluding others (e.g. personal/sensitive material) at the same time.

### Use Case

As a home user running NOMAD for offline access to a large personal ZIM library (technical docs, reference material, etc.), I want browsing content in Kiwix to never automatically make it part of the AI's searchable knowledge base, so that I control exactly what the assistant can retrieve from — and can toggle a file in or out of search instantly if I change my mind, without needing to delete and re-embed it.

All Users

Already implemented and tested locally — happy to open a PR:
- New `active` boolean column on kb_ingest_state (migration included), with a KbIngestState.setActive() helper.
- RagService.searchSimilarDocuments() excludes inactive sources via a Qdrant must_not filter.
- RagService.embedSingleFile() validates directly against disk instead of via _discoverKbFiles(), so a ZIM can still be explicitly added even though it's no longer auto-discovered.
- New POST /api/rag/files/toggle-active endpoint + validator + controller method.
- Content Manager UI gets a new "Knowledge Base" column: "Add to Knowledge Base" button for not-yet-added files, and the shared Switch component for the active/inactive toggle once a file's been added (per the UI Consistency guidelines).

### Examples or References

_No response_

Yes - I can write the code

### Additional Context

I have a branch (zim-kb-opt-in) with this fully implemented and manually tested against a running NOMAD instance: verified no auto-embed on upload/sync, verified explicit add-to-KB still works, verified the toggle persists instantly and correctly filters search results without re-embedding, verified hard-delete cleanup leaves no orphaned kb_ingest_state rows, and npm run typecheck passes. Ready to open as a PR pending feedback on this approach.

### Pre-submission Checklist

just-jbc:
Closes #1119

## Summary

ZIM/Kiwix files were previously auto-discovered and embedded into the AI knowledge base as soon as they existed on disk, with no way to exclude one without deleting it entirely. This decoupled the two decisions "browse this in Kiwix" and "let the AI search this," which weren't the same thing.

- Stop auto-discovering `ZIM_STORAGE_PATH` during KB sync — ZIM files must now be explicitly added via the Content Manager's new "Add to Knowledge Base" action.
- Add a per-file active/inactive toggle (`kb_ingest_state.active`) so an already-indexed file can be excluded from/re-included in search instantly, with no re-embedding.
- `searchSimilarDocuments()` excludes inactive sources via a Qdrant `must_not` filter.
- Hard-deleting an indexed ZIM now also cleans up its `kb_ingest_state` row and Qdrant points, instead of leaving an orphaned KB entry behind.
- Batched the `kb_ingest_state` lookup in `ZimService.list()` into a single query instead of one per file.

## Testing

Tested manually against a running NOMAD instance:
- Confirmed ZIM files no longer auto-embed on upload or Sync Storage.
- Confirmed explicit "Add to Knowledge Base" still embeds correctly.
- Confirmed the active/inactive toggle persists to the DB instantly and correctly includes/excludes the source from search results with no re-embedding.
- Confirmed hard-deleting an indexed ZIM cleans up its `kb_ingest_state` row (verified via direct DB query).
- `npm run typecheck` passes.

## UI

Content Manager's ZIM listing gets a new "Knowledge Base" column: an "Add to Knowledge Base" button for not-yet-added files, and the shared `Switch` component for the toggle once added — per the UI Consistency guidelines in CONTRIBUTING.md.


just-jbc:
Resolved a merge conflict with `dev` — the collections feature (PR #1063) landed after I branched, so `searchSimilarDocuments()`'s filter logic needed to combine the new `collection` scoping with this PR's inactive-source exclusion into a single Qdrant filter object instead of one silently overwriting the other. Also picked up the new OpenAPI `documented()` wrapping in `routes.ts` and added the `/files/toggle-active` route in that same style for consistency.

Rebuilt and re-tested locally after the merge — still working as described above. Ready for review whenever you get a chance.

just-jbc:
Merged latest dev (b4e4c64) to pull in the collections feature and OpenAPI generator changes, resolving conflicts in rag_service.ts (Qdrant filter logic) and routes.ts (route registration). After merging:

npm run typecheck passes with 0 errors
Re-ran the functional tests above against the merged branch — ZIM opt-in behavior, active/inactive toggle, and hard-delete cleanup all still work as described, no regressions from the merge

just-jbc:
Closing this — everything valuable here has since shipped through separate, more focused PRs: the ingest-policy bug fix landed in #1184, the orphaned-vectors cleanup in #1227, and the per-file toggle in #1192. The remaining piece (disabling ZIM auto-discovery by default) was correctly steered away from per the feedback above. No need to keep this one open.

PR #1130, opened 2026-07-21
Make the curated drug-reference data JSON the single source of truth

caweis:
Follow-up to #1040. In review, @jakeaturner suggested dropping the duplication between the repo-root `collections/{conditions,natural_remedies,home_remedies}.json` files and the `app/data/*.ts` constants, which currently have to be kept in sync by hand.

This makes the JSON the single source. `npm run gen:curated-data` generates the `app/data/*.ts` modules from the JSON, and a standalone test (`curated_data_sync.standalone.ts`) fails CI if a generated module drifts from its JSON, so the JSON is the only file anyone edits.

On the approach: you floated reading the JSON from the filesystem at runtime. I went with build-time codegen instead. The data stays compiled into the image the same way the `.ts` constant did, so there's no runtime file read and no container path resolution to get wrong. That was the original reason for the `.ts` constant, and it keeps the change verifiable in CI without a container run. If you'd rather read the JSON at runtime, I'm glad to switch; this felt like the lower-risk place to start.

One note on the diff: `conditions.ts` looks large, but it's formatting only. The generated data is byte-identical to what was there. The drift test asserts deep equality against the JSON, and I confirmed the same before committing. The generator emits standard JSON formatting rather than the previous hand-formatting.

The data is unchanged, and all standalone suites pass.

### DIFF
--- a/admin/app/data/conditions.ts
+++ b/admin/app/data/conditions.ts
@@ -1,342 +1,438 @@
+/* eslint-disable */
 /**
- * "When to use what" — curated condition spine (Phase 1, runtime source of truth).
+ * GENERATED FILE — DO NOT EDIT.
  *
- * A bounded, hand-curated list of common first-aid / emergency situations — a
- * "small booklet" of when-to-use-what. Each entry's `searchTerms` drive the
- * FULLTEXT search over `drug_labels.indications`.
+ * Single source of truth: collections/conditions.json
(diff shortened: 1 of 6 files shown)


jakeaturner:
@caweis Love this approach!

caweis:
Glad it fits. Ready whenever you want to pull it in. From here it's just the JSON to edit, and the drift test keeps the generated modules in step.

jakeaturner (approved):

PR #1135, opened 2026-07-22
fix(rag): stop re-creating payload indexes on every embedded document

### LINKED ISSUE #1129: [Bug]: RAG ingestion: _ensureCollection() re-creates payload indexes on every document
### Issue Category

Knowledge Base/RAG (Document Upload)

### Bug Description

_ensureCollection() in app/services/rag_service.js issues two createPayloadIndex
calls unconditionally on every invocation. Since it's called from the per-document
embed path (line 205), large ingestion jobs spend roughly 45% of their Qdrant time
re-creating payload indexes that already exist.

In _ensureCollection() (line 97), the `if (!collectionExists)` guard wraps only
createCollection. The two createPayloadIndex calls at lines 110 and 114 sit outside
that guard, so they run every time.

_ensureCollection() is called from 11 sites: lines 205, 540, 700, 710, 807, 860,
1061, 1081, 1175, 1236.

Qdrant treats repeat createPayloadIndex calls as idempotent, so there's no data
loss or corruption — the cost is purely wasted time. But it scales with corpus
size. The user-visible symptom is that large ingestion jobs (a full Wikipedia ZIM,
for example) appear stalled. They're progressing, just far slower than necessary.

Suggested fix: memoize per collection name with a Set on the service, checked at
the top of the method and set only after success, so a failed attempt still
retries. Alternatively, move the two createPayloadIndex calls inside the existing
if (!collectionExists) block — though that wouldn't create indexes on collections
predating the payload schema, so memoizing is probably safer.

### Steps to Reproduce

1. Start ingesting a large ZIM into the knowledge base
2. Run: sudo docker logs -f nomad_qdrant
3. Observe the 4-request cycle repeating for every document

### Expected Behavior

Payload indexes should be created once per collection, not once per document.
Steady-state ingestion should show only PUT /points calls.

### Actual Behavior

Two createPayloadIndex calls fire for every document ingested, in addition to the
one upsert that does real work.

Qdrant access log during a ZIM ingestion — this 4-request cycle repeats every ~46ms:

PUT /collections/nomad_knowledge_base/points?wait=true   0.005641
GET /collections                                          0.000064
PUT /collections/nomad_knowledge_base/index               0.010690
PUT /collections/nomad_knowledge_base/index               0.010746
PUT /collections/nomad_knowledge_base/points?wait=true    0.005516
GET /collections                                          0.000048
PUT /collections/nomad_knowledge_base/index               0.010474
PUT /collections/nomad_knowledge_base/index               0.010701

Only PUT /points does useful work. Each PUT /index costs ~10.5ms, so roughly 21ms
per document is spent re-creating indexes that already exist — about 45% of the
~46ms cycle.

No errors are produced. Qdrant treats repeat createPayloadIndex calls as
idempotent, so nothing is corrupted and no data is lost. The cost is purely wasted
time, and it scales with corpus size. On a collection of 5,084,133 points the
overhead runs to many hours of avoidable work.

User-visible symptom: large ingestion jobs appear to stall. They are progressing,
just far slower than necessary.

### N.O.M.A.D. Version

1.33.0

### Operating System

Ubuntu 24.04

### Docker Version

Docker version 29.6.1, build 8900f1d

Yes

### GPU Model (if applicable)

Gigabyte RTX 5060 Ti 16 GB

### System Specifications

 CPU: AMD Ryzen 7 7700X
RAM: 64 GB DDR5 (48GB virtual Proxmox 9.1)
GPU: Gigabyte RTX 5060 Ti 16GB
Disk space: Storage note: the 2TB NVMe is passed through raw to the VM by disk ID
(not a virtual disk), mounted ext4 at /mnt/nomad-data.

All services running normally. This is a performance issue in the ingestion
code path, not a service failure.

nomad_qdrant: Up 4 days
nomad_ollama: Up 3 days
nomad_admin: Up 4 days (healthy)

### Relevant Logs

```shell
docker logs nomad_qdrant — during ZIM ingestion. This 4-request cycle repeats
for every document, roughly every 46ms:

2026-07-19T04:49:02.929435Z INFO PUT /collections/nomad_knowledge_base/points?wait=true 200 0.005641
2026-07-19T04:49:02.930110Z INFO GET /collections                                        200 0.000064
2026-07-19T04:49:02.941194Z INFO PUT /collections/nomad_knowledge_base/index              200 0.010690
2026-07-19T04:49:02.952327Z INFO PUT /collections/nomad_knowledge_base/index              200 0.010746
2026-07-19T04:49:02.975579Z INFO PUT /collections/nomad_knowledge_base/points?wait=true 200 0.005516
2026-07-19T04:49:02.976152Z INFO GET /collections                                        200 0.000048
2026-07-19T04:49:02.986955Z INFO PUT /collections/nomad_knowledge_base/index              200 0.010474
2026-07-19T04:49:02.997982Z INFO PUT /collections/nomad_knowledge_base/index              200 0.010701
2026-07-19T04:49:03.023923Z INFO PUT /collections/nomad_knowledge_base/points?wait=true 200 0.005630
2026-07-19T04:49:03.024603Z INFO GET /collections                                        200 0.000068
2026-07-19T04:49:03.035743Z INFO PUT /collections/nomad_knowledge_base/index              200 0.010719
2026-07-19T04:49:03.046459Z INFO PUT /collections/nomad_knowledge_base/index              200 0.010360

All return 200 — no errors. The two PUT /index calls per cycle are redundant work.

Collection state (curl http://localhost:6333/collections/nomad_knowledge_base):
status: green, points_count: 5084133, indexed_vectors_count: 5276067,
vectors.size: 768, distance: Cosine, segments_count: 12,
payload_schema: content_type (keyword), source (keyword)
```

```javascript

```

### Screenshots

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

buraga-kyo:
## Summary

Closes #1129 — RAG ingestion re-creates Qdrant payload indexes on every document.

`RagService._ensureCollection()` runs on every `embedAndStoreText()` call (once per document on the ingestion path), but only `createCollection` sits behind the `if (!collectionExists)` guard. The `getCollections` probe and the three `createPayloadIndex` calls ran unconditionally — about 21ms of redundant Qdrant work in every ~46ms document cycle (~45%), which makes large ZIM ingestions look stalled while they're actually progressing (see the access logs in the issue).

## What changed

- Added a private `ensuredCollections: Set<string>` to `RagService`, following the same session-memoization pattern as the existing `embeddingModelVerified` flag.
- `_ensureCollection()` returns early for collections already ensured by this instance. The name is recorded only after the existence check and all three `createPayloadIndex` calls succeed, so a partial failure is retried on the next call.
- `checkQdrantHealth()` clears the Set when it resets the client — if Qdrant went away it may have been restarted or recreated, so collections are re-verified on reconnect.
- `resetAndRebuild()` drops the collection from the Set right after `deleteCollection`, so the `_ensureCollection()` call that follows actually recreates it.

Memoizing (rather than moving the index calls inside the `!collectionExists` guard, the alternative floated in the issue) keeps the current self-healing behavior: missing payload indexes are still created on collections that predate the current schema — e.g. the `collection` keyword index added for #1062.

## After the change

Steady-state ingestion issues only `PUT /points` per document. The 4-request cycle from the issue (`PUT /points` → `GET /collections` → 2–3× `PUT /index`) collapses to a single upsert; the ensure round-trips happen once per collection per worker session (each `EmbedFileJob` constructs a fresh `RagService`, so in practice once per job — negligible next to per-document).

## Testing

- `npm run typecheck` — clean
- `npm run build` — clean (same check as the PR CI workflow)
- `npx eslint app/services/rag_service.ts` — no new findings vs. `dev` (the file's pre-existing prettier findings are untouched to keep this diff focused)
- Not manually tested against a live ingestion — I don't have a NOMAD install with a large ZIM handy. The reproduction in #1129 (`docker logs -f nomad_qdrant` during ingestion) is the easiest way to confirm: after the first document only `PUT /points` should appear, and Settings → Reset & Rebuild should still recreate the collection and its indexes.

## Suggested release note

- **Fixed:** Knowledge Base ingestion no longer re-creates Qdrant payload indexes for every document, removing up to ~45% of per-document overhead on large ingestions (#1129)

### DIFF
--- a/admin/app/services/rag_service.ts
+++ b/admin/app/services/rag_service.ts
@@ -45,6 +45,10 @@ export class RagService {
   private qdrantInitPromise: Promise<void> | null = null
   private embeddingModelVerified = false
   private resolvedEmbeddingModel: string | null = null
+  // Collections already verified this session (created + payload indexes in place).
+  // Skips the getCollections/createPayloadIndex round-trips that otherwise run on
+  // every embed call — ~45% of per-document Qdrant time on large ingestions (#1129)
+  private ensuredCollections = new Set<string>()
   public static UPLOADS_STORAGE_PATH = 'storage/kb_uploads'
   public static CONTENT_COLLECTION_NAME = 'nomad_knowledge_base'
   public static EMBEDDING_DIMENSION = 768 // Nomic Embed Text v1.5 dimension is 768
@@ -94,6 +98,8 @@ export class RagService {
     } catch {
       this.qdrant = null
       this.qdrantInitPromise = null
+      // Qdrant may have restarted (or been recreated) — re-verify collections on reconnect
+      this.ensuredCollections.clear()
       return {
         online: false,
         message: 'Qdrant vector database is offline. Restart the AI Assistant service in Settings to restore the Knowledge Base.',
@@ -113,6 +119,11 @@ export class RagService {
   ) {
     try {
       await this._ensureDependencies()
+
+      if (this.ensuredCollections.has(collectionName)) {
+        return
+      }
+
       const collections = await this.qdrant!.getCollections()
       const collectionExists = collections.collections.some((col) => col.name === collectionName)
 
@@ -138,6 +149,9 @@ export class RagService {
         field_name: 'collection',
         field_schema: 'keyword',
       })
+
+      // Only memoize after every step succeeded, so a partial failure is retried
+      this.ensuredCollections.add(collectionName)
     } catch (error) {
       logger.error('Error ensuring Qdrant collection:', error)
       throw error
@@ -2103,6 +2117,10 @@ export class RagService {
         // Collection may not exist yet on a fresh install — log and continue.
         logger.warn(`[RAG] deleteCollection failed (may not exist): ${(err as Error).message}`)
       }
+
+      // The collection is gone — drop it from the ensured cache so the
+      // _ensureCollection call below actually recreates it
+      this.ensuredCollections.delete(RagService.CONTENT_COLLECTION_NAME)
 
       await this._ensureCollection(
         RagService.CONTENT_COLLECTION_NAME,


jakeaturner (approved):
@bragaus this is a great find and fix, thank you for contributing!

PR #1136, opened 2026-07-23
fix(benchmark): remove stale progress setter

### LINKED ISSUE #1133: [Bug]: System Benchmark fails
### Issue Category

System Performance/Resources

### Bug Description

After installing the new version: v1.34.0-rc.1, tried to run the Stem Benchmark... I get an error every time:

Benchmark Failed
setProgress is not defined

It does not matter which test I select, same failure with all 3.

Interesting to note, even with these failure message, if I scroll below, I do see Benchmark results

### Steps to Reproduce

Run the System Benchmark

### Expected Behavior

Should work

### Actual Behavior

Gives an error every time:

Benchmark Failed
setProgress is not defined

### N.O.M.A.D. Version

v1.34.0-rc.1

### Operating System

Ubuntu 24.04

### Docker Version

_No response_

Yes

### GPU Model (if applicable)

_No response_

### System Specifications

Intel i9
64GB RAM
2TB SSD
etc.

I'm using, and have been using for some time now, Ubuntu 26.04

_No response_

### Relevant Logs

```shell

```

```javascript

```

### Screenshots

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

NgoQuocViet2001:
## Summary

- Remove the orphaned `setProgress(...)` call from the benchmark start mutation.
- Keep live progress state owned by `useBenchmarkRun`, as intended by the benchmark telemetry refactor.
- Allow Full, System Only, and AI Only runs to reach `api.runBenchmark(...)` instead of throwing in the browser first.

## Root cause

The live benchmark UI moved `progress` state into `useBenchmarkRun`, removing the page-local setter. A later stacked benchmark change reintroduced the old `setProgress(...)` call without restoring that setter, so every benchmark button failed with `ReferenceError: setProgress is not defined` before the API request was made.

The run view already handles a `null`/`starting` status while it waits for the first SSE update, so deleting the stale optimistic update restores the hook boundary without changing benchmark execution or progress handling.

Closes #1133

## Validation

- `npm run build` — production build completed successfully.
- `npx tsc -p inertia/tsconfig.json --noEmit` — the `benchmark.tsx` diagnostic is gone; the command still reports pre-existing diagnostics elsewhere in the client tree.
- Diff is limited to the stale eight-line progress update in `admin/inertia/pages/settings/benchmark.tsx`.

## Suggested release note

- **Benchmark:** Fixed Full, System Only, and AI Only runs failing immediately with `setProgress is not defined` (#1133).

### DIFF
--- a/admin/inertia/pages/settings/benchmark.tsx
+++ b/admin/inertia/pages/settings/benchmark.tsx
@@ -93,14 +93,6 @@ export default function BenchmarkPage(props: {
       setErrorMsg(null)
       run.reset()
       setIsRunning(true)
-      setProgress({
-        status: 'starting',
-        progress: 5,
-        message: 'Starting benchmark... This takes 3-6 minutes.',
-        current_stage: 'Starting',
-        benchmark_id: '',
-        timestamp: new Date().toISOString(),
-      })
 
       // Use sync mode - runs inline without needing Redis/queue worker
       return await api.runBenchmark(type, true)


jakeaturner (approved):
Thanks for the fix!

PR #1162, opened 2026-07-27
feat(chat): add selectable AI personas

bgauger:
## Summary

Adds selectable AI chat personas that provide domain-focused guidance while preserving Project N.O.M.A.D.'s existing chat, RAG, model-management, and `NOMAD.md` behavior.

### Included

- Built-in Generalist, Medic, Electrician, Mechanic, Plumber, Builder, Bushcraft, Comms Operator, Homesteader, Vet, Security Planner, Grid, and Custom personas
- Per-chat persona selection and persistence
- Editable persona overrides with reset support
- Persona compatibility toggle for deployments that prefer the original prompt behavior
- Voice-anchored prompts and few-shot examples
- Plain-text output cleanup for LaTeX, trailing hedge paragraphs, and script leakage
- Persona Manager UI and documented persona API endpoints
- Database migrations for session persona keys and persona overrides
- Deterministic unit tests and a 73-question persona evaluation harness

## Compatibility

Persona prompts are composed with the current chat architecture, including:

- `NOMAD.md` instructions
- RAG context
- thinking-capable models
- model lifecycle and unloading behavior
- existing chat sessions and mobile layouts

Disabling personas preserves the existing default system prompt behavior.

## Validation

- `npm run typecheck`
- `npm run build`
- Persona unit tests: 6/6 passed
- OpenAPI structure and route-completeness tests: 2/2 passed
- Python evaluation harness compilation and question-bank JSON validation
- `git diff --check upstream/dev...HEAD`
- Layered added-line security review: 0 failures, 0 warnings
- Manual lifecycle verification against Ollama:
  - persona selection
  - session creation and persistence
  - model response generation
  - persona switching
  - invalid-persona validation (HTTP 422)
  - browser console verification

The broader upstream unit suite has three unrelated baseline/environment failures in custom-app path assumptions and drug-label whitespace handling. The affected tests and implementations are unchanged by this branch.

## Notes

This PR intentionally excludes the earlier experimental KB/RAG work and ports only the persona feature onto the current `dev` architecture.


chriscrosstalk:
Thank you for this — it's a lot of careful work, and the thinking behind it is right even though we're not going to merge it in this shape. I want to explain properly rather than just close it.

We've been designing the same surface from the other direction, and landed on a different structure. The short version: Knowledge Collections and personas are different mechanisms but the same user intent. Someone choosing "medical" wants their medical documents searched **and** a medically careful answer. Shipping both as separate selectors puts two dropdowns in the chat header for one intent, and asks the user to understand the difference between scoping retrieval and shaping tone — which is exactly the kind of thing our design principle says we've failed at if it needs explaining.

So the direction is to make a Collection do both jobs: it stays the tag that scopes what the knowledge base searches, and it gains an instructions field that shapes how the assistant answers on that subject. One dropdown, already in the header, already labelled for it. Editing lives in the existing Manage Collections modal rather than a new screen.

Two specifics from your PR that fed directly into that:

- Your persona prompts are the expensive part and they are being carried forward as the starting text for the built-in collection instructions. Thirteen domains of voice and guidance, written well, is not work we'd want to redo. They'll be shortened considerably — collection instructions are meant to be a hundred words or two, not multi-page — but the content and the judgement in them is the input.
- One line has to change wherever it lands: the Electrician persona instructs the model to "cite the knowledge base when you use it", and the existing RAG context rules explicitly tell it not to name sources. We resolved that in favour of keeping the model quiet in prose and surfacing sources as a UI element built from retrieval metadata instead, which is what #1287 now does. So provenance is coming — just not by asking the model to narrate it.

There's also a timing reason to be straight about: the developer is actively reworking the LLM behaviour in parallel right now, and persona prompts are system-prompt composition. Merging a large change to that surface underneath him would create conflicts neither of us can see.

Closing this as won't-merge-as-is rather than because it's wrong. The idea is one we're adopting; it's the packaging we're changing. Genuinely appreciated, and I hope you'll keep contributing.

Issue #1175, opened 2026-07-30
[Feature]: Download offline aerial imagery and topographic maps by region


gabegraves:
## Feature Category

Content Management (ZIM, Maps, Collections)

## Problem Statement

NOMAD's offline map currently provides a vector street basemap only. Users cannot
download local aerial imagery or rendered topographic maps for offline visual reference,
and the map pipeline assumes PMTiles archives are vector data.

## Proposed Solution

Add explicit vector/raster archive metadata, a provider-driven regional raster download
pipeline, and map modes for Street, Aerial imagery, Topographic, and Hybrid. Users choose
eligible regions, layers, and maximum zoom; an authoritative preflight shows tile count,
final size, temporary workspace, and remaining disk before any download starts.
Downloads are resumable, verified, removable, and activated atomically.

The implementation should be provider- and region-driven rather than hardcoding one
regional pack. A first provider can use allowlisted USGS ArcGIS services and validated
administrative boundary IDs. The UI should call the imagery product **Aerial imagery**,
preserve the exact source attribution, and restrict presets to regions whose source
terms permit direct export.

## Alternative Solutions

- Continue using vector street maps only.
- Require users to package raster PMTiles externally and copy them into NOMAD manually.
- Ship one fixed regional pack. This is less maintainable and does not serve the wider
  NOMAD community.

## Use Case

As a user preparing for disconnected travel, education, research, field reference, or
emergency use, I want to select one or more regions and download aerial and/or
topographic coverage so I can switch map modes while completely offline without losing
the existing vector map, markers, camera, or scale.

## Who Would Benefit

All Users; Teachers/Educators; Students; Survivalists/Preppers; Organizations

## Priority

High - Would significantly improve my experience

## Implementation Ideas

I have a working proof of concept with:

- Explicit raster/vector archive metadata and defensive manifest parsing.
- USGS authoritative preflight and recursive splitting below the 100,000-tile export
  limit.
- Resumable range downloads, durable chunk state, checksums, PMTiles verification,
  atomic activation, pause/resume/cancel, and disk gates.
- Street, aerial, topographic, and hybrid rendering with immediate layer switching.
- 56 focused unit/integration tests and full offline TX/OK/GA proof data.

For reviewability, I propose three focused PRs:

1. Provider-neutral raster PMTiles manifest/style/conversion foundation.
2. Regional provider, preflight, resumable download, update, and uninstall lifecycle.
3. Settings and map presentation controls.

[...]

PR #1177, opened 2026-07-30
feat(maps): add raster PMTiles manifest and conversion foundation

gabegraves:
## Summary

- add a versioned, filename-keyed, discriminated PMTiles sidecar manifest for explicit vector and raster archive metadata
- reject contradictory vector/raster metadata and fail closed when a manifest is malformed or an archive entry is invalid
- add a provider-neutral MBTiles-to-PMTiles conversion boundary that converts to hidden staging, verifies the resulting raster header and metadata, then activates the archive and manifest atomically
- validate tile type, role, region, bounds, zooms, provenance, checksum, installed size, and verification timestamp before style activation
- compose collision-free hidden raster layers below existing vector layers without mutating the base style template
- preserve existing unmanifested vector maps and newest-version deduplication without allowing malformed raster archives to fall back to vector handling

## Why

This completes PR 1 from #1175: the provider-neutral manifest, style, and conversion foundation for regional offline aerial and topographic map packs. Provider catalogs, download lifecycle, storage UI, and map-layer controls remain intentionally out of scope for this focused change.

Part of #1175

## Storage findings and revised lifecycle

These are measured installed PMTiles sizes from verified USGS 3.3.0 proof archives, not area-based estimates. Decimal GB is shown because that is how disk vendors report capacity; exact bytes are authoritative.

| Region | Layer | Max zoom | Exact bytes | Decimal GB |
|---|---|---:|---:|---:|
| Oklahoma | Aerial imagery | 14 | 1,399,966,769 | 1.400 |
| Oklahoma | Aerial imagery | 15 | 5,402,512,265 | 5.403 |
| Oklahoma | Topographic | 14 | 1,199,003,986 | 1.199 |
| Georgia | Aerial imagery | 14 | 1,419,013,810 | 1.419 |
| Georgia | Topographic | 14 | 1,191,841,546 | 1.192 |
| Texas | Aerial imagery | 14 | 4,717,245,187 | 4.717 |
| Texas | Topographic | 14 | 3,655,543,463 | 3.656 |

The Oklahoma z14 aerial value was measured by extracting `--maxzoom=14` from the verified Oklahoma z15 archive and statting the resulting PMTiles file. The z15 Oklahoma aerial archive is 3.859 times the z14 archive. The initial provider proposal will therefore offer aerial z14/z15 and topographic z14; topographic z15 will not be offered until it has its own measured storage record.

The follow-up lifecycle PR will use NOMAD's existing machinery rather than the separate resume/chunk system from the proof of concept:

- each provider export is dispatched through `RunDownloadJob`/`doResumableDownload` with a raster-workspace file type, so resume, retry, progress, cancellation, and queue visibility remain in `DownloadService`
- only the final verified archive becomes an `InstalledResource`; provider chunks and MBTiles files remain a deterministic workspace owned by that resource ID
- one authoritative preflight response reports `installedBytes`, `workspaceBytes`, and `peakAdditionalBytes`; `peakAdditionalBytes` feeds the existing `StorageProjectionBar` using `getPrimaryDiskInfo`, so the download modal and setup storage projection cannot disagree
- cancellation, terminal failure, restart recovery, and uninstall all target the same deterministic workspace; uninstall stats the installed archive plus any workspace before deletion and returns the actual reclaimed byte count
- the final archive checksum remains part of the manifest and conversion verification; no second checksum ledger is introduced

This PR's converter already removes its own hidden `.partial` archive on success or failure. Caller-owned provider workspaces are deliberately left to the lifecycle boundary above, which is not implemented in this foundation PR.

## Testing

- `npm run typecheck`
- focused ESLint and Prettier checks for the changed files
- exact-commit Docker test run for `map_style_generation.spec.ts` and `raster_pmtiles_conversion.spec.ts` (13 passing)
- `npm run build`
- verified an installed raster archive with `pmtiles show --header-json` and confirmed the converter reads the real CLI fields (`tile_type`, zooms, and bounds)
- measured the installed proof archives above with filesystem `stat`; the temporary Oklahoma z14 measurement archive was removed afterward
- `git diff --check`

Environment: macOS 26.5.1 (Apple Silicon), Node.js 24.14.0 workspace runtime, Docker Desktop 29.6.2.

## Suggested release note

Adds the map-style foundation required for verified offline raster PMTiles layers while preserving existing vector-map behavior.


chriscrosstalk:
Thanks for this, and for the storage measurements in particular — measuring real archives rather than estimating from area is what made this decidable. The numbers are accepted: offline satellite and topographic maps are worth the disk, and a user who wants both can decide for themselves whether they have the space.

The lifecycle design in your description is the part that unblocks this. Routing provider exports through the existing `RunDownloadJob`/`doResumableDownload` path, so resume, retry, progress and cancellation stay in `DownloadService` instead of a parallel resume system, is the right answer to the question this was previously stuck on.

Your branch had drifted onto a stale base, so I've cherry-picked both commits onto a fresh branch off `dev` with your authorship intact: #1289. It is deliberately a draft — on its own the foundation is scaffolding nothing calls, and it also refactors live map-service and style-generation code, so merging it alone would put risk into a release for no user-visible benefit. It should land together with the lifecycle and UI work.

To be straight with you about what happens next rather than leaving you waiting: we're going to build the remaining pieces in-house and ship the whole feature together in a later version. Your foundation and your lifecycle design are carried forward and credited.

Closing this in favour of that draft. Appreciated.

PR #1178, opened 2026-07-30
feat(ai): support local vision attachments

### LINKED ISSUE #1176: [Feature]: Attach images to chats with local multimodal models
## Feature Category

AI Assistant Enhancement

## Problem Statement

NOMAD can use local OpenAI-compatible and Ollama chat models, but its chat request schema
and UI only accept text. Users with an installed multimodal model cannot analyze a
photo, screenshot, diagram, worksheet, or document image while offline.

## Proposed Solution

Add secure, ephemeral JPEG/PNG/WebP attachments to the latest user message. Detect model
capabilities when the backend exposes them, normalize images locally, send
OpenAI-compatible `image_url` content parts upstream, and keep binary image data out of
chat persistence, RAG, title generation, logs, and remote URL fetching.

Known text-only models should reject attachments before inference. Backends that do not
report capabilities should be represented as unknown rather than guessed from their
model names, with a clear warning and actionable compatibility errors.

## Alternative Solutions

- Use a separate model-specific vision application.
- Ask users to base64-encode images manually and call the backend API.
- Persist every attachment in NOMAD. That creates substantially more privacy, storage,
  migration, and lifecycle scope than an initial ephemeral implementation.

## Use Case

As an offline NOMAD user, I want to attach a diagram or photo and ask my selected local
vision model to describe, explain, compare, transcribe, or answer questions about it.

## Who Would Benefit

All Users; Teachers/Educators; Students; Developers/Contributors; Organizations

## Priority

## Implementation Ideas

- Detect Ollama vision capability from `/api/show`.
- Detect llama.cpp image support from `/props.modalities`.
- Represent unsupported and unknown capability separately.
- Keep existing JSON chat requests unchanged for text-only messages.
- Use multipart uploads for image-bearing requests instead of increasing the global JSON
  body limit.
- Validate count and size, then decode and normalize every image server-side with the
  existing `sharp` dependency.
- Reject corrupt, forged, oversized, excessive-pixel, animated, and multipage inputs.
- Normalize orientation, resize, strip EXIF/GPS metadata, and base64-encode only the
  bounded normalized image.
- Attach content parts only to the newest user turn.
- Delete temporary files on success, error, and abort.
- Use revocable browser object URLs rather than retaining base64 strings in React state.
- Preserve current `dev` behavior for NOMAD.md, RAG collection filters, thinking,
  reasoning normalization, sessions, title generation, model switching, and
  client-disconnect aborts.

The local proof of concept already sends OpenAI-compatible `image_url` content to
llama.cpp and works with a local multimodal model, but it needs the validation,
capability, privacy, and regression work above before it is suitable upstream.

## Examples or References

- Ollama model capabilities:
  https://docs.ollama.com/api-reference/show-model-details
- Ollama OpenAI-compatible vision:
  https://docs.ollama.com/api/openai-compatibility
- llama.cpp multimodal server:
  https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md
- AdonisJS file-upload security:
  https://docs.adonisjs.com/guides/basics/file-uploads

## Would You Be Willing to Help Implement This?

Yes - I can write the code

## Additional Context

Initial non-goals: persistent image history, PDFs, video/audio attachments, OCR indexing
into RAG, remote image URL fetching, and automatic vision-model/projector installation.
V1 attachments are intentionally ephemeral and are not restored after a page reload.

The implementation will include focused unit/controller tests, text-only regression
tests, and real-browser/backend verification against Ollama vision, llama.cpp vision,
a known text-only model, and an OpenAI-compatible backend with unknown capability.

Open PR #1162 modifies several of the same chat files, so the implementation will be
rebased and regression-tested against its accepted result if it merges.

## Pre-submission Checklist

- [x] I searched open issues and pull requests for similar vision, multimodal, image,
  and chat-attachment work.
- [x] This feature aligns with NOMAD's offline-first knowledge and education mission.
- [x] I understand that feature requests are not guaranteed to be implemented.

gabegraves:
## Summary

- add JPEG, PNG, and WebP attachments to chat using multipart requests while leaving text-only JSON requests unchanged
- normalize images server-side to bounded JPEG, strip metadata, cap count/size/pixels/output, and clean temporary uploads on every validation path
- attach image parts only to the latest user message sent to the local OpenAI-compatible backend
- resolve vision support per model from OpenAI-compatible `/v1/models` metadata, Ollama `/api/show`, then llama.cpp `/props`; allow an explicit `unknown` state when no backend reports capability
- disable attachments for known text-only models and translate both streaming and non-streaming image rejections into plain, actionable compatibility guidance
- always explain in the chat UI that images are request-only, are not saved, and disappear after sending or reloading
- ensure recommended offline models include an image-capable option when the catalog offers one, provide Qwen3-VL in the offline fallback list, and mark compatible setup choices as `Supports images`
- preview attachments in the current chat while excluding image bytes from RAG, titles, logs, and persisted chat history; keep explanatory tooltips inside desktop and mobile viewports

Closes #1176

## Testing

- `npm run typecheck`
- focused ESLint/Prettier checks for the new helper and regression-test files
- exact-source Docker build using the repository's Node 22 image: `node ace build` passed
- exact-source Docker tests for image validation, text-request compatibility, stream handling, capability resolution, recommendation selection, temporary-image guidance, and controller behavior (21 passing)
- regression coverage for misleading filenames: WebP bytes uploaded with a `.png` name are decoded by content and normalized to a valid bounded JPEG
- real browser/backend matrix:
  - Ollama `qwen2.5vl:3b`: attachment accepted and the model correctly identified the test image
  - llama.cpp `qwen3.6:35b-a3b-hauhau-aggressive` Q4_K_M with its matching F16 vision projector: attachment accepted and the model correctly identified the test image
  - known text-only Qwopus 27B: attachment disabled with a visible explanation; text chat returned the expected response
  - unknown OpenAI-compatible backend: attachment allowed with a warning and backend rejection surfaced the actionable compatibility message
- verified the tooltip and attachment states in real desktop and mobile viewports without clipping
- `git diff --check`

The 35B configuration was also validated directly through the same OpenAI-compatible router used by NOMAD. The router advertised multimodal capability through `/v1/models`; a live image request returned HTTP 200 and the model correctly identified the Project NOMAD interface. Its low-refusal fine-tuning is independent of vision support, which comes from the multimodal architecture and matching projector.

Environment: macOS 26.5.1 (Apple Silicon), Node.js 22.23.2 in Docker, Docker Desktop 29.6.2, llama.cpp with a local Qwen 35B multimodal model.

## Local deployment context

Our primary local reasoning model is `qwen3.6:35b-a3b-hauhau-aggressive`, an unofficial community fine-tune commonly described by its users as a “heretical” or low-refusal model. We selected it because it provides strong capability for its memory footprint and preserves user-controlled offline discussion of ambiguous or gray-area questions that may arise in a survival scenario.

That model is a user-selected validation configuration, not a model bundled, downloaded, or endorsed by this PR. The feature remains model-neutral and does not infer vision support from a model name: image attachments depend on the selected backend reporting compatible multimodal capability (and, where required, loading the appropriate vision projector).

## Coordination

This branch overlaps five files touched by open PR #1162. Both currently target the same `dev` base; if #1162 merges first, this branch will be rebased and the full regression matrix rerun before merge.

## Suggested release note

Adds secure, ephemeral image attachments for compatible local multimodal chat models and makes compatible offline model choices discoverable.


chriscrosstalk:
Thanks for this — it's the most thorough submission we've had. Resolving vision support from what the backend actually reports rather than from the model name is the right call and not the obvious one; it means this keeps working as new models appear. The mislabeled-upload coverage (WebP bytes with a `.png` name) and translating a backend rejection into plain guidance instead of a raw error are both the kind of detail that usually gets skipped.

The chat retrieval path was refactored into a pipeline service after you filed this, so the controller half didn't drop in as-is. Rather than ask you to redo eight commits, I've rebased your work onto the new structure with your authorship intact: #1288.

Three things worth flagging from that:

- `dev` split chat into a native Ollama path and an OpenAI-compat path in the meantime. Ollama's native `/api/chat` doesn't take OpenAI content parts — it wants a sibling `images` array of bare base64 — so I added a converter. Without it, image requests on a native Ollama backend would have failed. Your original only had the compat path, so this wasn't reachable when you wrote it.
- Your multipart-aware collection filter is now the only one. `dev`'s version reads the collection with `request.input()`, which would have silently dropped it on every image request — good instinct.
- `tests/unit/ollama_controller_chat.spec.ts` can't run under the plain `node:test` runner, because constructing the controller pulls in the Adonis logger service at module scope. It fails the same way on your branch, so that file's coverage had never actually executed. I've moved it to the Japa runner where it has a booted app.

I've also held back the recommendation change (`selectRecommendedModels` plus the Qwen3-VL fallback entry) — displacing a model in the first-time setup list is a product call being taken separately. The "Supports images" label stays, since it describes what's already on offer.

Verified on real hardware: attached an image on a GPU box running `qwen2.5vl:3b` and it described the contents correctly, and a model reporting no vision support disables the control with your explanation.

Closing this in favour of that PR. Genuinely good work.

Issue #1170, opened 2026-07-29
Knowledge base retains Qdrant vectors for ZIM files that were deleted or replaced


chriscrosstalk:
### Summary

The knowledge base never reaps Qdrant points whose source file is gone from disk. A user can be handed a search citation pointing at a ZIM that no longer exists, and nothing surfaces that the hit is stale — the orphaned vectors score normally and compete with live ones on merit.

### Evidence

On an install with a 77,939-point knowledge base, faceting `source` returns seven sources that are not on disk:

| Indexed source | On disk now |
|---|---|
| `devdocs_en_node_2026-02.zim` | `_2026-05` |
| `devdocs_en_git_2026-04.zim` | `_2026-07` |
| `devdocs_en_html_2026-04.zim` | `_2026-07` |
| `devdocs_en_css_2026-04.zim` | `_2026-07` |
| `devdocs_en_javascript_2026-04.zim` | `_2026-07` |
| `cooking.stackexchange.com_en_all_2025-12.zim` | `_2026-02` |
| `wikinews_en_all_maxi_2026-04.zim` | **absent entirely** |

```bash
curl -s -X POST 'localhost:6333/collections/nomad_knowledge_base/facet' \
  -H 'Content-Type: application/json' -d '{"key":"source","limit":20}'
```

### Three gaps, one fix

**1. The delete path does not touch Qdrant.** `ZimService.delete()` removes the file, calls `KiwixLibraryService.removeBook()`, drops the `InstalledResource` row and clears any `WikipediaSelection` — but never deletes the file's points, and never drops its `KbIngestState` row. This is the `wikinews` case.

**2. The replace path has a documented permanent leak.** `RagService.reconcileReplacedContentFile()` is otherwise careful, but case 5 (old file was indexed, Qdrant not currently running) is a deliberate no-op whose comment states the old points "are NOT auto-reaped here." Nothing reaps them later either, so a content update while Qdrant is down leaks permanently.

**3. The sync only walks one direction.** `syncKnowledgeBase()` already builds `sourcesInQdrant` from a facet and `embeddableFiles` from disk — both sides of the comparison, in memory, at the same moment. It only ever asks "is this on-disk file already embedded?" It never takes the reverse set difference, so orphans from (1) and (2) are never caught up.

### Suggested fix

Adding the reverse sweep to `syncKnowledgeBase()` closes all three, since sync already runs on the relevant triggers and already has the data:

```
[...]

PR #1180, opened 2026-07-31
feat: show source document and date as citations under RAG chat answers

### LINKED ISSUE #1179: [Feature]: Surface source document + date as citations under AI chat answers
### Feature Category

### Problem Statement

When NOMAD's AI answers a question using retrieved knowledge-base context, the answer gives no indication of which document backed it or how current that information is. For an offline tool meant to be relied on without internet access to cross-check anything, that's a real gap — especially for time-sensitive or safety-relevant content (e.g. the drug reference), where staleness and provenance matter.

### Proposed Solution

Surface a deduplicated "Sources" list (title + date) under AI answers. ZIM-derived chunks already carry `archive_date`/`archive_title` in their Qdrant payload at ingestion time, but `searchSimilarDocuments()`'s returned metadata currently only maps `article_title`/`full_title`/`section_title` — the date fields exist in storage but aren't surfaced. This adds them to that mapping so both the internal context label and a new structured `sources` array can use them, adds a nullable `sources` column to `chat_messages` (small migration) threaded through `addMessage()` so sources persist with the message and still show when reopening a past conversation, and renders the list under assistant messages in the chat UI.

### Alternative Solutions

Inline numbered citations ([1], [2] in the answer text) were considered and rejected for v1 — relies on the model reliably emitting correctly-attributed markers, which isn't guaranteed and risks misattributing a claim to the wrong source.

### Use Case

As a NOMAD user, when the AI answers a question using knowledge-base context, I want to see which document it came from and how current that information is, so that I can judge how much to trust the answer before relying on it — especially for time-sensitive or safety-relevant content like the drug reference.

All Users

- Add `archive_date` and `archive_title` (already stored in the Qdrant payload at ingestion, `rag_service.ts`) to the `metadata` object returned by `searchSimilarDocuments()`, which currently omits them.
- Build a deduplicated `sources` array (title, date, filename) from the retrieved chunks in `ollama_controller.ts`'s chat method, alongside the existing context-label logic.
- Add a nullable `sources` column to `chat_messages` via migration; extend `ChatService.addMessage()` to accept and store it.
- Return `sources` in both the streaming and non-streaming chat response shapes.
- Frontend: render a small "Sources" list under assistant messages when present.

### Examples or References

Similar in spirit to how AI search tools (e.g. Perplexity) show source citations under generated answers, adapted here for NOMAD's offline knowledge base.

Yes - I can write the code

### Additional Context

Non-ZIM sources (plain PDFs) don't carry embedded date/author metadata the way ZIM archives do, so citations for those will show title/filename only, no date — an inherent data limitation, not a bug in this feature.

### Pre-submission Checklist

just-jbc:
Closes #1179

## What
Surfaces the underlying document title and date (when available) below assistant chat responses that used RAG context, so users can see where an answer came from. Falls back to a clean filename for sources lacking embedded metadata (e.g. user-uploaded PDFs).

## Changes
- Added `archive_title`/`archive_date` to `RAGResult` and threaded them through both metadata mapping stages in `rag_service.ts`
- Built a deduplicated sources list in `ollama_controller.ts`, streamed as a trailing SSE event for the streaming chat path, and persisted for the non-streaming path
- Added a nullable `sources` column to `chat_messages` (migration) plus model/service support for storing and returning it
- Rendered a Sources list in `ChatMessageBubble.tsx`; wired an `onSources` callback through the streaming client (`api.ts`) and chat UI (`index.tsx`)

## Testing
Manually tested end-to-end on a self-hosted instance:
- ZIM-sourced answers show source title + date
- User-uploaded PDF sources (no embedded metadata) fall back to a clean filename
- Sources persist correctly when reopening a past chat session
- No sources shown for answers that didn't use RAG context


chriscrosstalk:
Nice work. This is a clean fix and the diagnosis is exactly right: the `archive_date`/`archive_title` fields were already being written into the Qdrant payload at ingestion and just weren't making it through the metadata mapping, so most of this is surfacing data we already had.

Verified the approach against `dev`. Threading through both mapping stages in `rag_service.ts` is the right place for it, the dedupe belongs in the controller where you put it, and falling back to a clean filename for PDFs without embedded metadata is the correct call rather than showing a blank date.

Two notes, neither blocking:

Once this is in, answers will name the source document on screen. There's a known issue where the knowledge base can retain vectors for a ZIM that's since been deleted or replaced (#1170), which means this could occasionally surface a title for a file that's no longer on disk. That's arguably a point in its favor since it makes a silent problem visible, but flagging it so it isn't a surprise.

The new migration bumps the count, so our QA runner will throw its usual benign count warning on the next release pass. Nothing for you to do.

Approving. This one goes in front of our dev team for merge.

chriscrosstalk:
Thanks for this — it's a feature the project needed. Offline there's no second opinion to check an answer against, so being able to see which document backed it is most of what makes the assistant trustworthy for anything serious.

Two things worth calling out from your version: you built the list from the trimmed set rather than everything retrieval returned, which is the correct call and not an obvious one — citing a chunk that got dropped before the model saw it would credit the answer to a document it wasn't based on. And routing the citations as a trailing event with no `message` key so the client can tell it apart from Ollama's own chunks is a clean way to do it.

The chat retrieval path was refactored into a pipeline service after you filed this, so the controller half didn't drop in as-is. Rather than ask you to redo it, I've rebased your commit onto the new structure with your authorship intact: #1287. The dedupe and fallback logic moved into a helper next to the context-block builder so it could be unit-tested on its own, and it now dedupes on the document path rather than the title so two editions of the same archive stay distinct.

One nice thing that fell out: the archive title and date were already being written into the vector payload at ingest and had simply never been read back, so this works on content people have already embedded.

Closing this in favour of that PR. Much appreciated.

PR #1184, opened 2026-07-31
fix(kb): respect ingest policy when a ZIM is uploaded locally

chriscrosstalk:
## Summary

`ZimService.registerLocalUpload()` dispatched `EmbedFileJob` unconditionally, gated only on whether Ollama was reachable. It never read `rag.defaultIngestPolicy`:

```ts
const ollamaUrl = await this.dockerService.getServiceURL('nomad_ollama')
if (ollamaUrl) {
  const { EmbedFileJob } = await import('#jobs/embed_file_job')
  await EmbedFileJob.dispatch({ fileName: filename, ... })
}
```

So a user who deliberately set **Manual** and then sideloaded a ZIM got it embedded into the knowledge base anyway.

**PR #919 fixed exactly this for the post-download path. The local-upload path was missed.**

## Approach

Rather than inline a third copy of the Always/Manual conditional, this reuses `decideScanAction`, the same helper `scanAndSyncStorage` uses.

That's not just tidier, it fixes a second case: an existing `browse_only` or `pending_decision` row is now honored, instead of being overridden by the act of re-uploading the file. The inline version in `run_download_job.ts` doesn't do that.

Unset policy is still treated as `Always`, so existing installs keep current behavior.

## Testing

Verified live on NOMAD3 (v1.34.0-rc.3) by uploading the same ZIM through `POST /api/zim/upload` under each policy, then counting the resulting vectors in Qdrant:

| policy | `kb_ingest_state` row | `EmbedFileJob` | points in `nomad_knowledge_base` |
|---|---|---|---|
| **Manual** | `pending_decision`, chunks=0 | not dispatched, no log lines | **0** |
| **Always** | dispatched and processed | `[EmbedFileJob] Dispatched...` | **1076** |

Same file, same code path, difference is entirely the policy. Under Manual the file lands as `pending_decision`, which is what surfaces the per-file Index affordance in the KB panel (PR #909), so the user can still opt it in deliberately.

All test artifacts removed afterward and the box restored to its prior state (11 indexed / 8 pending, policy back to Manual, test vectors deleted).

`npm run typecheck` clean.

Decision logic itself is already covered exhaustively by `tests/unit/kb_ingest_decision.spec.ts`, including every case this call site depends on: no-row + Manual → `create_pending`, no-row + Always → `dispatch`, `browse_only` → `skip`, `pending_decision` + Manual → `skip`.

## Scope

Reported by @just-jbc on #1119, which also proposed disabling ZIM auto-discovery entirely. **That larger change is deliberately not included here.** Turning discovery off by default would mean someone who downloads Wikipedia through our own curated flow gets no AI answers from it and no explanation why. This PR fixes only the part where we ignore a setting the user already chose.

Refs #1119

### DIFF
--- a/admin/app/services/zim_service.ts
+++ b/admin/app/services/zim_service.ts
@@ -597,14 +597,40 @@ export class ZimService {
 
     const ollamaUrl = await this.dockerService.getServiceURL('nomad_ollama')
     if (ollamaUrl) {
+      // Respect the global ingest policy, same as the post-download path (PR #919).
+      // This used to dispatch unconditionally, so a user who deliberately chose
+      // Manual still got sideloaded ZIMs embedded behind their back.
+      //
+      // Reuses decideScanAction rather than re-inlining the Always/Manual check,
+      // so an existing browse_only or pending_decision row is honored too instead
+      // of being overridden by the act of re-uploading the file.
+      const filePath = join(process.cwd(), ZIM_STORAGE_PATH, filename)
       try {
-        const { EmbedFileJob } = await import('#jobs/embed_file_job')
-        await EmbedFileJob.dispatch({
-          fileName: filename,
-          filePath: join(process.cwd(), ZIM_STORAGE_PATH, filename),
-        })
+        const { default: KVStore } = await import('#models/kv_store')
+        const { default: KbIngestState } = await import('#models/kb_ingest_state')
+        const { decideScanAction } = await import('../utils/kb_ingest_decision.js')
+
+        // Unset is treated as Always, preserving legacy behavior — mirrors
+        // rag_service.ts and run_download_job.ts.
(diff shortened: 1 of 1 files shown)


chriscrosstalk:
Retargeted from `dev` to `rc` for inclusion in **v1.34.0-rc.4** (Chris's call).

The branch was cut from `12b2494`, which is already in `rc`'s history, so this rebases cleanly and adds exactly one commit on top of `rc`. One file, +32/-6, no conflicts with the five PRs that landed on `rc` on 2026-07-29.

Rationale for pulling it into the release rather than holding for 1.35: the bug means we ignore a setting the user explicitly chose. Verified live on NOMAD3 — same file uploaded under each policy produced 0 vectors under Manual vs 1,076 under Always.

jakeaturner (approved):

PR #1192, opened 2026-08-03
feat(kb): per-file active toggle + collection grouping/bulk toggle

### LINKED ISSUE #1119: [Feature]: ZIM/Kiwix files are silently embedded into the AI knowledge base with no opt-in or way to exclude them

### LINKED ISSUE #1193: KB panel: group personal uploads by collection to reduce clutter (+ optional bulk toggle)
**Problem**
When a user uploads many personal files to the knowledge base, each file shows as its own row in the KB panel's file table. With dozens of uploads this becomes a long, hard-to-scan list.

**Proposed solution**
- Group files by their `collection` value under collapsible header rows — one header per collection, expandable to show member files.
- Files with no collection assigned group under an "Uncategorized" header.
- Nice-to-have: each collection header shows an aggregate active/inactive state (all-active / all-inactive / mixed) with a bulk toggle to activate/deactivate every file in that collection at once.

**Priority**
Decluttering the file list is the primary goal; the bulk toggle is secondary.

just-jbc:
Closes #1119
Closes #1193

## What this does
Two related KB panel improvements, shipped together since the second builds on the first:

1. **Per-file active/inactive toggle** (#1119) — lets you exclude a specific KB file from search results without deleting it.
2. **Group personal uploads by collection** (#1193) — collapses personal KB uploads into collapsible per-collection header rows (with an "Uncategorized" bucket) to cut down on file-list clutter. Each collection header also exposes a bulk active/inactive toggle to flip every file in that collection at once — secondary to the decluttering, and derived (all-active/all-inactive/mixed) rather than independently stored, since there's no first-class "collection" entity in this codebase.

## Found and fixed during development
While building the grouping feature, found and fixed three pre-existing bugs in the collection-assignment dropdown (CollectionCombobox), unrelated to grouping itself but only surfaced by it:
- Popover was clipped by the table cell's truncation/overflow-hidden styling
- Popover was also clipped by the table wrapper's implicit overflow-y
- The filter query was pre-seeded with the current value and never cleared, so the option list only showed "Uncategorized" + the current tag until the user typed

Fixed by adding a `noTruncate` opt-out on the affected column and rendering the popover through a React portal positioned via getBoundingClientRect, plus tracking whether the user has typed since opening before filtering.

## Testing
Verified live against real KB content (existing files untouched) and synthetic test uploads: per-file toggle works, grouping/collapsing works, bulk toggle works, dropdown fix confirmed on the two previously-broken files, no regressions to citations (#1180).


chriscrosstalk:
Thanks for this — the feature is solid, and the retrieval filter being a denylist rather than an allowlist is the right call; it means search correctness doesn't depend on the backfill having run.

The branch had drifted onto a stale base, so GitHub was showing 216 changed files where the real feature is 13. Rather than ask you to rebase 60+ commits, I've cherry-picked both of your commits onto a fresh branch off `dev` with your authorship intact: #1286. Your `userTyped` fix came along with it — `CollectionCombobox` was rewritten on `dev` in the meantime, so that one is grafted onto the newer version.

Closing this in favour of that PR. Much appreciated.

Issue #127, opened 2026-02-02
Curated collection URLs can become stale and break downloads


chriscrosstalk:
## Problem

Curated collection JSON files (`kiwix-categories.json`, `kiwix.json`, `wikipedia.json`) contain hardcoded download URLs that can become stale when Kiwix updates their file catalog. This causes:

1. **Download failures** - URLs return 404 when files are rotated/renamed
2. **Silent failures** - Users don't know why certain content didn't download
3. **Cascade failures** - A broken URL shouldn't prevent other valid downloads from completing

### Example
URLs like `devdocs.io_en_python_2026-01.zim` fail with 404 because that specific dated version doesn't exist yet on Kiwix servers.

### Phase 1: Graceful Error Handling (Quick Fix)
Ensure that when a curated download URL fails (404, timeout, etc.), it:
- Logs the failure clearly
- Continues processing other downloads in the queue
- Reports failed downloads to the user in the UI (not silently swallowed)

Currently, failed downloads may not be clearly communicated to users about which specific items failed and why.

### Phase 2: Automated URL Validation (Short-Term)
Create a GitHub Action that runs weekly to:
1. Parse all URLs from curated collection JSON files
2. HEAD request each URL to verify it returns 200
3. Query Kiwix catalog API for the latest version of each file
4. Auto-create a PR if URLs need updating
5. Open an issue if URLs are broken and no replacement found

This catches staleness proactively before users encounter it.

### Phase 3: Dynamic URL Resolution (Long-Term Best Solution)
Instead of hardcoding full URLs:
```json
"url": "https://download.kiwix.org/zim/devdocs/devdocs.io_en_python_2026-01.zim"
```

Store metadata patterns and resolve URLs at download time:
```json
{
  "source": "kiwix",
  "category": "devdocs", 
  "name_pattern": "devdocs.io_en_python",
  "latest": true
}
```

Then NOMAD queries the Kiwix catalog API (`https://library.kiwix.org/catalog/v2/entries`) at download time to resolve the actual latest URL.

**Benefits:**
- Always uses current files - self-healing
- No manual maintenance of URLs
- Graceful degradation if API unavailable (fall back to cached/hardcoded URL)
- Could show users "newer version available" notifications

**Considerations:**
- Requires Kiwix API availability at download time
- Need fallback strategy for offline scenarios
- Cache resolved URLs to avoid repeated API calls

## Affected Files
- `collections/kiwix-categories.json`
[...]

PR #1217, opened 2026-08-06
ci(collections): add weekly schedule trigger to url validation

### LINKED ISSUE #1216: [Feature]: Add a scheduled trigger to collection URL validation
**Feature Category:** Content Management (ZIM, Maps, Collections)

**Problem Statement:**
Collection URLs in `collections/kiwix-categories.json` and `collections/wikipedia.json` go stale when Kiwix rotates a dated filename or openZIM renames a resource, and nothing catches it proactively. `validate-collection-urls.yml` checks every URL but only runs on push/pull_request for `collections/**.json`, so if nobody's touching those files, drift just sits there until a user hits a 404.

This has happened repeatedly: #127, #269, #358, #345, and #760 are all past issues from the same cause, and #1171 is the most recent one still open. #127 actually proposed a weekly check for this back in February. Only the push/PR trigger ended up shipping.

**Proposed Solution:**
Add a weekly `schedule` trigger to `validate-collection-urls.yml` alongside the existing push/PR ones. The validation logic itself is fine (recently hardened in #1188), this just makes it also run on a timer so drift gets caught within a week instead of whenever someone next touches those files.

**Alternative Solutions:**
#1045's live-catalog resolution at download time helps with date rotation and is a real improvement, but it doesn't check what's actually committed to the repo, and doesn't cover identifier renames (see #1171). A scheduled check complements that, doesn't replace it.

**Use Case:**
Right now, drift isn't caught until a user hits a 404 and files a bug report. A weekly check would catch it within days, so fixes like #1148 and #1189 become routine maintenance instead of reactive firefighting.

**Who would benefit:** Developers/Contributors, All Users

**Priority:** Medium

**Implementation Ideas:** Add a weekly cron schedule to the workflow's existing `on:` block. Could later extend to auto-filing an issue on failure, but keeping this scoped to just the trigger keeps it easy to review.

**Additional Context:** One thing worth flagging: scheduled triggers only run from the workflow file on the repo's default branch (main), not dev. PR #1188 ran into this same split for the curl fix in this file. So a PR against dev alone won't actually start running on schedule until it also reaches main. Wanted to call that out now rather than have it be a surprise later.

**Willing to contribute:** Yes, I can write the code. Have a fix ready to open right after this.

cclavin:
## Summary

Adds a weekly `schedule` trigger to `validate-collection-urls.yml`, alongside the existing `push`/`pull_request` triggers.

Closes #1216

## Why

The workflow only runs when `collections/**.json` changes. If nobody edits those files, a URL going stale after merge (Kiwix rotating a dated filename, openZIM renaming a resource) is never re-checked until a user hits a 404 and files a bug report. #127 proposed a weekly check for exactly this reason back in February; only the push/PR trigger ended up shipping.

## Changes

- `.github/workflows/validate-collection-urls.yml`: add a weekly cron schedule (`17 6 * * 1`, Monday morning UTC, offset from the top of the hour) to the existing `on:` block. No change to the validation logic itself, which already checks every URL correctly via a range request (hardened recently in #1188).

## Testing

Verified the cron syntax and confirmed the existing push/PR triggers are unchanged. Since this only adds a new trigger condition, there's no functional change to test beyond the workflow still parsing and running as before.

## Note on branches

This targets `dev` per CONTRIBUTING.md, but scheduled triggers only run from the workflow file on the default branch (`main`). Same situation #1188 had to solve for the curl fix in this same file. Called this out in #1216 so it's not a surprise. Happy to open a follow-up against `main` once this lands, or let me know if you'd rather handle that port yourselves.

### DIFF
--- a/.github/workflows/validate-collection-urls.yml
+++ b/.github/workflows/validate-collection-urls.yml
@@ -7,6 +7,8 @@ on:
   pull_request:
     paths:
       - 'collections/**.json'
+  schedule:
+    - cron: '17 6 * * 1'
 
 jobs:
   validate-urls:


chriscrosstalk:
Reviewed and approved.

This is the right fix for a problem we keep paying for. The catalog only gets validated when someone edits it, so a URL that goes stale after merge is found by a user hitting a 404 rather than by us. That is exactly how the four dead Wikipedia packages in #1189 and the renamed `preppingfood` in #1187 both reached production.

Checked that the workflow currently passes on `main`, so the schedule will run green rather than start a weekly red X.

jakeaturner (approved):

Issue #1050, opened 2026-06-25
[Bug]: Modified data path not respected by admin console


JakeShelf:
### Issue Category

Content Downloads (ZIM, Maps, Collections)

### Bug Description

Modifying the location of the data files (a new feature in update 1.33.0) for the information library is not reflected or respected by the admin console. Either for the rescan of the library or for the installation of additional resources from the content manager.

Excellent feature addition, but some linking back to admin tools seems to be missing.

### Steps to Reproduce

Shut down the Information Library
Move zim storage location outside opt/project-nomad
Update path for /data in Information Library settings: <server path> /data location.
Start Information Library
Access to existing material works flawlessly.
Add additional information/rescan library, the change is not recognised, and new material downloads into the old opt/project-nomad/storage/zim path.

Shut down Nomad, change compose.yml data path references for new location (as suggested for prior installs in non-standard installs), restart, no change in behaviour. 

### Expected Behavior

Admin panel to check/use the container path for refresh and new material downloads and install.

### Actual Behavior

Downloads and installs to the old path and is not recognised by the Information Library in the content provided.

### N.O.M.A.D. Version

1.33.0 on Linux Mint 22.3

### Operating System

Other Debian-based

### Docker Version

Docker 29.6.0

### Do you have a dedicated GPU?

No

### System Specifications

Xeon® D-1518
32GB RAM
11TB for Nomad.

### Relevant Logs

```shell

```

### Browser Console Errors (if UI issue)

```javascript

```

### Pre-submission Checklist

- [x] I have searched for existing issues that might be related to this bug
- [x] I am running the latest version of Project N.O.M.A.D. (or have noted my version above)
- [x] I have redacted any personal or sensitive information from logs and screenshots
- [x] This issue is NOT related to running N.O.M.A.D. on an unsupported/non-Debian-based OS

chriscrosstalk:
Good catch, and thanks for the precise steps. You've found a real gap in the new data-location feature: reading existing content from the moved path works, but the rescan and the new-download logic are still resolving to the old `/opt/project-nomad/storage/zim` target instead of the path you set. So the setting isn't being carried through to the download and library-rescan side.

We'll treat this as a bug to fix rather than a config workaround. Could you attach the Debug Info bundle from Settings so we capture exactly what path the app has stored versus where it's writing? That'll help us make sure the fix covers both the rescan and the content-manager install paths you mentioned.

JakeShelf:
[...]

PR #1227, opened 2026-08-11
fix(kb): stop leaking Qdrant vectors when a ZIM is deleted or replaced

### LINKED ISSUE #1170: Knowledge base retains Qdrant vectors for ZIM files that were deleted or replaced

just-jbc:
Closes #1170

Went through the three gaps you laid out and all three still check out against current `dev`:

1. `ZimService.delete()` never touches Qdrant - it removes the file, the kiwix library entry, and the `InstalledResource` row, but that's it.
2. `reconcileReplacedContentFile`'s case 5 is still a documented no-op ("NOT auto-reaped here").
3. `scanAndSyncStorage` only ever asks "is this file on disk already embedded?" - never the reverse.

## What this does

- Adds a reverse sweep to `scanAndSyncStorage`: diffs `sourcesInQdrant` against `embeddableFiles` and purges anything on the Qdrant side with no file backing it. No-ops if the disk scan comes back empty, so a transient failure can't get misread as "wipe everything."
- Tightens `ZimService.delete()` to purge the file's points immediately instead of waiting on the next sync.
- New `RagService.purgeIndexedSource()`, which is just the points-delete + state-row-remove pair that was already duplicated between `reconcileReplacedContentFile` and `deleteFileBySource` - now shared by both the sweep and the delete path.

## One thing I want your eyes on

While building this I found a real bug in my own first pass: `scanAndSyncStorage` calls `discoverNomadDocs(true)` right before the sweep runs, and that embeds `README.md`/`docs/*` under sources that `_discoverKbFiles()` never scans (it only walks `kb_uploads`/`zim`). Without an exclusion, the sweep would've flagged every docs source as an orphan and wiped the docs KB on the very first sync it ran. Fixed by excluding those two roots, pulled into one `_nomadDocsRoots()` helper so the paths aren't hardcoded twice.

~~Judgment call I made and would like your take on: if purging one orphan fails mid-sweep, it logs and moves to the next one instead of aborting the whole sync.~~ **Superseded by the batching follow-up commit below** - orphans are now purged as one batch, so this per-orphan resilience no longer applies: either the whole batch's delete succeeds or it's logged and retried on the next sync, rather than partial per-source failure handling. Still doesn't block legitimate new files from getting embedded either way, since that's a separate step later in `scanAndSyncStorage`.

## Testing

Ran this against my own live install instead of just the unit tests - real Qdrant, real disk, a ~2.28M-point KB:

- Faceted `nomad_knowledge_base` and got 29 distinct sources, 9 of which were docs/README - confirms the docs-collision issue above wasn't hypothetical, my own instance already had them indexed.
- Ran the actual `decideOrphans()` against that real data (read-only) and it found exactly one real orphan, `devdocs_en_bash_2026-01.zim` (102 points, left over from a since-superseded download), correctly leaving the other 27 alone.
- Deleted those 102 points for real and dropped the (already-absent) state row. Point count went 2,280,860 -> 2,280,758, exactly -102, everything else - the current zim, the docs, the uploads - untouched.

`npm run typecheck` is clean and unit tests pass, including a new `kb_orphan_decision.spec.ts` covering `decideOrphans`, with the empty-disk-scan guard as its own case.

## Follow-up commits (post-review)

Five more commits on top of the above, from a self-review pass:

- **Guard `ZimService.delete()`'s purge on Qdrant being installed.** It was calling `purgeIndexedSource()` unconditionally, so an install without the AI Assistant feature logged a false "Qdrant offline" error on every single ZIM delete. Now checks `SERVICE_NAMES.QDRANT` first, same as `reconcileReplacedContentFile()` already did.
- **Pointed `reconcileReplacedContentFile()` at `purgeIndexedSource()`.** It still had its own inline copy of the points-delete + state-row-remove pair, contradicting what I said above about that duplication being unified - there were three copies instead of one. Now there's one.
- **Reordered `purgeIndexedSource()` to remove the state row before the Qdrant points, and made it self-healing.** Orphan detection only looks at what's still in Qdrant, so deleting points first meant a failure between the two steps left a ghost `KbIngestState` row that nothing would ever revisit. Flipping the order means a failure instead leaves points in Qdrant, which the next sync's reverse sweep will find and retry - both steps are already idempotent, so a retry from either partial state is safe.
- **Switched the orphan sweep from a denylist to an allowlist.** It was excluding known non-scanned roots (README/docs) rather than allowlisting the roots `_discoverKbFiles()` actually scans (`kb_uploads`, `zim`) via a new shared `_kbScanRoots()` helper. The denylist form would've silently treated any future source root outside `kb_uploads`/`zim` as an orphan the first time it showed up - the exact bug class the denylist itself was written to fix for docs.
- **Batched the orphan sweep's deletes.** It was purging orphans one at a time, each a separate Qdrant round trip. Added `purgeIndexedSources()` / `_deletePointsBySources()` using Qdrant's `match.any` filter plus a single `whereIn()` query, so a sweep with dozens of orphans (e.g. after a bulk content change) does one round trip instead of dozens.
- **Extracted the allowlist check into its own tested function.** The `kb_uploads`/`zim` root check from the previous commit was inline in `scanAndSyncStorage` and untested. Pulled it into `filterOrphanCandidates()` in `kb_orphan_decision.ts`, next to `decideOrphans()`, and added unit tests - including one for the sibling-directory false positive a prefix check without a trailing separator would let through (e.g. `zim-backup/` matching `zim`). Also fixed two prettier violations from earlier follow-up commits that my per-edit lint checks missed.

For these five, `npm run typecheck` and the unit suite are clean (same pre-existing failures on `dev`, unrelated to these files), but I haven't re-run them against a live instance the way I did for the first two commits - flagging that gap rather than claiming coverage I don't have.


chriscrosstalk:
Sorry for the wait on this. Read the whole thing including the five follow-up commits, and the self-review pass caught more than most reviews would have.

Your docs-exclusion call was right, and the allowlist reasoning is the part I'd have pushed for. Denylisting README/docs by name fixes today's collision and quietly re-arms the same trap for whatever source root gets added next. The state-row-before-points ordering is right for exactly the reason you give, and the sibling-directory case is a good test to have written. The superseded judgment call reads as resolved to me, batching it away was the better answer.

I pushed one commit to your branch rather than sending you round again, since it's the same bug class you already found, one level down.

`decideOrphans` guards on the disk scan coming back empty, which is a whole-scan guard. The failure is per-root. `_discoverKbFiles()` walks kb_uploads and zim in a loop and skips a missing root rather than failing, because a fresh install legitimately has no kb_uploads yet. So if the zim root is relocated or not yet mounted, the scan still returns a non-empty list from kb_uploads, the guard doesn't fire, and `filterOrphanCandidates` can't tell a root that scanned clean from one that was never there. Every ZIM source passes the prefix test and the batch delete takes out the whole ZIM side of the index in one call.

That's reachable today. Relocating the storage path is a known rough edge (#1050) and the usual symptom is the admin resolving a different physical directory than the content is in. Which means the person most likely to hit it is the person whose knowledge base already looks wrong, pressing Sync to fix it.

The fix follows what you'd already built: `_discoverKbFilesWithRoots()` reports the roots it actually walked, and `filterOrphanCandidates` allowlists against those instead of the configured ones. A root that wasn't there contributes nothing; a root that scanned clean and held no files still yields its orphans, so the real purge path is unchanged. Your empty-scan guard becomes the degenerate case of the same rule.

One other thing: the unit-test paths were POSIX string literals, so the `filterOrphanCandidates` cases passed on Linux CI and failed on Windows, where the separator is a backslash and no source ever matches a prefix. Rebuilt them with `join()` so they run anywhere, including on the machine someone is most likely to be using while editing this file.

Your note about not re-running the last five commits against a live instance is fair, and it now covers mine too. A single live run over a delete and a sync would close it out.

just-jbc:
Thanks again for catching and fixing this — appreciate you pushing the commit directly rather than sending it back.

I'm stepping back from active development on this for now, so I won't be the one running the live delete-and-sync verification. Wanted to flag that clearly rather than leave it silently waiting on me — happy for you or anyone else to pick it up whenever convenient. Really appreciated the collaboration on this one.

jakeaturner (approved):

PR #1239, opened 2026-08-13
feat(qol): admin and app feature implementation

### LINKED ISSUE #1237: [Feature]: QOL add existing docker container
### Feature Category

UI/UX Improvement

### Problem Statement

I have existing running docker containers that i want to integrate into the UI but I don't want nomad to necessarily manage it.

### Proposed Solution

Add button to supply depot that allows the user to add an app that will be placed in the home screen grid. Should not be taken over by nomad only displayed.

### Alternative Solutions

_No response_

### Use Case

Any user that already has some of the featured apps installed and wants to use their existing containers.

All Users

_No response_

### Examples or References

_No response_

Yes - I can write the code

### Additional Context

Done, submitting a pull request with this as one of the features.

### Pre-submission Checklist

- [x] This feature aligns with NOMAD's mission of offline-first knowledge and education

### LINKED ISSUE #1238: [Feature]: QOL admin only features, lock guest users out of admin components
### Feature Category

New Service/Tool Integration

### Problem Statement

no user segmentation

### Proposed Solution

add a single admin user to remove unrestricted access

### Alternative Solutions

_No response_

### Use Case

more than one person is using this system. don't want everyone to be able to delete apps 

All Users

_No response_

### Examples or References

_No response_

Yes - I can write the code

### Additional Context

In my forked repo

### Pre-submission Checklist

- [x] This feature aligns with NOMAD's mission of offline-first knowledge and education

the1daBread:
Closes #1237 
Closes #1238 

Changes directly implement the above requests that I made. 

#1237 (Existing App)
Adds button to the supply depot for existing docker containers.

#1238 (Admin Login)
Adds a login button on the top left of the home screen. Login is cached locally as a session cookie.
Uses the following environment variables in the docker-compose yaml file
      - ADMIN_USER=admin
      - ADMIN_PASS=replaceme


chriscrosstalk:
Thanks for both of these, and for opening issues before writing code. I want to give you a real answer rather than a one-line close.

Taking the second one first: we are not going to take the admin login. Authentication is on our roadmap, and when we do it we are going to build that piece ourselves so it fits the rest of the product. That is not a reflection on your implementation. It is a decision about who owns that surface.

On surfacing existing containers, I dug into it properly and we are going to solve the underlying need a different way, so I am declining this half too. The reasoning, in case it is useful:

What you are actually asking for, and what #1237 describes, is that things you already run should appear on the NOMAD dashboard. Adoption via the Docker daemon is one way to get there, but it can only ever reach containers on the same Docker host, and it has to guess. Reading the code, the link is built from the lowest published host port with an assumed http scheme. That picks the wrong port for anything publishing more than one, Portainer publishes 8000 and 9443, Jellyfin 8096 and 8920, and it produces a broken link for anything serving HTTPS.

The direction we are going instead is a user-added link tile: give it a name, a URL, and an icon, and it appears on the dashboard. It solves a strictly larger version of your problem, because it also covers a NAS, a router, a printer, or a machine that is not running Docker at all. There is no guessing involved, since you supply the address. And it has no lifecycle to manage, so it cannot break in ways neither of us can reproduce.

Worth saying plainly: if what you want is real Docker management rather than dashboard presence, you can add Portainer through the existing custom app form today, and it will do that far better than anything we would build.

I have written the link tile up on our roadmap. Closing both of these, but the need you identified is the one we are building for, and you identified it clearly. Thanks for that.

Issue #1185, opened 2026-08-01
[Bug]: ZIM ingestion is O(n²) — every batch re-scans the archive from entry 0, making large ZIMs unfinishable


PairsOfTwoSocks:
### Summary

`ZIMExtractionService.extractZIMContent()` re-opens the archive and re-scans it from entry 0 on **every** batch, so the cost of batch *k* is proportional to *k*. Ingestion of a single ZIM is therefore **O(n²)** in the number of article entries.

For small ZIMs this is invisible. For `wikipedia_en_all_maxi` (18,982,214 entries) it makes ingestion effectively unfinishable: on my box it ran **13.7 days to reach 5.83% of the corpus**, which is only ~1% of the total work, with a projected ~482 days of continuous CPU remaining.

The embedding half of the pipeline is fine — this is purely the extraction seek.

### Root cause

`app/services/zim_extraction_service.js` (lines 45–61):

```js
const archive = new Archive(filePath);          // L45  fresh archive every batch
...
const startOffset = opts.startOffset || 0;      // L52
const batchSize = opts.batchSize || (opts.maxArticles || Infinity);
for (const entry of archive.iterByPath()) {     // L54  always starts at entry 0
    if (articlesSkipped < startOffset) {
        if (this.isArticleEntry(entry) && !processedPaths.has(entry.path)) {
            articlesSkipped++;                  // L57  advance one entry at a time
        }
        continue;
    }
    if (articlesProcessed >= batchSize) break;
    ...
}
```

`EmbedFileJob` dispatches each continuation with `batchOffset = previous + articlesProcessed` and `ZIM_BATCH_SIZE = 50`, so to reach offset *X* the service walks *X* entries first.

The skip is not a cheap counter increment, because `isArticleEntry()` dereferences the item to read its mimetype:

```js
isArticleEntry(entry) {
    if (entry.isRedirect) return false;
    const item = entry.item;        // <-- dereferenced for every skipped entry
    const mimeType = item.mimetype;
    return mimeType === 'text/html' || mimeType === 'application/xhtml+xml';
}
```

The design constraint underneath this: **`startOffset` is defined over the filtered subsequence of article entries**, not over raw entry index. That forces a full classification pass over everything before the offset, which is what makes a cheap seek impossible in the current shape.

Total skip work for a ZIM of *n* article entries at batch size *B* is `n²/2B` entry dereferences — for full-English Wikipedia at B=50, ~3.6 × 10¹² of them.

[...]

PR #1242, opened 2026-08-14
fix(rag): don't abandon a ZIM after one sparse batch

### LINKED ISSUE #1240: ZIM ingestion silently stops after the first batch with sparse text content
## Summary

Batched ZIM ingestion stops after the first batch whose articles mostly contain no extractable text, silently skipping the rest of the archive. The file is then recorded as `indexed`, so nothing surfaces the gap.

## Cause

In `RagService.processZIMFile`:

```ts
const articlesInBatch = new Set(zimChunks.map((c) => c.documentId)).size
const hasMoreBatches = articlesInBatch >= ZIM_BATCH_SIZE
```

`zimChunks` only contains articles that produced a non-empty chunk — `ZIMExtractionService` filters empty chunks before returning. Redirect stubs, category listings, and media/PDF wrapper pages are ordinary ZIM entries that yield no text, so they contribute no `documentId`. A batch that consumed a full 50 articles but produced text for only 10 evaluates `10 >= 50` → `false`, and no continuation is dispatched.

The same value is returned as `articlesProcessed`, which `EmbedFileJob` uses to advance the offset (`nextOffset = batchOffset + result.articlesProcessed`). A sparse batch therefore also re-reads its own overlap. The two defects mask each other in current code — the continuation only runs when the count is exactly `ZIM_BATCH_SIZE`, which is also the only case that advances correctly — so fixing either one alone is not safe. Fixing `hasMoreBatches` alone would let a zero-text batch advance the offset by zero and re-run the same window indefinitely.

## Impact

Measured on a v1.34.0 install (35 ZIMs, ~2M chunks). Archives with dense text — Stack Exchange, iFixit, Gutenberg, WikiMed, MedlinePlus, Libre Pathology — are unaffected, because for them articles-with-content and articles-consumed are the same number. Scraped-site archives, where navigation and asset pages are interleaved with content, are hit hard:

| Archive | Articles read before stopping | Total ingestible | Chunks |
|---|---|---|---|
| `www.ready.gov_en` | 50 | 2,437 | 231 |
| `wwwnc.cdc.gov_en_all` | 50 | 1,878 | 177 |
| `cd3wdproject.org_en_all` | 150 | 13,297 | 1,181 |

After the fix, the same three archives produce 10,684 / 15,529 / 96,830 chunks — 46×, 88×, and 82× more. CDC Travelers' Health alone went from 177 chunks to 15,529; 1,875 of its 1,878 HTML entries extract cleanly at an average of 10,917 characters.

## Reproduce

1. Download a curated ZIM whose leading entries by path are non-content (e.g. Ready.gov or CDC Travelers' Health from Medicine/Survival tiers).
2. Let auto-ingest run to completion.
3. `[EmbedFileJob] Successfully embedded N chunks ... (final batch)` appears after a single batch.
4. `kb_ingest_state` shows `indexed`, and Qdrant holds a few hundred chunks for an archive with thousands of articles.

## Proposed fix

Gate the continuation and the offset advance on the count of articles the extractor actually consumed. `extractZIMContent` already computes and logs `articlesProcessed` but does not return it. End-of-archive is then `articlesProcessed < batchSize` (the iterator ran dry).

Note `archive.articleCount` cannot serve as the bound — `iterByPath()` legitimately yields more article entries than that figure for some archives, and for others `articleCount` counts redirects that `isArticleEntry` skips (WikiMed: 360,628 `articleCount` vs 101,665 non-redirect HTML entries).

Happy to submit a PR — I have the change and a unit test ready.

rcorvus:
## TLDR

When NOMAD indexes a ZIM file for AI search, it works through the file 50 articles at a time and asks after each batch: "was there more?" It answered that question by counting how many articles produced *text* — but plenty of pages in a ZIM legitimately have none (redirects, category listings, pages that just wrap a video or a PDF). So the first time a batch of 50 pages happened to be mostly those, NOMAD decided it had reached the end of the file and quietly stopped. The file was still marked "indexed", so nothing looked wrong.

On my install this meant CDC Travelers' Health was indexed from 50 of its 1,878 pages. Ready.gov and CD3WD were the same story. The AI simply couldn't see almost any of that content, and there was no error to go looking for.

The fix is to count the articles NOMAD actually *read* instead of the ones that happened to have text. Same three files afterward: **46×, 88×, and 82× more content indexed.**

Closes #1240

---

## The bug

`RagService.processZIMFile` gated the batch continuation on chunk-derived data:

```ts
const articlesInBatch = new Set(zimChunks.map((c) => c.documentId)).size
const hasMoreBatches = articlesInBatch >= ZIM_BATCH_SIZE
```

`zimChunks` holds only articles that produced a non-empty chunk — `ZIMExtractionService` filters empties before returning. A batch that consumed a full 50 articles but produced text for 10 evaluates `10 >= 50` → `false` and never dispatches a continuation.

The same value was returned as `articlesProcessed`, which `EmbedFileJob` uses to advance the offset:

```ts
const nextOffset = (batchOffset || 0) + (result.articlesProcessed || 0)
```

**The two defects mask each other, so they have to be fixed together.** The continuation only ran when the count was exactly `ZIM_BATCH_SIZE`, which is also the only case that advanced the offset correctly — so no overlap re-reads occur in current code, and I confirmed that empirically (duplicate rate 0/8,000 in WikiMed, 3/8,000 in MedlinePlus). But fixing `hasMoreBatches` alone would let a zero-text batch advance the offset by zero and re-run the same window forever.

## The fix

`extractZIMContent` already computes and logs `articlesProcessed` but discards it; it now returns it. Both the continuation gate and the offset advance use it.

End-of-archive is `articlesProcessed < batchSize` — the iterator ran dry. `archive.articleCount` deliberately isn't used as the bound: `iterByPath()` yields more article entries than that figure for some archives, and for others it counts redirects that `isArticleEntry` skips (WikiMed reports 360,628 but has 101,665 non-redirect HTML entries).

A full final batch costs one extra dispatch that extracts nothing and stops. That's intentional — over-running by one empty batch is cheap; stopping one batch early loses the rest of the archive.

The decision is extracted into `app/utils/zim_batch_decision.ts` to match the existing `kb_ingest_decision.ts` / `content_reindex_decision.ts` pattern, with unit tests covering both regressions.

The log line now reports both counts (`16/50 articles`) so a sparse batch is visible without re-deriving it.

## Results

Measured on a v1.34.0 install, same hardware, `force: true` re-embed:

| Archive | Before | After | |
|---|---|---|---|
| `cd3wdproject.org_en_all` | 1,181 | 96,830 | 82× |
| `wwwnc.cdc.gov_en_all` | 177 | 15,529 | 88× |
| `www.ready.gov_en` | 231 | 10,684 | 46× |
| `librepathology_en_all_maxi` | 16,174 | 16,174 | unchanged (was already complete) |
| `medlineplus.gov_en_all` | 114,740 | 114,741 | unchanged (was already complete) |

The last two are the control: dense-text archives where articles-with-content and articles-consumed are the same number were never affected, and re-running reproduces their counts near-exactly. Stack Exchange, iFixit, Gutenberg and WikiMed are all in that category.

## Testing

- `tests/unit/zim_batch_decision.spec.ts` — 6 cases, including named regressions for the LibreTexts (sparse batch) and media-archive (zero-text batch) shapes
- `npx tsc --noEmit` clean
- Verified live: the re-run walked all 23,171 articles of a previously-truncated archive across 464 batches instead of stopping at 50

**On the RAG eval harness** (CONTRIBUTING step 4, added in #1233): I have not included eval numbers. My install runs the v1.34.0 image, which predates the harness, so I couldn't run `eval:retrieval` / `eval:generation` against the deployed build. I'd also expect them to be flat by construction — this changes *how much of a file gets ingested*, not chunking, retrieval, reranking, thresholds, prompts, or context assembly, and the eval corpus is dense-text (the shape this bug never affected). The real-world metric is the table above. Glad to run the harness if you'd like it before merging; just let me know.

## Note for maintainers

Existing installs stay truncated after upgrading — the fix changes ingestion, not stored state. Affected files need a re-embed with `force: true` to pick up the missing content. Worth a release note, and possibly a one-time reconcile, since `kb_ingest_state` currently reports these files as `indexed` with no indication that they're partial.

### DIFF
--- a/admin/app/services/rag_service.ts
+++ b/admin/app/services/rag_service.ts
@@ -37,6 +37,7 @@ const EXCLUDE_EVAL_FILTER = {
 import type { KbIngestStateValue } from '../../types/kb_ingest_state.js'
 import { ZIMExtractionService } from './zim_extraction_service.js'
 import { ZIM_BATCH_SIZE } from '../../constants/zim_extraction.js'
+import { hasMoreArticleBatches } from '../utils/zim_batch_decision.js'
 import { EMBEDDING_MODEL_NAME } from '../../constants/ollama.js'
 import { ProcessAndEmbedFileResponse, ProcessZIMFileResponse, RAGResult, RerankedRAGResult } from '../../types/rag.js'
 
@@ -556,10 +557,14 @@ export class RagService {
       `[RAG] Extracting ZIM content (batch: offset=${startOffset}, size=${ZIM_BATCH_SIZE})`
     )
 
-    const { chunks: zimChunks, totalArticles } = await zimExtractionService.extractZIMContent(
-      filepath,
-      { startOffset, batchSize: ZIM_BATCH_SIZE }
-    )
+    const {
+      chunks: zimChunks,
+      totalArticles,
+      articlesProcessed,
+    } = await zimExtractionService.extractZIMContent(filepath, {
+      startOffset,
+      batchSize: ZIM_BATCH_SIZE,
+    })
 
     logger.info(
       `[RAG] Extracted ${zimChunks.length} chunks from ZIM file with enhanced metadata (file totalArticles=${totalArticles})`
@@ -610,15 +615,19 @@ export class RagService {
       }
     }
 
-    // Count unique articles processed in this batch. hasMoreBatches gates on the article
-    // count — zimChunks.length counts section-level chunks (multiple per article under the
-    // 'structured' strategy), so comparing it to ZIM_BATCH_SIZE (an article limit) caps
-    // processing at the first batch for any real archive.
-    const articlesInBatch = new Set(zimChunks.map((c) => c.documentId)).size
-    const hasMoreBatches = articlesInBatch >= ZIM_BATCH_SIZE
+    // Gate the continuation on articles the extractor CONSUMED, not on articles that
+    // produced chunks. Articles whose text is empty after cleaning (redirect stubs,
+    // category pages, media/PDF wrappers) contribute no documentIds, so a chunk-derived
+    // count reads a normal sparse batch as end-of-archive and silently abandons the rest
+    // of the file. See `zim_batch_decision.ts` for the full rationale.
+    const articlesWithContent = new Set(zimChunks.map((c) => c.documentId)).size
+    const hasMoreBatches = hasMoreArticleBatches({
+      articlesProcessed,
+      batchSize: ZIM_BATCH_SIZE,
+    })
 
     logger.info(
-      `[RAG] Successfully embedded ${totalChunks} total chunks from ${articlesInBatch} articles (hasMore: ${hasMoreBatches})`
+      `[RAG] Successfully embedded ${totalChunks} total chunks from ${articlesWithContent}/${articlesProcessed} articles (hasMore: ${hasMoreBatches})`
     )
 
     // Only delete the file when:
@@ -640,7 +649,11 @@ export class RagService {
         : 'ZIM file processed and embedded successfully with enhanced metadata.',
       chunks: totalChunks,
       hasMoreBatches,
-      articlesProcessed: articlesInBatch,
+      // MUST be the consumed count: EmbedFileJob advances the next batch offset by this
+      // value. Reporting the content-bearing count instead re-reads the overlap on every
+      // sparse batch, and a batch that yields no text at all would advance the offset by
+      // zero and re-run the same window forever.
+      articlesProcessed,
       totalArticles,
     }
   }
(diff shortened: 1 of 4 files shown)


chriscrosstalk:
Verified and approved, nothing needed from you on this one.

Reproduced against v1.34.0 on our test box. Different ZIM build dates than yours so the absolute numbers shift a little, but the result is the same: CDC Travelers' Health 121 chunks to 12,434, Ready.gov 219 to 10,050, which is the same 46x you measured.

Your call to fix both defects together was the right one. Correcting the gate alone would have left a zero-text batch advancing the offset by zero and re-running the same window forever, and that would have been much harder to diagnose than the bug you started with.

One scope note so it doesn't come as a surprise later: this doesn't close #1185. That's the separate problem where each batch re-walks the archive from entry 0. I measured it while I was in here, and it's about 5µs per skipped entry, so it's invisible below roughly 100k articles and becomes hours of pure seeking at Wikipedia scale. Yours makes archives finish at all. That one only bites the very largest.

Queued for review.

jakeaturner (approved):
Excellent fix, thanks for the contribution!

PR #1167, opened 2026-07-27
chore(collections): surface FDA Drug Reference in Medicine > Standard

chriscrosstalk:
> **⛔ DO NOT MERGE UNTIL v1.34.0 GA IS PUBLISHED.** Draft deliberately. Merging early breaks v1.33.0 installs — see below.

## What

Adds the `openfda-drug-labels` entry to the **Medicine → Standard** tier. The entry is byte-identical to the one already on `rc`.

```json
{
  "id": "openfda-drug-labels",
  "type": "dataset",
  "version": "2025-01",
  "title": "FDA Drug Reference",
  "description": "Offline FDA drug labels - search by drug name",
  "url": "https://api.fda.gov/download.json",
  "size_mb": 1700
}
```

## Why it must wait for GA

The category manifest is fetched **live from `main`**, not baked into the image:

```ts
// admin/app/services/collection_manifest_service.ts
zim_categories: 'https://raw.githubusercontent.com/.../refs/heads/main/collections/kiwix-categories.json'
```

So this reaches **every install on every version the moment it merges** — not on their next upgrade.

`type: "dataset"` does not exist in v1.33.0. A grep across `admin/app`, `admin/inertia` and `admin/types` at tag `v1.33.0` returns **zero hits**; it is new in 1.34.0. On a v1.33.0 box the entry would render as an installable resource with no code that understands its type and a URL (`https://api.fda.gov/download.json`) that is not a ZIM.

## Why it must not be forgotten either

This is the discovery path for the Drug Reference. Without it, 1.34.0 ships a working, fully documented feature that cannot be found from Content Explorer — which is how QA found the gap in the first place.

A draft PR rather than a reminder or an issue, deliberately: it cannot be merged by accident, it is visible in the PR queue, and at release it is one click rather than a task someone has to remember and reconstruct.

## Verification

- JSON parses
- The resulting `medicine-standard` resource list is identical to `origin/rc`'s, confirmed by structural comparison rather than eyeballing
- No other tier or category touched

## Release checklist

1. Publish v1.34.0 GA
2. Un-draft and merge this
3. Confirm `GET /api/easy-setup/curated-categories` on a 1.34.0 box lists **FDA Drug Reference** under Medicine → Standard

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1243, opened 2026-08-14
fix(db): allow 'dataset' in installed_resources.resource_type

### LINKED ISSUE #1241: openFDA drug reference never records an installed_resources row (resource_type enum missing 'dataset')
## Summary

`IngestDrugDataJob`'s install-state write-back fails on every run. The openFDA drug labels ingest correctly, but no `installed_resources` row is ever created, so the Medicine category never resolves its Standard or Comprehensive tier.

## Cause

`installed_resources.resource_type` is still `enum('zim','map')` from the table's original migration (`1770849119787_create_create_installed_resources_table.ts`):

```ts
table.enum('resource_type', ['zim', 'map']).notNullable()
```

Everything else already accounts for `'dataset'`:

- `InstalledResource` declares `resource_type: 'zim' | 'map' | 'dataset'`
- `IngestDrugDataJob` writes `{ resource_id, resource_type: 'dataset' }`, and its comment notes the type "was widened onto the model in the foundational slice"
- `CollectionManifestService` queries `whereIn('resource_type', ['zim', 'dataset'])`

Only the column was left behind, so MySQL rejects the write:

```
Data truncated for column 'resource_type' at row 1
insert into `installed_resources` (`collection_ref`, `file_path`, `file_size_bytes`,
  `installed_at`, `resource_id`, `resource_type`, `url`, `version`)
values ('medicine', '/app/storage/drug-data', 1852880110, '2026-08-13 02:19:23',
  'openfda-drug-labels', 'dataset', 'https://api.fda.gov/download.json', '2026-08-12')
```

The job logs the failure at error level and returns normally, so the ingest looks successful.

## Impact

On a v1.34.0 install: `drug_labels` held 261,671 rows while `installed_resources` held zero rows for `openfda-drug-labels`. Because `getCategoriesWithStatus()` resolves a tier only when every resolved resource id is accounted for, and #1167 added the FDA Drug Reference to Medicine > Standard, the Medicine category could not resolve Standard or Comprehensive even with every other resource installed. `GET /api/zim/curated-categories` returned `installedTierSlug: null` for Medicine.

After widening the column and replaying the failed insert, the same endpoint returns `installedTierSlug: 'medicine-comprehensive'`.

## Reproduce

1. Install the Medicine tier that includes the FDA Drug Reference.
2. Let the drug data ingest finish — `drug_labels` fills correctly.
3. `SELECT * FROM installed_resources WHERE resource_id = 'openfda-drug-labels'` returns nothing.
4. `SHOW COLUMNS FROM installed_resources LIKE 'resource_type'` → `enum('zim','map')`.
5. Medicine shows `installedTierSlug: null`.

## Proposed fix

A migration widening the column to `enum('zim','map','dataset')`. No application code needs to change.

Happy to submit a PR — I have the migration ready and verified against a live install.

rcorvus:
## TLDR

The FDA Drug Reference downloads and ingests fine — all 261,671 drug labels land in the database. But the very last step, where NOMAD writes down "this resource is installed", fails every single time and always has. The database column that records what *kind* of resource it is only accepts `zim` or `map`, and the drug reference is a `dataset`.

The rest of the code was already written for `dataset` — the model, the job, the tier-status query all handle it. Only the column was missed.

The visible symptom: the Medicine category never shows as fully installed, even when everything in it is. Since #1167 put the FDA Drug Reference in Medicine > Standard, that tier can't complete without this row.

One-line migration to widen the column. No application code changes.

Closes #1241

---

## The bug

`installed_resources.resource_type` is still `enum('zim','map')` from the table's original migration:

```ts
// 1770849119787_create_create_installed_resources_table.ts
table.enum('resource_type', ['zim', 'map']).notNullable()
```

Everything else already accounts for `'dataset'`:

- `InstalledResource` declares `resource_type: 'zim' | 'map' | 'dataset'`
- `IngestDrugDataJob` writes `resource_type: 'dataset'`, and its own comment says the type "was widened onto the model in the foundational slice"
- `CollectionManifestService.getCategoriesWithStatus()` queries `whereIn('resource_type', ['zim', 'dataset'])`

So MySQL rejects the write:

```
Data truncated for column 'resource_type' at row 1
```

The job logs this at error level and returns normally, so the ingest reports success and the missing row goes unnoticed.

## Impact

On my v1.34.0 install: `drug_labels` had 261,671 rows, `installed_resources` had zero rows for `openfda-drug-labels`. `getCategoriesWithStatus()` resolves a tier only when every resolved resource id is accounted for, so Medicine returned `installedTierSlug: null` despite every other resource in the tier being installed.

## Verification

Against a live install, after running the migration:

```
SHOW COLUMNS FROM installed_resources LIKE 'resource_type';
  resource_type   enum('zim','map','dataset')   NO
```

Replaying the exact insert from the original error log now succeeds, and the tier resolves:

```
GET /api/zim/curated-categories
  medicine     installedTierSlug='medicine-comprehensive'   (was null)
  survival     installedTierSlug='survival-comprehensive'
  diy          installedTierSlug='diy-comprehensive'
  agriculture  installedTierSlug='agriculture-comprehensive'
```

## Notes

`down()` deletes `'dataset'` rows before narrowing the column — MySQL would otherwise truncate them to `''` or fail under `STRICT_TRANS_TABLES`.

I deliberately left `types/collections.ts` and `types/system.ts` at `'zim' | 'map'`. Those describe `ResourceUpdateCheckRequest` and `ContentAutoUpdateResourceStatus` — the remote update-check and content-auto-update flows, which a dataset doesn't participate in (openFDA tracks freshness via `drugReference.lastUpdatedExportDate`). Widening them would invite datasets into paths that don't handle them. Happy to revisit if you'd rather they were widened too.

## Note for maintainers

Existing installs that already ingested the drug data won't get the row retroactively — the migration only unblocks future writes. A one-time backfill, or letting the next drug-data refresh write it, would repair them. On my install I replayed the original insert with the values from the job (`version` = openFDA `export_date`, `file_path` = staging dir, `file_size_bytes` = total downloaded bytes).


jeffmcm64:
Reproduced on **N.O.M.A.D. 1.34.0 / Ubuntu 26.04**.

- `installed_resources.resource_type` was `enum('zim','map')`
- `drug_labels` already contained **261,991 rows**
- Medicine displayed only **Essential**
- Selecting **Comprehensive** produced the “Started Downloading” toast, but no download began

I tested the proposed fix by:

1. Widening `resource_type` to include `dataset`
2. Backfilling the missing `openfda-drug-labels` entry in `installed_resources`

After that, `/api/zim/curated-categories` immediately returned:

```json
"installedTierSlug": "medicine-comprehensive"
```

So the fix worked on my existing 1.34.0 installation.

This also appears to explain the user-facing symptom reported in #1264.

chriscrosstalk:
Good find, and thanks for quoting the actual truncation error, that made this a thirty second read. The `down()` handling the existing rows before narrowing is the right call too, most people skip that.

One thing before this goes in. This fixes the write path, but anyone who already ran the FDA ingest on 1.34.0 is sitting on 261k drug label rows and no `installed_resources` row, and nothing in the migration gives them one. They upgrade, the enum is fixed, and their Medicine tier is still stuck, because the job that would have written the row already ran and reported success.

Can you add a backfill to the same migration? Something like: if `drug_labels` has rows and there's no `openfda-drug-labels` row in `installed_resources`, insert one. The `file_path`, `url` and `version` values are all in the job already. That way the upgrade actually repairs the installs that hit this rather than just stopping new ones.

jakeaturner (approved):

PR #1245, opened 2026-08-14
fix(downloads): guard orphaned model jobs

### LINKED ISSUE #1194: fix(downloads): guard the model-downloads queue against orphaned jobs
## Problem

#1191 added a guard so an orphaned BullMQ job can no longer 500 `GET /api/downloads/jobs`. It covers three of the four queues that endpoint reads. The **model-downloads** queue is not covered, and still reproduces #1190 exactly.

`listDownloadJobs()` routes three queues through `fetchJobsWithStates()`, which carries the new filter (`download_service.ts:55`):

```ts
].filter(({ job }) => job?.id != null && job.data != null)
```

The model queue bypasses it — `modelQueue.getJobs(...)` is called directly at `download_service.ts:63`, and the mapper at `:101` is unguarded:

```ts
const modelDownloads = modelJobs.map((job) => ({
  jobId: job.id!.toString(),
  url: job.data.modelName || 'Unknown Model',
```

`getJobs()` returns `undefined` for an id whose payload hash is gone, so `job.id` throws before any `data` access.

## Steps to Reproduce

On a v1.34.0-rc.4 box:

1. `docker exec nomad_redis redis-cli ZADD bull:model-downloads:failed 1 orphan-model`
2. `curl -s -o /dev/null -w '%{http_code}\n' http://localhost:8080/api/downloads/jobs`

## Expected vs Actual

**Expected:** 200, the orphan skipped, real jobs still listed — the behaviour #1191 established for the other three queues.
**Actual:** 500 on every call. Observed 10/10, then 6/6 more **after `docker restart nomad_admin`**.

```
TypeError: Cannot read properties of undefined (reading 'id')
    at file:///app/app/services/download_service.js:93:24
    at Array.map (<anonymous>)
    at DownloadService.listDownloadJobs (download_service.js:92:42)
```

Removing only that orphan, while three other injected orphans remained in the guarded queues, restored 200. Failed jobs are never evicted, so this does not self-recover — same as #1190. Content Explorer polls this endpoint every ~3s, so the page never loads.

## Impact

Not currently live: `bull:model-downloads:failed` is empty on both test boxes, so no user is affected today. This is an unpatched instance of a bug class we have already shipped a fix for, reachable by the same mechanism as the filed one — a job id outliving its payload hash.

## Suggested Fix

Apply the same filter to `modelJobs`, or route the model queue through `fetchJobsWithStates()` like the other three so the guard is structural rather than repeated.

## Files Involved

- `admin/app/services/download_service.ts:63` — direct `getJobs()` call bypasses the guard
- `admin/app/services/download_service.ts:101` — unguarded mapper

Found during v1.34.0-rc.4 QA.

NgoQuocViet2001:
## Summary

- route the model-download queue through the existing guarded `fetchJobsWithStates()` path
- skip orphaned BullMQ model entries before mapping the downloads response

The model queue was the only path in `listDownloadJobs()` that bypassed the shared orphan guard. A retained orphan could therefore make `GET /api/downloads/jobs` return 500 on every poll.

Closes #1194

## Testing

- reproduced an orphaned `model-downloads` entry: upstream throws on the missing job, while this branch skips it
- `npm run typecheck`
- `npm run build`
- `git diff --check`

### DIFF
--- a/admin/app/services/download_service.ts
+++ b/admin/app/services/download_service.ts
@@ -56,11 +56,10 @@ export class DownloadService {
   }
 
   async listDownloadJobs(filetype?: string): Promise<DownloadJobWithProgress[]> {
-    const modelQueue = this.queueService.getQueue(DownloadModelJob.queue)
-    const [fileTagged, extractTagged, modelJobs, drugTagged] = await Promise.all([
+    const [fileTagged, extractTagged, modelTagged, drugTagged] = await Promise.all([
       this.fetchJobsWithStates(RunDownloadJob.queue),
       this.fetchJobsWithStates(RunExtractPmtilesJob.queue),
-      modelQueue.getJobs(['waiting', 'active', 'delayed', 'failed']),
+      this.fetchJobsWithStates(DownloadModelJob.queue),
       this.fetchJobsWithStates(DownloadDrugDataJob.queue),
     ])
 
@@ -98,7 +97,7 @@ export class DownloadService {
       }
     })
 
-    const modelDownloads = modelJobs.map((job) => ({
+    const modelDownloads = modelTagged.map(({ job }) => ({
       jobId: job.id!.toString(),
       url: job.data.modelName || 'Unknown Model',
       progress: parseInt(job.progress.toString(), 10),


chriscrosstalk:
Reviewed and approved.

Right call routing this through `fetchJobsWithStates` rather than adding a second guard. The model queue was the only one of the four still reading jobs raw, which is why a single orphan could 500 the endpoint on every poll.

Verified the helper fetches exactly the same states the raw call did (waiting, active, delayed, failed), so the only behaviour change is the orphan skip. Nothing drops out of the Downloads list.

NgoQuocViet2001:
Rebased onto current `dev` — the diff is back to one file, `+3 −4`.

`dev` was rebased and force-pushed on Sep 8, which re-wrote #1233 and #1244 under new shas. My branch still carried the old ones, so GitHub showed this PR as 71 files / +10865 and flagged it conflicting. Dropped both duplicates and kept only the fix commit.

Two things I checked rather than assumed, since a rebase here could quietly undo someone else's work:

- `retryFailedJob` now carries **dev's** version — the remove-then-dispatch ordering from 17916a73, not my branch's older text. Only the `listDownloadJobs` hunk is mine.
- `fetchJobsWithStates` on current dev still fetches the same four states (`waiting`/`active`/`delayed`/`failed`) and returns `{ job, state }`, so the behaviour @chriscrosstalk confirmed still holds: the only change is that an orphaned model entry is skipped instead of throwing.

One caveat on verification: I could not re-run `npm run typecheck` locally this time — `@openzim/libzim` fails its node-gyp build on this Windows host, unrelated to the change. The rebase itself I verified by diffing the file against the reviewed commit, which is identical apart from dev's `retryFailedJob` edit above. CI will cover the typecheck.

@chriscrosstalk your approval was a plain comment rather than a formal review, so it did not carry over — happy to re-request whenever suits.

jakeaturner (approved):

PR #1257, opened 2026-08-18
fix(maps): serve map assets with CORS

### LINKED ISSUE #1256: [Bug]: Map assets never send Access-Control-Allow-Origin, blocking cross-origin reads
**Issue Category:** Other

**Bug Description:**
NOMAD never sends an `Access-Control-Allow-Origin` header on any cross-origin request, on any route, which blocks other apps on the same NOMAD (Supply Depot apps, custom apps) or other tools on the LAN from reading NOMAD's own map data through the browser.

The cause is in `admin/config/cors.ts`. `origin` is set to `['*']`, an array containing the string `*`. `@adonisjs/cors` only treats `*` as a wildcard when it's passed as a plain string, not inside an array. Inside an array it gets compared for exact equality against the real incoming `Origin` header, which is always something like `http://192.168.1.x:8460`, so it never matches the literal text `*` and the header never gets sent.

**Steps to Reproduce:**
1. Pick any other host or port than NOMAD's own admin UI. From a browser on that origin, run: `fetch('http://<your-nomad-host>:8080/api/maps/styles')`
2. The request fails in the browser with a CORS error, even though the same URL works fine when opened directly or fetched from NOMAD's own origin.
3. Same result on any PMTiles file (`/pmtiles/<region>.pmtiles`) or basemap asset (`/basemaps-assets/...`).

**Expected Behavior:**
A cross-origin `fetch` against NOMAD's own map endpoints should succeed, the same way it would for any other static, unauthenticated file NOMAD serves to its own UI.

**Actual Behavior:**
The request is blocked by the browser before a response is ever read. Console output from Chrome:
```
Access to fetch at 'http://nomad:8080/api/maps/styles' from origin 'http://localhost:8199' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```
Server-side, confirmed directly against a live NOMAD instance:
```
$ curl -sSi -H "Origin: http://nomad:8460" http://nomad:8080/api/maps/styles | head -10
HTTP/1.1 200 OK
x-frame-options: DENY
x-content-type-options: nosniff
...
(no Access-Control-Allow-Origin anywhere in the response)

$ curl -sSi -X OPTIONS -H "Origin: http://nomad:8460" \
    -H "Access-Control-Request-Method: GET" http://nomad:8080/api/maps/styles
HTTP/1.1 204 No Content
(the preflight response has no CORS headers either)
```

**NOMAD Version:** 1.34.0
**Operating System:** Ubuntu 24.04
**Docker Version:** Docker version 29.6.1
**Do you have a dedicated GPU?:** Yes
**GPU Model:** AMD Radeon 780M (ROCm)

**Additional Context:**
I ran into this while building a separate map/navigation tool meant to read NOMAD's own downloaded map tiles. I have a fix ready: a small middleware that adds CORS headers only on the four map asset path prefixes, leaving every other route untouched, so it doesn't need to touch the broader global CORS config at all. I'll open the PR against this issue right after filing it.

cclavin:
## Summary

Adds a middleware that sends `Access-Control-Allow-Origin` on NOMAD's map asset routes, scoped to just those paths rather than touching the global CORS config.

Closes #1256

## Why

`origin: ['*']` in `admin/config/cors.ts` never matches a real `Origin` header. `@adonisjs/cors` only treats `*` as a wildcard when it's passed as the bare string `'*'`, not inside an array, so it falls into the exact-match branch and nothing ever matches. No `Access-Control-Allow-Origin` is sent on any route, which blocks any other app (Supply Depot apps, custom apps, other tools on the LAN) from reading NOMAD's own map tiles cross-origin.

I deliberately didn't fix `origin` globally. `config/cors.ts` already has `credentials: true` and allows `POST/PUT/DELETE`, so turning the array into a working wildcard would also expose the full write API to any website a LAN user's browser happens to have open, not just make map tiles readable. That's a materially worse change than what's actually needed here.

## Changes

- `admin/app/middleware/maps_cors_middleware.ts`: new middleware that grants anonymous, read-only CORS access to exactly `/api/maps/`, `/pmtiles/`, `/basemaps-assets/`, and `/storage/maps/`. Uses `response.safeHeader()` for `Access-Control-Allow-Origin` so it stays inert if the global config ever gets fixed, and answers `OPTIONS` preflights directly since `serve-static` only handles `GET`/`HEAD` and would otherwise 404 them.
- `admin/start/kernel.ts`: registers the middleware before `@adonisjs/cors`. This ordering matters: the built-in CORS middleware ends any preflight it can't match an origin for (`#endPreFlight`, no `next()`), which given the broken `origin: ['*']` is every preflight. Registered after it, this middleware would never see an `OPTIONS` request at all. Registered before it, it can answer map-path preflights itself; every other route still falls through to `@adonisjs/cors` unchanged.
- `admin/tests/unit/maps_cors.spec.ts`: unit tests on the pure path-matching predicate, 13 cases including near misses like `/api/mapsfoo`.

## Alternative considered

I also built a second version that changes `config/cors.ts`'s `origin` to a function scoped the same way, instead of adding a middleware (`@adonisjs/cors` supports `origin: (origin, ctx) => ...`). It's a smaller diff and fixes the actual config instead of leaving it in place, but only `origin` can be scoped per-request in that library, not `methods` or `credentials`, so a map-path preflight ends up sending a reflected origin, `Access-Control-Allow-Credentials: true`, and the full global method list, none of which the middleware approach sends. Went with the middleware since it's tighter on every axis even though it's a larger diff. Happy to switch approaches if you'd rather keep the diff smaller.

## Testing

Verified live against a real NOMAD instance (not just reasoned about): confirmed the bug pre-fix (`GET` came back 200 with no CORS header, `OPTIONS` preflight came back a bare 204), built this branch into a Docker image and ran it as an isolated container against a read-only mount of real map data, and confirmed post-fix that `GET` and `OPTIONS` both return the expected CORS headers on map paths while a non-map path (`/api/health`) is untouched. `node ace build` and the unit tests both pass.

### DIFF
--- a/admin/app/middleware/maps_cors_middleware.ts
+++ b/admin/app/middleware/maps_cors_middleware.ts
@@ -1,0 +1,34 @@
+import type { HttpContext } from '@adonisjs/core/http'
+import type { NextFn } from '@adonisjs/core/types/http'
+
+// origin: ['*'] in config/cors.ts never matches a real Origin header
+// (@adonisjs/cors only wildcards on the bare string origin: '*'), and a
+// global fix would also expose the write API since credentials: true is
+// set. Scope anonymous, read-only CORS to the map asset paths instead.
+
+const MAP_ASSET_PREFIXES = ['/api/maps/', '/pmtiles/', '/basemaps-assets/', '/storage/maps/']
+
+export function isMapAssetPath(pathname: string): boolean {
+  return MAP_ASSET_PREFIXES.some((prefix) => pathname.startsWith(prefix))
+}
+
+export default class MapsCorsMiddleware {
+  async handle({ request, response }: HttpContext, next: NextFn) {
+    if (!isMapAssetPath(request.url())) return next()
+
+    // safeHeader so this stays inert if the global CORS config is ever fixed.
+    response.safeHeader('Access-Control-Allow-Origin', '*')
+    response.header('Access-Control-Expose-Headers', 'ETag, Content-Range, Accept-Ranges, Content-Length')
+    response.header('Vary', 'Origin')
+
+    if (request.method() === 'OPTIONS') {
+      // serve-static only handles GET/HEAD, so OPTIONS would otherwise 404.
+      response.header('Access-Control-Allow-Methods', 'GET, HEAD, OPTIONS')
+      response.header('Access-Control-Allow-Headers', 'Range, If-None-Match, If-Match')
+      response.header('Access-Control-Max-Age', '600')
+      return response.status(204)
+    }
+
+    return next()
+  }
+}
--- a/admin/app/middleware/maps_cors_middleware.ts
+++ b/admin/app/middleware/maps_cors_middleware.ts
@@ -4,17 +4,24 @@ import type { NextFn } from '@adonisjs/core/types/http'
 // origin: ['*'] in config/cors.ts never matches a real Origin header
 // (@adonisjs/cors only wildcards on the bare string origin: '*'), and a
 // global fix would also expose the write API since credentials: true is
-// set. Scope anonymous, read-only CORS to the map asset paths instead.
+// set. Scope anonymous, read-only CORS to map tiles/assets and the one
+// JSON route another app actually reads, GET/HEAD/OPTIONS only.
 
-const MAP_ASSET_PREFIXES = ['/api/maps/', '/pmtiles/', '/basemaps-assets/', '/storage/maps/']
+const MAP_ASSET_EXACT_PATHS = ['/api/maps/styles']
+const MAP_ASSET_PREFIXES = ['/pmtiles/', '/basemaps-assets/', '/storage/maps/']
+const READABLE_METHODS = new Set(['GET', 'HEAD', 'OPTIONS'])
 
 export function isMapAssetPath(pathname: string): boolean {
-  return MAP_ASSET_PREFIXES.some((prefix) => pathname.startsWith(prefix))
+  return MAP_ASSET_EXACT_PATHS.includes(pathname) || MAP_ASSET_PREFIXES.some((prefix) => pathname.startsWith(prefix))
+}
+
+export function isReadableMethod(method: string): boolean {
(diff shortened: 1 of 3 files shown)


chriscrosstalk:
This is good work, and thanks for stopping short of the global fix. You read that correctly: `credentials: true` is set and the write methods are allowed, so turning the array into a working wildcard would be a much bigger change than the one you needed.

One thing to tighten before this goes in. The `/api/maps/` prefix isn't only static assets, it has eight POST routes under it (`download-remote`, `download-collection`, `download-global-map`, `setup-world-basemap`, and friends). The middleware sets `Access-Control-Allow-Origin` before it looks at the method, so those responses get the header too. The `Allow-Methods` list keeps preflighted requests out, but not everything preflights, and there's no reason for a write response to be cross-origin readable in the first place.

Can you gate the header emission on the method, so only GET, HEAD, and OPTIONS get it? Everything else falls through untouched.

Worth considering too whether `/api/maps/` needs to be in the prefix list at all. The cross-origin need you described is tiles and assets, which is `/pmtiles/`, `/basemaps-assets/`, and `/storage/maps/`. If `/api/maps/styles` is the only JSON route another app actually reads, matching it exactly would be tighter than the whole prefix.

The `safeHeader` choice and the middleware ordering note in `kernel.ts` are both right, and the near-miss cases in the test are the ones that matter.

cclavin:
Thanks for the review, appreciate you digging into it.

Solid catch on the method gate, that's a real gap. Fixed, the header's now only emitted for GET/HEAD/OPTIONS so the write routes fall through untouched either way.

`/api/maps/styles` is the only JSON route I actually need cross-origin, so I narrowed it to that exact path instead of the whole prefix. The other three (`/pmtiles/`, `/basemaps-assets/`, `/storage/maps/`) were never affected either way, they're served through the static provider, no write route could land under them.

Pushed as a new commit rather than rewriting the first one, happy to squash before merge if you'd rather have a clean single commit.

PR #1259, opened 2026-08-19
feat(AI): constrain the structured task calls instead of asking nicely

jakeaturner:
Ollama's format parameter turns a formatting request into a decoder constraint, and NOMAD's three ancillary calls — chat title, suggestion chips, RAG query rewrite — were the ones paying for not having it. The rewrite is particularly important: whatever it returned got embedded and sent to Qdrant, so a malformed response degraded retrieval for the rest of the conversation.

Three explicit design choices made:
- Native-only. The compat path sends nothing new and keeps its string parsing, so a non-Ollama backend that rejects response_format can't take out title generation entirely.
- parseStructured returning null is the control flow. It's how each caller falls back to the parser it already had, which is why it's written to be total and never throw.
- {queries: []} not {query: ""} for the rewrite (tee-ing up future improvements)

### DIFF
--- a/admin/app/services/chat_service.ts
+++ b/admin/app/services/chat_service.ts
@@ -8,6 +8,13 @@ import { OllamaService } from './ollama_service.js'
 import { SYSTEM_PROMPTS } from '../../constants/ollama.js'
 import { toTitleCase } from '../utils/misc.js'
 import { resolveTasksModel } from '../utils/tasks_model.js'
+import {
+  SUGGESTIONS_SCHEMA,
+  TITLE_SCHEMA,
+  parseStructured,
+  pickSuggestions,
+  pickTitle,
+} from '../utils/structured_output.js'
 
 @inject()
 export class ChatService {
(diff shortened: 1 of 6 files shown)


QexleOSS (commented):
Copilot's 3 findings are the same root bug: the ?? raw / ?? content fallback conflates two different failures. On a non-native backend the model returns prose and raw text is the right answer. On the native path with format set the usual failure is a truncated or malformed JSON object and dumping that into the title, the chips or the RAG query is the exact thing this PR is trying to stop.

The includes('{') && includes('suggestions') sniff Copilot suggests works but it's duplicated across all three call sites and it's a guess. The caller already knows whether it sent format so branch on that instead: if parseStructured is null and you requested the grammar, return the safe empty (no chips/no title/user message) otherwise run the legacy prose parser. That kills the heuristic and makes the fallback principled.

For the RAG path specifically make sure the empty result after a malformed structured response actually routes back to the user's message, not to an empty string embed. Your own comment calls out that an empty query quietly poisons retrieval so don't stop at ''.

(Minor) in chat_service.ts the suggestions warning still says "falling back to text parsing" but with the short-circuit it no longer does so the log is lying now.

But other than that the design is solid👌. Native-only, a total parseStructured that never throws, required + additionalProperties: false on the schemas and the unit tests actually exercise the fallback paths. Nice.

PR #1265, opened 2026-08-23
feat(AI): support API key for remote OpenAI-compatible endpoints

kuswardhanietidims-svg:
## What & why

Project NOMAD's AI Assistant already supports pointing at any OpenAI-compatible API server — the README mentions Ollama, LM Studio, and llama.cpp. All of those are *local* backends that ignore the API key, so the client currently sends a hardcoded `nomad` placeholder.

Cloud AI gateways are OpenAI-compatible too, but they **require a real API key**. This PR adds an optional API key to NOMAD's existing "Remote Connection" configuration so cloud gateways work out of the box, and calls out [OrcaRouter](https://www.orcarouter.ai) as the first named example.

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway built for both models and agents. Like OpenRouter, it exposes a provider/model namespace across many models — but it also combines adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance behind the same endpoint. Adding it as a first-class provider means this project's users can use that stack directly, without treating OrcaRouter as an anonymous custom base URL. It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

## Implementation

- **`admin/constants/kv_store.ts` / `admin/types/kv_store.ts`** — new `ai.remoteOllamaApiKey` setting.
- **`admin/app/services/ollama_service.ts`** — `_initialize()` now reads the stored key and sends it with the OpenAI client. Local backends still get the harmless `nomad` placeholder.
- **`admin/app/controllers/ollama_controller.ts`** — `configureRemote` accepts an optional `apiKey`; a separate `configure-remote-api-key` endpoint saves/clears the key without re-running the URL connectivity test; `remote-status` reports `hasApiKey` and probes with the key; a new `remote-status-api-key` endpoint tests the key against `/v1/chat/completions` (the path that actually requires it).
- **`admin/app/controllers/settings_controller.ts`** — only the *presence* of a key is sent to the UI, never the key itself.
- **`admin/inertia/lib/api.ts` + `admin/inertia/pages/settings/models.tsx`** — the Models → Remote Connection page gains a password-style API key field, with a "saved" indicator and save/clear controls.

## Verification

- `npm run typecheck` — passes
- `npm run build` (the CI gate) — passes
- Live test against `https://api.orcarouter.ai/v1` using a real key:
  - `GET /v1/models` with the stored key → `HTTP 200`
  - `POST /v1/chat/completions` (model `orcarouter/free`) with the stored key → `HTTP 200`

Note: the full `npm test` suite can't run in this sandbox because the native `@openzim/libzim` binding requires `make`/`g++`, which aren't available without root. The changed modules are covered by typecheck and the build.

## Notes

For a user, this looks like: enter `https://api.orcarouter.ai/v1` as the remote URL, paste an OrcaRouter API key, and the AI Assistant uses it for chat, streaming, and embeddings — no separate client needed.

I'm an engineer on the OrcaRouter team.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -225,13 +225,70 @@ export default class OllamaController {
 
   async remoteStatus() {
     const remoteUrl = await KVStore.getValue('ai.remoteOllamaUrl')
+    const apiKey = await KVStore.getValue('ai.remoteOllamaApiKey')
     if (!remoteUrl) {
       return { configured: false, connected: false }
     }
     try {
       const testResponse = await fetch(`${remoteUrl.replace(/\/$/, '')}/v1/models`, {
+        headers: apiKey ? { Authorization: `Bearer ${apiKey}` } : undefined,
         signal: AbortSignal.timeout(3000),
       })
+      return { configured: true, connected: testResponse.ok, hasApiKey: !!apiKey }
+    } catch {
+      return { configured: true, connected: false, hasApiKey: !!apiKey }
+    }
+  }
+
+  /**
+   * Saves the API key for the remote OpenAI-compatible endpoint. Kept separate
+   * from configureRemote so an already-configured URL can stay in place while
+   * the key is changed (or cleared) without re-running the URL connectivity
+   * test. Null/empty clears the key.
+   */
+  async configureRemoteApiKey({ request }: HttpContext) {
+    const apiKey: string | null = request.input('apiKey', null)
+
+    if (!apiKey || apiKey.trim() === '') {
+      await KVStore.clearValue('ai.remoteOllamaApiKey')
+      return { success: true, message: 'Remote API key cleared.' }
+    }
+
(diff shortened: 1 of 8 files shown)


chriscrosstalk:
Thanks for taking the time on this.

We're not going to add cloud gateway support to NOMAD. Every remote backend the Remote Connection setting is built for runs on your own network: LM Studio, llama.cpp, or another Ollama instance. None of those authenticate, which is why there's no key field today. Adding one exists only to reach a paid API over the internet, and a box whose entire purpose is working when the internet doesn't is the wrong place for that dependency.

The named provider in the UI and the branded placeholder in the input field are also not something we'd ship regardless of the underlying feature.

Closing this one. If you want to make the case for cloud inference as a NOMAD feature, roadmap.projectnomad.us is the right place and we do read it.

PR #648, opened 2026-04-04
feat(i18n): Add internationalization support with English and Chinese

Yeluo10123:
## Summary
- Add i18next internationalization framework
- Add English (en) and Chinese (zh) translation files
- Create LanguageSwitcher component for runtime language switching
- Integrate i18n initialization in app.tsx

## Translation Structure
Translations are organized by feature:
- `common`: Common UI elements (buttons, labels)
- `home`: Dashboard/home page
- `menu`: Navigation menu items
- `maps`: Maps feature
- `chat`: AI chat feature
- `settings`: Settings pages
- `system`: System settings
- `apps`: App management
- `models`: AI models
- `easySetup`: Setup wizard
- `docs`: Documentation
- `about`: About page
- `errors`: Error pages

## Usage
Components can use translations like:
```tsx
import { useTranslation } from 'react-i18next'

function MyComponent() {
  const { t } = useTranslation()
  return <button>{t('common.save')}</button>
}
```

## Language Switcher
A `LanguageSwitcher` component is provided that can be added to the settings UI for users to switch between English and Chinese.

## Next Steps
- [ ] Add LanguageSwitcher to settings UI
- [ ] Translate remaining hardcoded strings in page components
- [ ] Add more languages (e.g., Spanish, French)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1267, opened 2026-08-25
Feature/i18n fr en

bocal-pol:


bocal-pol:
Hello, I'm a French speaker and I just translated what I could. If my family wants to use the nomad system, they'll be able to understand everything. And thank you for your work!

chriscrosstalk:
Thanks for the work here, and for translating the docs as well as the UI.

Two separate reasons this can't merge, and the first is specific to this PR rather than to i18n.

It changes `collections/wikipedia.json` and adds two new collection files, on a branch based on `main`. Curated manifests are fetched live from `main`, so those reach every existing install the moment they merge, while the UI changes only reach people who update the app. Since the French entries are only hidden by the `lang` filter added in this same PR, merging it would put French Wikipedia options in the picker of every English user still on the current build. Content manifest changes and app changes land on very different schedules, and they can't be mixed in one branch.

The second is the i18n framework itself. There are now four community i18n PRs using two different frameworks, each covering a different language, and #490, #518 and #648 were all closed for the same reason: the framework choice, key naming, locale loading, content-language wiring and ongoing translation maintenance need to be one coordinated set of decisions made in-house. Taking any one of the four strands the other three.

Closing this. i18n is on the roadmap at roadmap.projectnomad.us, and that's the right place to add weight to it. When the in-house framework lands, the prior community PRs will be credited, and adding an `fr.json` becomes a straightforward contribution.

PR #1278, opened 2026-08-26
docs: state how much disk NOMAD needs before any content

### LINKED ISSUE #1271: [Feature]: Readme: Clarify required disk space.
### Feature Category

Documentation

### Problem Statement

When I read the quickstart in the readme, I wondered how much free diskspace I'll need to install nomad.

### Proposed Solution

```diff
--- README.md@c3a413a0
+++ README.md
@@ -17,3 +17,8 @@
 ## Installation & Quickstart
-Project NOMAD can be installed on any Debian-based operating system (we recommend Ubuntu 26.04 LTS; 24.04 LTS and Debian 12 are also supported). Installation is completely terminal-based, and all tools and resources are designed to be accessed through the browser, so there's no need for a desktop environment if you'd rather setup NOMAD as a "server" and access it through other clients.
+Project NOMAD can be installed on any Debian-based operating system (we recommend
+Ubuntu 26.04 LTS; 24.04 LTS and Debian 12
+are also supported) with
+≈10 (5–25) GB
+available disk space (see "disk space" below).
+Installation is completely terminal-based, and all tools and resources are designed to be accessed through the browser, so there's no need for a desktop environment if you'd rather setup NOMAD as a "server" and access it through other clients.

```

<del>… with the real value instead of "0 GB".</del>
The extra line breaks are meant to reduce future diff noise when the data lines will be updated. (The update in commit c3a413a0 could have been much simpler to verify if the relevant data had its own line.)

### Alternative Solutions

_No response_

### Use Case

As a GitHub wanderer, when I discover an interesting project, I want to quickly categorize whether I can play with it right now of if I have to file it away for later.

All Users

_No response_

### Examples or References

_No response_

No - I don't have the skills/time

### Additional Context

_No response_

### Pre-submission Checklist

- [x] This feature aligns with NOMAD's mission of offline-first knowledge and education

chriscrosstalk:
Closes #1271.

The quickstart never said how much free space is needed to install. That is the first thing anyone plans around on a project whose whole purpose is downloading large offline libraries, and the reporter proposed the exact sentence with the number left blank for us to fill in.

The FAQ's `How much storage do I need?` question only covered content sizes, so it answered "how big is Wikipedia" but never "can I install this at all". Both are updated.

## Measured, not estimated

On a v1.34.0 install:

| Image | Size |
|---|---|
| admin (`project-nomad:v1.34.0`) | 2.81 GB |
| MySQL 8.0 | 1.1 GB |
| sidecar-updater | 137 MB |
| dozzle | 76.5 MB |
| redis 7-alpine | 57.8 MB |
| disk-collector | 25.1 MB |
| **core total** | **~4.2 GB** |

So ~5 GB installed, and 10 GB free to install comfortably.

The AI Assistant is the big optional add: `ollama/ollama` is **10.6 GB** on its own, and a general-purpose model is several more (llama3.1:8b is ~4.9 GB). That is where the ~25 GB figure comes from.

## Noted, not changed

The FAQ lists "Full Wikipedia: ~95GB". The `wikipedia_en_all_maxi_2026-02` archive is actually **124 GB** on disk. I left it alone because per-build ZIM sizes move and PR #1189 already re-measured the catalog, so that number should be corrected from the catalog rather than from my one sample.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Project NOMAD is a self-contained, offline-first knowledge and education server 
 ## Installation & Quickstart
 Project NOMAD can be installed on any Debian-based operating system (we recommend Ubuntu 26.04 LTS; 24.04 LTS and Debian 12 are also supported). Installation is completely terminal-based, and all tools and resources are designed to be accessed through the browser, so there's no need for a desktop environment if you'd rather setup NOMAD as a "server" and access it through other clients.
 
-**Disk space:** NOMAD itself needs about **5 GB**, so plan for **10 GB free** to install comfortably. Adding the AI Assistant needs roughly **25 GB** in total, since Ollama and a general-purpose model are large. Offline content is downloaded separately and is what actually fills a drive: a full Wikipedia or a regional map set runs to tens or hundreds of gigabytes, so an SSD of 500 GB or more is the practical recommendation once you start adding libraries. See the [FAQ](https://www.projectnomad.us/docs/faq) for content sizes.
+**Before you start:** NOMAD itself needs about **5 GB of disk and under 1 GB of RAM**. The offline content you add afterwards is what actually fills a drive, and the AI Assistant is what drives the memory requirement up. See [Device Requirements](#device-requirements) for both, and the [FAQ](https://www.projectnomad.us/docs/faq) for individual content sizes.
 
 *Note: sudo/root privileges are required to run the install script*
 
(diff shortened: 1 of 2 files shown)


chriscrosstalk:
Extended this to cover RAM as well, per @mk-pmb's follow-up on #1271. Same approach as the disk numbers: measured on a running install rather than estimated.

The README already stated 4 GB and 32 GB. What was missing is why the gap is that wide. The whole stack without AI runs under 1 GB, so the jump is entirely the model, and whether it lands in VRAM or system RAM depends on whether there is a discrete GPU.

jakeaturner (approved):

PR #1286, opened 2026-08-30
feat(kb): per-file active toggle + collection grouping/bulk toggle

chriscrosstalk:
Salvages @just-jbc's work from #1192, which had drifted onto a stale base (216 files in GitHub's diff, almost all unsquashed duplicates of work `dev` already has). Both commits are cherry-picked with authorship preserved.

**What it adds**
- A per-file active/inactive toggle controlling whether a KB file is included in search.
- Grouping of personal KB uploads by collection, with a bulk toggle per group and a mixed/indeterminate state when a group's members disagree.
- `POST /rag/files/active` and `POST /rag/collection-active`.
- Migration adding `active` to `kb_ingest_state`, non-null, defaulting to true.

**Retrieval correctness.** There is exactly one Qdrant read path in the admin app (`RagService.searchDocuments`), and the filter sits on it, so a file toggled off cannot surface through another route. It's built as a **denylist** (`must_not: active == false`) rather than an allowlist, so points predating the field are still found and search correctness never depends on the backfill's timing. The backfill itself filters on `is_empty`, so re-running it can never clobber an explicit toggle-off.

**Verification.** Typecheck clean; `vite build` clean; 22/22 in `kb_file_grouping.spec.ts`. The unit-test failure set is identical to `origin/dev` (all pre-existing Windows-environment issues — `@openzim/libzim` has no Windows binding). Not yet browser-verified on hardware.

Reduced from his originals: `routes.ts` 131 lines of churn to 8, `api.ts` 242 to 10, and `CollectionCombobox.tsx` kept on dev's post-#1198 version since he'd independently re-solved the same portaling problem — his `userTyped` fix is retained.


chriscrosstalk:
Browser-verified on NOMAD3 against a full build of `dev` + this branch. It found two defects, both now fixed in `4708f84`.

**Verified working**
- Toggling a file off writes through end to end: the Qdrant points for that source flip to `active: false` and the `kb_ingest_state` row goes to 0. Toggling back on restores both.
- The `is_empty` backfill stamped all 117,535 existing points on first touch of the collection, and never clobbers an explicit toggle-off.
- Collection grouping, the per-row combobox, and the read-only "Managed by NOMAD" documentation group all render correctly.

**Fixed: the Active column reintroduced the defect #1198 fixed.** The new column adds ~57px, which took the table to 1018px against 961px of usable width at `max-w-5xl` — putting the Delete button back off the right edge at every viewport, behind a horizontal scrollbar. Widened to `max-w-6xl`: 1081px in 1089px, zero overflow.

**Fixed: `aria-checked="1"`.** MySQL returns `tinyint(1)` as 0/1 and the raw value reached the DOM. That is not a valid ARIA value, so a screen reader is told nothing useful about the toggle's state. Coerced in `rag_service` so every consumer sees a real boolean, with a `!!` guard in `Switch` as well. Now renders `aria-checked="true"`.

Neither was reachable from typecheck, unit tests, or reading the diff.

Unit suite after the fixes: 438 passing, failure set unchanged from `origin/dev`.

jakeaturner (approved):

PR #1287, opened 2026-08-30
feat: show source document and date as citations under RAG chat answers

### LINKED ISSUE #1179: [Feature]: Surface source document + date as citations under AI chat answers

chriscrosstalk:
Closes #1179. Salvages @just-jbc's work from #1180, which had drifted onto a stale base. His commit is preserved with authorship intact and rebased onto the `RagPipelineService` refactor that landed after he wrote it.

**What it does.** Shows the documents an answer was drawn from as a "Sources" list under the assistant message, with the archive date where one is known, and persists it so reopening a past conversation still shows what backed each answer. User uploads with no embedded metadata fall back to their filename.

**Why.** Offline there is no second opinion to check an answer against, so provenance is the only signal a user has that a claim came from a real document rather than from the model. This is the other half of letting retrieval decline: one says when nothing relevant was found, this says what was found.

**Citations come off `trace.injected`, not `trace.retrieved`.** A chunk dropped by the budget planner never reached the model, so citing it would credit the answer to a document it was not based on — the exact failure the list exists to prevent.

**No re-embed needed.** `archive_title`/`archive_date` have been written into the vector payload by the ZIM ingestion path all along and were simply never read back, so this lights up on already-embedded content.

**Changes from the original**
- The controller half is rebuilt against the pipeline trace; the inline retrieval block it hooked into no longer exists.
- The dedupe/fallback logic moved into `buildCitations` in `rag_prompt.ts`, next to `buildContextBlock`, so it is unit-testable away from the controller. 9 tests added.
- Entries with neither a path nor a title are skipped rather than rendered as "Unknown source" — a row the user cannot act on makes an answer look sourced when it is not.
- Dedupe keys on the originating path rather than the title, so two editions of the same archive stay distinct.
- The repeated inline source shape is now a named `ChatSource` type.
- Dropped a duplicate `source` key his patch would have added to the chunk metadata; `dev` already added it.

**Verification.** Typecheck clean; `vite build` clean; 9/9 new tests; full unit suite failure set identical to `origin/dev` (441 passing, the 12 failures all pre-existing Windows-environment issues). Not browser-verified yet: NOMAD3's running image predates the `RagPipelineService` refactor entirely, so a single-file hot patch would be inconsistent with the rest of the container. Verifying this needs a full build of current `dev` on the box.

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -14,6 +14,7 @@ import type { HttpContext } from '@adonisjs/core/http'
 import { SERVICE_NAMES } from '../../constants/service_names.js'
 import { DEFAULT_KEEP_ALIVE } from '../../constants/ollama.js'
 import type { PipelineTrace } from '../../types/rag.js'
+import { buildCitations } from '../utils/rag_prompt.js'
 import logger from '@adonisjs/core/services/logger'
 
 @inject()
@@ -79,6 +80,11 @@ export default class OllamaController {
       reqData.messages = trace.messages
       const numCtx = trace.numCtx
       const numPredict = trace.numPredict
+      // Provenance for the answer about to be generated (#1179). Built from
+      // trace.injected -- what the model actually read -- and surfaced under the
+      // answer as "Sources". Empty whenever retrieval was skipped or declined,
+      // which is the honest result: no context, no citations.
+      const sources = buildCitations(trace.injected)
       // Keeping the model resident is what makes the KV cache worth building:
       // Ollama's default evicts after 5 minutes, which is well inside the time a
       // user spends reading an answer and typing the next question.
@@ -145,11 +151,18 @@ export default class OllamaController {
           }
           throw err
         }
+        // Trailing citation event, written before end(). It carries no `message`
+        // key, which is how the client tells it apart from Ollama's own chunks.
+        if (sources.length > 0) {
+          response.response.write(`data: ${JSON.stringify({ sources })}
+
+`)
+        }
         response.response.end()
 
         // Save assistant message and optionally generate title
         if (sessionId && fullContent) {
-          await this.chatService.addMessage(sessionId, 'assistant', fullContent)
+          await this.chatService.addMessage(sessionId, 'assistant', fullContent, sources)
           const messageCount = await this.chatService.getMessageCount(sessionId)
           if (messageCount <= 2 && userContent) {
             this.chatService.generateTitle(sessionId, userContent, fullContent, reqData.model).catch((err) => {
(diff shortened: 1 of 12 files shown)


chriscrosstalk:
Browser-verified on NOMAD3 against a full build of `dev` + this branch. No changes needed.

Asked a gardening question with the knowledge base on. The answer rendered:

> **Sources**
> Gardening & Landscaping Q&A (2026-02-02)

The archive title and date resolved from the ZIM's own embedded metadata, and multiple retrieved chunks from that archive collapsed to a single entry. Reloading the page and reopening the conversation still showed the list, so the persisted path works too.

Confirmed on the box that this needed no re-embed: `archive_title`/`archive_date` were already present on the existing vectors, having been written at ingest and never read back.

Unit suite unchanged from `origin/dev`.

jakeaturner (approved):

PR #1288, opened 2026-08-30
feat(ai): support local vision attachments

### LINKED ISSUE #1176: [Feature]: Attach images to chats with local multimodal models

chriscrosstalk:
Closes #1176. Salvages @gabegraves's work from #1178, which had drifted onto a stale base. His eight commits are preserved as one, with authorship intact, rebased onto the `RagPipelineService` refactor that landed after he wrote them.

**What it does.** Attach a JPEG, PNG or WebP to a chat message and ask about it. Images are normalized server-side to bounded JPEG with metadata stripped, capped on count/size/pixels/output, and the temporary uploads are cleaned up on every validation path. They go only with that one request: excluded from RAG, titles, logs and persisted history, and never written to storage.

Vision support is resolved from what the backend actually reports — OpenAI-compatible `/v1/models` metadata, then Ollama `/api/show`, then llama.cpp `/props` — never inferred from a model name. A model reporting no vision support has attachments disabled with a plain explanation; a backend reporting nothing is allowed with a warning, and its rejection is translated into actionable guidance rather than a raw error.

**Changes from the original**

- The controller half is rebuilt against the pipeline trace; the inline retrieval block it hooked into no longer exists.
- **Added a native-path converter.** `dev` split chat into native and OpenAI-compat paths after this was written. Ollama's native `/api/chat` does not accept OpenAI content parts — it takes a sibling `images` array of bare base64. Without this, every image request on a native Ollama backend would have failed.
- **The multipart-aware collection filter is now the only one.** `dev`'s `request.input('collection')` would have silently dropped the collection on every image request, since a multipart chat carries it inside the JSON payload.
- **Dropped the model-recommendation change** (`selectRecommendedModels` and the Qwen3-VL fallback entry). Displacing a recommended model during first-time setup is a product decision being taken separately. The "Supports images" label in Easy Setup is kept — it describes what is already on offer rather than changing it.
- **Dropped the `DEFAULT_QUERY_REWRITE_MODEL` download prompt** his branch still carried; `dev` deleted it in `d022c2d`.
- Removed a duplicate `checkModelHasThinking`, keeping dev's memoized `getModelInfo` path.
- **Moved the controller spec to the Japa runner.** It constructs `OllamaController`, which imports the Adonis logger service at module scope, so it cannot run under the plain `node:test` runner — it failed identically on the original branch, meaning that coverage had never actually executed.

**Verification.** Typecheck clean; `vite build` clean. `chat_images`, `chat_request_compat`, `chat_stream` and `vision_guidance` all pass. The unit-suite failure set is `origin/dev`'s seven Japa specs plus this one, all of which need a booted app under the plain runner.

Browser-verified on NOMAD3 against a build of `dev` + this branch, with `qwen2.5vl:3b` on an RTX 5060:

- Attaching an image shows a preview with a remove control and the ephemerality notice.
- Asked what was in a generated test image; the model answered "a blue circle, a red square, and a yellow triangle" — correct, and it exercises the new native-path conversion.
- Switching to a model that reports no vision support (`llama3.1:8b`, capabilities `['completion','tools']`) disables the attach button and shows "This model cannot use images."
- Capability resolution checked against `/api/show` for every installed model.

### DIFF
--- a/admin/app/controllers/ollama_controller.ts
+++ b/admin/app/controllers/ollama_controller.ts
@@ -16,6 +16,16 @@ import { DEFAULT_KEEP_ALIVE } from '../../constants/ollama.js'
 import type { PipelineTrace } from '../../types/rag.js'
 import { buildCitations } from '../utils/rag_prompt.js'
 import logger from '@adonisjs/core/services/logger'
+import { rm } from 'node:fs/promises'
+import {
+  attachImagesToLatestUserMessage,
+  ChatImageError,
+  normalizeChatImages,
+  type NormalizedChatImage,
+} from '../utils/chat_images.js'
+
+const unknownVisionCompatibilityMessage = (model: string) =>
+  `NOMAD cannot confirm that "${model}" accepts images, and this image request failed. Choose a model marked "Supports images", or remove the image and try again.`
 
 @inject()
 export default class OllamaController {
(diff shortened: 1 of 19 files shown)


gabegraves:
Hey @chriscrosstalk, thanks for doing the work to get the vision attachment feature updated and merged into dev. I noticed the squash merge on #1288 changed the landed commit author to you and only kept Claude as a co-author, so GitHub no longer gives me contributor credit even though the PR says my authorship from #1178 was preserved.
Would you mind making sure a non-merge commit that eventually reaches main includes this?
Co-authored-by: Gabe Graves <37971265+gabegraves@users.noreply.github.com>
I know you had to do real integration work to get it onto the current codebase. I am just hoping to keep attribution and visibility for the original implementation too. Thanks!
cc @jakeaturner

chriscrosstalk:
You're right, and thanks for flagging it. The squash took its message from the last commit on the branch (the test-runner move) rather than the PR description, and your Co-authored-by trailer went with it. That's on us — the PR said your authorship was preserved and the merge undid that.

We'll get an attribution commit with
`Co-authored-by: Gabe Graves <37971265+gabegraves@users.noreply.github.com>`
onto a non-merge commit that reaches main. It's on the release checklist so it doesn't get lost.

Issue #1272, opened 2026-08-25
Offline translation for ZIM content (Supply Depot app + path-preserving Kiwix proxy)


chriscrosstalk:
## Problem

Most content on a NOMAD box is English. A user whose first language isn't English has a 200 GB library they can't read, and no offline way to fix that. This is the most-requested language capability we have: **"Offline translation engine" is 14 votes and 4th of 30 posts on the roadmap**, and it's the only language-related item anyone has voted on.

This is **not** UI internationalization. That's a separate piece of work and it can't translate a Wikipedia article.

## Why not the AI Assistant

Ollama can translate, so this was previously deferred. Measured on a test box, same passage, same machine, with an RTX 5060 doing the LLM work:

| Engine | Hardware | Throughput |
|---|---|---|
| Bergamot, 8 workers | CPU only | **12,818 words/s** |
| Bergamot, 1 worker | CPU only | 2,612 words/s |
| `llama3.1:8b` | RTX 5060 | **8 words/s** |
| `qwen2.5:3b` | RTX 5060 | 11 words/s |

Roughly 1,600x on batch. It's also more accurate: `llama3.1:8b` translated the product name ("Projet NOMAD"), and `qwen2.5:3b` hallucinated a clause that wasn't in the source. And it needs no GPU, so it works on CPU-only boxes where the LLM path is unusable.

## Why this isn't the LibreTranslate request

#707 and #723 were closed on resource footprint, correctly, for LibreTranslate and Argos. Bergamot (the models Firefox ships) is a different class:

- **37 MB per direction**, 74 MB for a bidirectional pair
- **CPU only**, no GPU contention with Ollama
- **0.04 s** model load, so pairs load on demand rather than sitting resident
- MPL-2.0, no licence contamination

Models come from Firefox's Remote Settings CDN, the same endpoint Firefox uses. 84 pairs, about 44 languages. **No Chinese, Japanese, Korean, Arabic or Thai**, so this does not close #647/#648, but it does address #1221.

## Proposed architecture: path-preserving proxy

A reverse proxy in front of `kiwix-serve` that **preserves every path**:

```
proxy
├── /viewer, /skin/*, /search, /random, /catalog/*  -> forwarded byte-for-byte
└── /content/<book>/<article>  (text/html, 200)     -> translated in place
```

[...]

PR #1292, opened 2026-08-31
feat(translate): offline translation for the Information Library

chriscrosstalk:
Implements #1272. Offline machine translation for the Information Library, shipped as a Supply Depot app so users who do not need it never see it.

## What it is

A **path-preserving reverse proxy** in front of `kiwix-serve`. `/viewer`, `/skin/*`, `/search`, `/random` and `/catalog/*` are forwarded byte for byte; only `text/html` under `/content/` is rewritten.

Path preservation is the whole design. `viewer.js` keeps the content iframe in sync with `location.hash` by comparing `contentWindow.location.pathname`, so moving the path starts the ping-pong its own comment warns about and forces us to patch Kiwix's JavaScript and re-verify it on every Kiwix bump. Because the path never moves, **no Kiwix JS is modified**, and Kiwix's header, search, library and random keep working because they are Kiwix. Language is a cookie, so links inside the ZIM need no rewriting either.

Reimplementing or wrapping the Kiwix reading view stays explicitly out of scope.

## Why Bergamot rather than the AI Assistant

Measured on the same box, same passage: **12,818 words/s on CPU** against `llama3.1:8b` at **8 words/s on an RTX 5060**. It is also more careful with names, where the LLM translated the product name and a smaller model hallucinated a clause. No GPU, so it works on CPU-only boxes where the LLM path is unusable.

Models come from **Firefox Remote Settings**, the endpoint Firefox itself uses, not the `mozilla/firefox-translations-models` repo, which was archived on 2026-08-21. MPL-2.0, about 37 MB per direction.

## Failure handling

Deliberate throughout, because translation being unavailable must never make the Information Library unreachable:

- A failed model fetch is non-fatal. The proxy still starts and still forwards the library.
- A box with models already on disk skips the network entirely, so restarting while genuinely offline works.
- A translation error returns the original article with an HTML comment, not an error page.
- Bergamot is imported lazily, so a box where models never downloaded still serves Kiwix untouched.
- Pages over 4 MB are forwarded untranslated rather than tying up every worker.

## Verified

**Transparency**, proxy against Kiwix direct:

| Path | Direct | Proxy |
|---|---|---|
| `/viewer` | 200, 4236 b | 200, **4236 b** |
| `/skin/viewer.js` | 200, 20736 b | 200, **20736 b** |
| `/catalog/v2/entries` | 200, 10260 b | 200, **10260 b** |
| `/search` malformed | **400, 1630 b** | **400, 1630 b** |

Reproducing Kiwix's own `400` byte for byte is the meaningful one.

**In a browser on real hardware**, against a live Wikipedia ZIM: the control renders inside the content frame, French and Spanish both translate correctly (332 words in 87 ms warm, 440 words in 268 ms), inline links land on the right translated words (`les filtres à sable lent`, `requin pèlerin`), switching back to Original works, and the language persists when moving to another article. The Supply Depot card renders with the right name, icon, category and port.

**Tests:** 24 checks for the block planner, no Bergamot and no network, run with `python3 install/nomad-translate/test_blocks.py`. Typecheck and `vite build` clean; unit-suite failure set identical to `origin/dev`.

## Three defects found by running it rather than reading it

- **The cache did not vary on the language cookie.** Kiwix sends `ETag` and `Cache-Control: max-age=3600`, which the proxy forwarded unchanged on a response that now varies by cookie. The browser served the English article back for an hour when the reader picked a language, so the feature looked broken; a shared cache would have served one reader's language to another. Translatable responses now send `Vary: Cookie` with a language-namespaced ETag.
- **Port 8480 is already Vaultwarden's.** My survey grepped `ui_location` for a bare number and Vaultwarden's is `'https:8480'`, so it looked free. Moved to 8460.
- **The image did not build on Python 3.12**, which is what I had set it to. Bergamot publishes no wheel for it. Pinned back to 3.10, which is what the spike used.

## Known gaps, carried from the spike

Tables and infoboxes are not translated, `<title>` is not translated, Kiwix's search results page is not translated and search still matches original-language terms, and there is no caching so every view re-translates. A block whose inner tag is never closed is dropped rather than translated: safe, but silent. All are documented in `install/nomad-translate/README.md`.

**Two language controls.** Kiwix's toolbar has a globe that changes the interface language; ours changes the article. Ours is labelled "Translate this page" rather than with a globe, which is the cheapest mitigation available, but we do not own the other button.

## Deliberately not in this PR

- **The Information Library link is untouched.** The spike notes call repointing it the riskiest part of the feature, and #1005 is an open bug where the seeder reverts fields on non-user-modified services, which is exactly the failure mode a programmatic `custom_url` would hit. The app is its own tile instead, which is how every other Supply Depot app behaves.
- **Model management through the download system** (resume, storage projection, Settings UI). Models are fetched on first start into `storage/translate/models`. That needs internet the first time, the same as installing any other app.

## Before merge

The image must be built and pushed first: run the new **Build Translate Proxy Image** workflow at version `0.1.0`, since the catalog entry points at `ghcr.io/crosstalk-solutions/project-nomad-translate:0.1.0`. The workflow pins `linux/amd64` deliberately: Bergamot's intgemm backend is x86 specific and there is no aarch64 wheel, so a multi-arch build would produce an ARM image that pulls cleanly and dies at start.

Also adds `.gitattributes` pinning `*.sh` to LF. The repo stores LF but Windows checks out CRLF, and a CRLF after a shebang makes the interpreter unresolvable inside an image. CI builds on Linux so it never bit, but it breaks a local image build and would have bitten the existing sidecar script too.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Shell scripts must keep LF endings. These are copied into Linux container
+# images, where a CRLF after the shebang makes the interpreter unresolvable and
+# the container exits immediately. Checkouts on Windows would otherwise convert
+# them, which only shows up when building an image locally rather than in CI.
+*.sh text eol=lf
--- a/.github/workflows/build-translate-proxy.yml
+++ b/.github/workflows/build-translate-proxy.yml
@@ -1,0 +1,52 @@
+name: Build Translate Proxy Image
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Semantic version to label the Docker image under (no "v" prefix, e.g. "1.2.3")'
+        required: true
+        type: string
+      tag_latest:
+        description: 'Also tag this image as :latest?'
+        required: false
+        type: boolean
+        default: false
+
+jobs:
+  check_authorization:
+    name: Check authorization to publish new Docker image
+    runs-on: ubuntu-latest
+    outputs:
+      isAuthorized: ${{ steps.check-auth.outputs.is_authorized }}
+    steps:
+      - name: check-auth
+        id: check-auth
+        run: echo "is_authorized=${{ contains(secrets.DEPLOYMENT_AUTHORIZED_USERS, github.triggering_actor) }}" >> $GITHUB_OUTPUT
+  build:
+    name: Build translate-proxy image
+    needs: check_authorization
+    if: needs.check_authorization.outputs.isAuthorized == 'true'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+      - name: Log in to GitHub Container Registry
+        uses: docker/login-action@v4
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push
+        uses: docker/build-push-action@v7
+        with:
+          context: install/nomad-translate
+          platforms: linux/amd64
+          push: true
+          tags: |
+            ghcr.io/crosstalk-solutions/project-nomad-translate:${{ inputs.version }}
+            ghcr.io/crosstalk-solutions/project-nomad-translate:v${{ inputs.version }}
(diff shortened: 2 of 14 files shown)


jakeaturner (approved):
Awesome

PR #1296, opened 2026-08-31
fix(zim): use the OPDS catalog host, not Kiwix's gated browse host

chriscrosstalk:
**Content Explorer's remote library is broken on current builds, and ZIM auto-update has been silently stalled since 2026-08-29.**

`browse.library.kiwix.org` is Kiwix's *human* browsing host. On 2026-08-29 (`Last-Modified` on the response) it grew an anti-crawler interstitial: "Due to unreasonnable traffic from (AI) crawlers, we're requiring an initial confirmation step to continue." It answers **HTTP 200** with an HTML page requiring JS and a cookie. Both of our catalog callers hardcode that host, and they fail differently:

- `zim_service.listRemote()` — the HTML fails `isRawListRemoteZimFilesResponse`, throws `Invalid response format from remote library`, and `GET /api/zim/list-remote` returns **500**. Reproduced in the browser on a test box: four consecutive 500s, page shows "An internal error occurred".
- `kiwix_catalog_service.fetchZimEntriesPage()` — `parseZimEntries` catches the parse failure and returns an empty feed, so every freshness check reads as **"no updates available"**. No error, no log, indefinitely.

Kiwix split the hosts: `library.kiwix.org` now 301s to **`opds.library.kiwix.org`**, which serves the same OPDS API ungated. This repoints both call sites there. No cookie handling, no user-agent games — just the host Kiwix now points machine clients at.

**Verified** on a test box by patching the running container and restarting: `list-remote` 500 → 200 with real entries, and the Content Explorer table renders.

Two lines of behaviour change; the added comments are there so nobody repoints this at the browse host again.

https://claude.ai/code/session_01F8Mgg1vd2eKSs8StZiuyMV

### DIFF
--- a/admin/app/services/kiwix_catalog_service.ts
+++ b/admin/app/services/kiwix_catalog_service.ts
@@ -4,6 +4,7 @@ import { DateTime } from 'luxon'
 import logger from '@adonisjs/core/services/logger'
 import InstalledResource from '#models/installed_resource'
 import { isRawListRemoteZimFilesResponse } from '../../util/zim.js'
+import { KIWIX_CATALOG_BASE_URL } from '../../constants/kiwix.js'
 
 /**
  * Local, in-process freshness check for installed content (Kiwix ZIM files +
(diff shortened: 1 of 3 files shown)


jakeaturner (approved):
@chriscrosstalk , I took this one step further and extracted the URL into a system-wide constant in `admin/app/constants/kiwix.ts` just to further de-dupe

PR #1298, opened 2026-08-31
fix(zim): resolve flavoured and non-English ZIMs in the update check

chriscrosstalk:
Stacked on #1296 so the diff is just this change; GitHub will retarget it to `dev` when that merges.

Two independent reasons `getLatestZim()` returns null for books that do have newer versions. Both predate the catalog-host problem in #1296 — that outage was masking them.

### 1. Flavoured books are indexed under their base name

OPDS carries the flavour in a separate `<flavour>` element:

```
name=wikipedia_en_all_maxi   -> 0 entries
name=wikipedia_en_all        -> 3 entries (maxi_2026-02, mini_2026-06, nopic_2026-06)
q=wikipedia_en_all_maxi      -> 0 entries
```

Our resource ids are filename bases, flavour included, so the exact-name path missed **every maxi/mini/nopic book — including every Wikipedia**. The `q=` fallback doesn't cover it either: `q=` searches title and summary, not the filename, and was not observed to resolve any resource id on its own.

This adds a retry against the flavour-stripped base. No flavour allowlist: openZIM adds flavours over time, and the authoritative `^<id>_YYYY-MM\.zim$` check still gates every result, so a wrong guess costs one request and nothing else.

### 2. `lang: 'eng'` on a freshness check

This is a lookup for an already-installed book, so its language is whatever the user installed. A French or Spanish ZIM was permanently unresolvable. The filename regex is what identifies the book; the language filter only removed correct answers.

Note this differs from the Content Explorer path, where a language filter is user-facing and should stay.

### 3. Drop the dead keyword-scan fallback

`scanZimByQuery` cost up to 5 catalog requests per unresolved book on every check and never resolved one, for the reason above. Removed rather than left as traffic aimed at a mirror that has just deployed anti-crawler measures.

### Verification

On a test box against the live catalog, via `POST /api/content-updates/check` with 30 installed ZIMs:

| | before | after |
|---|---|---|
| updates found | 11 | **13** |
| flavoured ids resolved | 0 | 2 |

The two that appeared are `wikibooks_en_all_nopic` (2026-01 → 2026-04) and `wikiversity_en_all_maxi` (2026-02 → 2026-05). Nothing dropped out.

Silent failure mode worth noting: when this lookup fails it does not error, it returns "no newer version", which is indistinguishable from "you are up to date".

https://claude.ai/code/session_01F8Mgg1vd2eKSs8StZiuyMV

### DIFF
--- a/admin/app/services/kiwix_catalog_service.ts
+++ b/admin/app/services/kiwix_catalog_service.ts
@@ -113,14 +113,43 @@ export class KiwixCatalogService {
   async getLatestZim(resourceId: string): Promise<CatalogResult | null> {
     const pattern = new RegExp(`^${escapeRegex(resourceId)}_(\\d{4}-\\d{2})\\.zim$`)
 
-    // 1. Exact-name lookup (the robust path).
+    // 1. Exact-name lookup (the robust path). Resolves most books, whose OPDS `<name>`
+    //    is the full filename base: devdocs_en_python, zimgit-water_en,
+    //    freecodecamp_en_all, nhs.uk_en_medicines ...
     const named = await this.fetchZimEntries({ name: resourceId, count: 50, start: 0 })
     const exact = this.pickNewestZim(named, pattern)
     if (exact) return exact
 
-    // 2. Fallback: bounded keyword scan in case the catalog ignored `name=` or
-    //    indexes the book under a slightly different name.
+    // 2. Flavoured books index under the base WITHOUT the flavour, which the catalog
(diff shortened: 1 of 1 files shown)


chriscrosstalk:
Step 3 (dropping `scanZimByQuery`) is now pushed, per Chris's call.

Re-verified on the same test box with the fallback removed: **13 updates, identical set, identical versions, nothing lost**. The check now completes in ~3s.

Net for the file: 8 insertions, 38 deletions.

jakeaturner (approved):

Issue #1320, opened 2026-09-05
install: preflight host mounts, management ports, and existing state before creating NOMAD resources


rcronk:
### Issue Category

Installation/Setup

### Bug Description

`install/install_nomad.sh` can discover predictable host incompatibilities only
after it has created `/opt/project-nomad`, downloaded and modified configuration,
generated credentials, pulled images, and attempted to start the management
stack. Two examples are:

1. The disk collector requires the host path used by its `rslave` bind mount to
       have compatible mount propagation. When it does not, Docker rejects container
       creation with `path / is mounted on / but it is not a shared or slave mount`.
2. A process already listening on a published management port, such as Dozzle's
       port 9999, causes `docker compose up` to leave a partially started stack.

Recovery is risky because the installer does not classify an existing complete
or partial installation. Continuing a rerun overwrites generated configuration,
regenerates credentials, and removes the existing MySQL data directory. The
initial confirmation warns broadly about overwrites, but does not identify the
state found or direct the user to update, uninstall, or an explicit clean
reinstall path.

I reproduced both failures under Ubuntu 26.04 with native Docker Engine. The
environment happened to be WSL2, but the requested checks are based on Linux
mount propagation, bound host ports, and filesystem/container state rather than
WSL detection. This proposal does not request official WSL support or removal of
the disk collector's `rslave` requirement.

#### Mount propagation

1. On a Debian/Ubuntu Docker host, make the host path used by the disk collector
       a private mount.
2. Run the installer and accept its prompts.
3. Observe that installation proceeds until Compose tries to create the disk
       collector, then fails with the shared/slave mount error after NOMAD resources
       have already been created.

#### Management port conflict

1. Bind host port 9999 with an unrelated process or container.
2. Run the installer and accept its prompts.
3. Observe that some management containers start before Dozzle fails with an
       address-in-use error, leaving a partial stack.

#### Rerun risk

1. Start from either partial state, or an existing installation with MySQL data.
2. Run the installer again and continue past the overwrite warning.
3. Observe that configuration and credentials are regenerated and the existing
       MySQL directory is removed.

### Expected Behavior

Before creating NOMAD directories, credentials, networks, volumes, or
containers, the installer should:

- verify that each bind mount requiring `rslave` has compatible propagation and
      stop with the failing path, observed propagation, and an actionable message;
- verify every host port published by the management Compose stack, currently
      8080 and 9999, and identify each conflict before starting any container; and
[...]

PR #1333, opened 2026-09-11
fix(installer): check mount propagation and ports before installing

rcronk:
### Summary

`install_nomad.sh` created `/opt/project-nomad`, generated credentials, and
started pulling images before it knew whether the host could run the management
stack. Two cheap host conditions are only discovered afterwards, as an opaque
`docker compose up` failure:

- Root mount propagation that is neither `shared` nor `slave`, which the
  management stack requires for its `/:/host:ro,rshared` mount.
- An already occupied published host port.

By then the installer has written to the host and the user has answered every
prompt.

This adds a side-effect-free preflight that runs after the existing environment
checks and immediately before `create_nomad_directory`, so a host that cannot
work is reported before anything is created.

### What changed

- `run_host_preflight` inspects mount propagation with `findmnt` and published
  ports with `ss`, aggregates every failure rather than stopping at the first,
  and prints the specific mount or port with its remediation.
- Both the propagation-sensitive mounts and the published ports are derived from
  `install/management_compose.yaml` rather than hardcoded, so ports beyond 8080
  and 9999 are covered automatically as the Compose file changes. Port ranges are
  expanded to their individual host ports, a `/udp` publish is checked as UDP
  rather than TCP, and entries that are not host paths, such as named volumes,
  are skipped rather than reported as uninspectable.
- The Compose file downloaded for the preflight is reused for the installation
  instead of being fetched a second time.
- The top-level call sequence moved into a guarded `main` so the script can be
  sourced by tests without installing anything. The call order is otherwise
  unchanged.

Existing-install classification and its update/reinstall/data-deletion UX are
left out per the split agreed in the issue, and remain for the follow-up.

### Known gap this PR does not close

Worth recording, since it sharpens the follow-up's requirements. The port check
changes rerun behavior, but only by coincidence and only in one direction:

- With the stack **running**, `nomad_admin` and `nomad_dozzle` hold 8080 and
  9999 themselves, so the preflight now aborts a rerun before
  `download_management_compose_file` regenerates credentials and runs
  `rm -rf "${NOMAD_DIR}/mysql"`. That is a real improvement, but an incidental
  one; it falls out of the port check rather than being designed.
- With the stack **stopped** — the likelier state after a failed install — the
  ports are free and propagation is fine, so the preflight passes and the rerun
  reaches the same data deletion it does today.

Because the preflight is the only gate ahead of that path, the second case is
unchanged by this PR. It is squarely the existing-install classification work,
so I have not tried to address it here.

Happy to open a follow-up issue for it if that is useful, or leave it to you if
you would rather scope that design yourselves.

### Tests

`install/tests/install_nomad_preflight_test.sh` is a new dependency-free harness
with 15 assertions covering:

- deriving every published management port and every propagation-sensitive mount
  from the Compose file
- `shared` and `slave` propagation passing
- `private` propagation failing with its remediation
- `findmnt` being unavailable failing clearly
- `ss` being unavailable failing clearly
- each published port being reported by number when occupied
- an occupied port warning that rerunning is not an update path
- port ranges expanding and the published protocol being preserved
- named volumes being ignored when deriving propagation mounts
- the preflight running before installation directory creation
- the installation reusing the preflighted Compose file
- `main` failing preflight without creating the installation directory

Run it with:

```bash
bash install/tests/install_nomad_preflight_test.sh
```

### Validation

- `bash -n` passes on both touched scripts.
- The 15-assertion harness passes.
- ShellCheck 0.11.0 reports exactly the same pre-existing warning counts on
  `install_nomad.sh` as `dev` (`SC2155` x3, `SC2034` x3, `SC2162` x2, `SC2024`
  x1), so the change introduces no new ShellCheck warning. The harness is clean
  under `shellcheck -x`.
- Validated end to end on a disposable Ubuntu 26.04 VM (kernel
  `7.0.0-30-generic`, Docker Engine `29.8.0`, Compose `v5.5.1`, no GPU), running
  the actual script with its real prompts:
  - **Root made `private`:** the installer ran its environment checks and
    prompts, then exited 1 at the preflight. Docker images, containers, volumes,
    and networks were unchanged, `/opt/project-nomad` was never created, and the
    temporary Compose file was cleaned up.
  - **Shared root, free ports:** exit 0. All six management containers started,
    Redis, MySQL, and Admin reported healthy, 8080 returned 302 and 9999
    returned 200, and every `replaceme` value in `compose.yml` was injected.
  - **Both published ports held by listeners:** both reported as `/tcp` with
    the port number, exit 1, before any resource was created.
  - **Real UDP socket:** detected as UDP and not as TCP.
- `/opt/project-nomad/compose.yml` keeps the umask-derived mode `curl -o`
  produced before; the copy is done by redirection specifically so the temp
  file's `0600` is not propagated.

The original report also involved WSL2, but nothing here is WSL-specific and no
WSL-specific branch was added; the checks are plain Linux mount and socket
inspection, developed and tested against supported Ubuntu.

Refs #1320

### DIFF
--- a/install/install_nomad.sh
+++ b/install/install_nomad.sh
@@ -37,6 +37,7 @@ UPDATE_SCRIPT_URL="https://raw.githubusercontent.com/Crosstalk-Solutions/project
 script_option_debug='true'
 accepted_terms='false'
 local_ip_address=''
+preflight_compose_file=''
 
 ###################################################################################################################################################################################################
 #                                                                                                                                                                                                 #
@@ -425,9 +426,17 @@ create_nomad_directory(){
 
 download_management_compose_file() {
   local compose_file_path="${NOMAD_DIR}/compose.yml"
+  local source_file="${1:-}"
 
   echo -e "${YELLOW}#${RESET} Downloading docker-compose file for management...\\n"
-  if ! curl -fsSL "$MANAGEMENT_COMPOSE_FILE_URL" -o "$compose_file_path"; then
+  if [[ -n "$source_file" ]]; then
+    # Redirection rather than cp so the file keeps the umask-derived mode that
+    # the curl path below produces; cp would copy the temp file's 0600 instead.
+    if ! cat "$source_file" > "$compose_file_path"; then
+      echo -e "${RED}#${RESET} Failed to copy the preflighted docker compose file."
+      exit 1
+    fi
+  elif ! curl -fsSL "$MANAGEMENT_COMPOSE_FILE_URL" -o "$compose_file_path"; then
     echo -e "${RED}#${RESET} Failed to download the docker compose file. Please check the URL and try again."
     exit 1
   fi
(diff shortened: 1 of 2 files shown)


chriscrosstalk (approved):
This landed exactly where we scoped it on #1320: side-effect-free checks only, everything derived from the compose file instead of hardcoded, and the occupied-port message calling out that a rerun is not an update path is the most important line in the PR. The test coverage without needing root or Docker is a nice touch.

Two non-blocking notes: a relative or ~ volume path would get inspected against the wrong base by findmnt (moot today since the management compose only mounts /), and the test script isn't wired into CI yet, so it only protects when run by hand. Neither needs to hold this up. Approving.

Issue #1323, opened 2026-09-07
feat(collections): make curated catalog sources configurable so regional content can ship without changing defaults


kumalabsconsulting:
## Feature Category

Content Management (ZIM, Maps, Collections)

## Problem Statement

Curated catalog manifests are fetched from hardcoded URLs in
`admin/app/services/collection_manifest_service.ts`:

```ts
const SPEC_URLS: Record<ManifestType, string> = {
  zim_categories: 'https://raw.githubusercontent.com/Crosstalk-Solutions/project-nomad/refs/heads/main/collections/kiwix-categories.json',
  maps: '.../collections/maps.json',
  wikipedia: '.../collections/wikipedia.json',
  creator_packs: '.../collections/creator-packs.json',
}
```

There is no environment variable, database setting or admin UI to point an
install at an additional catalog source.

This is the structural cause of a problem already diagnosed by a maintainer in
#1262. Because manifests are read live from `main`, adding regional entries to
the shared catalogs would surface them for every existing install immediately:

> "Curated collection manifests are fetched live from `main`, so a merge reaches
> every existing install immediately, while app changes only reach people who
> update. There is no language filter in the content picker yet. So French
> Wikipedia and Kiwix entries added today would appear for every English user on
> the current build."

The consequence today is that regional content cannot be distributed at all —
not through the shared catalogs (it would change defaults for everyone), and not
independently (the URLs cannot be changed). A non-English user must side-load
ZIM files manually and re-run a library rescan, outside any catalog.

I have read #486, #684, #738, #1221, #1262 and the framework PRs #648 and #1267
before writing this. This issue is deliberately **not** about i18n, UI
translation, or a translation framework — those are in-house and not scheduled,
and this proposal does not touch them.

## Proposed Solution

Allow catalog sources to be **extended**, not replaced. Defaults stay exactly as
they are.

- keep the current `SPEC_URLS` as built-in defaults, unchanged;
- allow one or more additional catalog source URLs (environment variable, or a
  row in the existing settings/`custom_library_sources` pattern already used for
  Kiwix mirrors);
- fetch, validate and cache additional sources through the existing
  `fetchAndCacheSpec()` path, with the same Vine schema validation and the same
  `getSpecWithFallback()` offline behaviour;
- entries from additional sources are merged into the picker; an install with no
  additional source configured behaves identically to today.

Why this shape:

- **it cannot change English defaults.** Nothing appears for anyone who has not
[...]

Issue #1262, opened 2026-08-21
Proposal: French localization and curated offline content


inkly:
### Context

I would like to contribute a complete French experience for Project NOMAD: Command Center UI, locale-aware formatting, documentation, and curated French offline content.

I reviewed the previous internationalization work and maintainer decisions in #486, #684, #738 and PRs #490, #518, and #648. I understand that the i18n framework, key naming, locale loading, content-language wiring, and translation maintenance are intended to be designed in-house. I do not want to submit a fourth competing framework or create rework for maintainers.

### Proposed contribution

Once the project's official i18n direction is available, I can contribute:

- complete native French translations with English fallback;
- French date, time, number, and unit formatting;
- translated accessibility labels, validation messages, notifications, and error states;
- a French Wikipedia option and availability-checked French Kiwix resources;
- French user documentation, kept separate if that makes review and maintenance easier;
- tests for locale selection, persistence, fallback behavior, and manifest validity.

The visual identity and existing user flows would remain unchanged.

### Direction requested before implementation

Is there now an in-house branch, translation schema, or planned milestone that a French contribution can target?

If the UI framework is still not ready for community contributions, would maintainers accept a smaller content-only PR that adds French Wikipedia/Kiwix choices without changing the English defaults, or should all French work wait for the official internationalization foundation?

I am happy to split the work into small PRs in the order maintainers prefer.

inkly:
While awaiting maintainer direction, I prepared a content-only proof-of-concept branch that does not introduce an i18n framework or change any English defaults:

https://github.com/inkly/project-nomad/tree/ixelia/french-offline-content

- one new `fr` catalog category with three additive tiers;
- eight official Kiwix resources covering general reference, health, science, education, repair, and literature;
- cumulative sizes of approximately 522 MB, 4.3 GB, and 17.5 GB;
- all eight URLs verified with HTTP 200;
- the complete manifest validates against the current VineJS schema;
- resource IDs and versions match NOMAD's existing filename parser.

The branch is based on `upstream/dev`. I have not opened a pull request yet and am happy to adjust or hold this content contribution based on your guidance.

chriscrosstalk:
Thanks for this, and specifically for reading #486, #684, #738 and the three PRs before writing anything. That's rare and it made this easy to answer properly.

[...]

Issue #884, opened 2026-05-14
bug(zim): onWikipediaDownloadComplete deletes ALL co-existing wikipedia_en_* ZIMs, not just stale versions of the same selection


chriscrosstalk:
# `onWikipediaDownloadComplete` deletes ALL co-existing `wikipedia_en_*` ZIMs, not just stale versions of the same selection

## Summary

`ZimService.onWikipediaDownloadComplete()` (called from `downloadRemoteSuccessCallback` on every download whose URL contains `wikipedia_en_`) deletes every other file in `/storage/zim/` whose name starts with `wikipedia_en_`. This treats *any* Wikipedia-prefixed ZIM as a competing version, but `wikipedia_en_simple_*`, `wikipedia_en_medicine_*`, `wikipedia_en_wikivoyage_*`, etc. are **distinct corpora**, not different versions of the same selection.

Consequence: downloading two Wikipedia-prefixed ZIMs in sequence silently deletes whichever finished first.

## Where

`admin/app/services/zim_service.ts` — `onWikipediaDownloadComplete()`:

```ts
if (success) {
  // ...update WikipediaSelection record...

  // Delete old Wikipedia files (keep only the newly installed one)
  const existingFiles = await this.list()
  const wikipediaFiles = existingFiles.files.filter((f) =>
    f.name.startsWith('wikipedia_en_') && f.name !== filename
  )

  for (const oldFile of wikipediaFiles) {
    try {
      await this.delete(oldFile.name)
      logger.info(`[ZimService] Deleted old Wikipedia file: ${oldFile.name}`)
    } catch (error) {
      logger.warn(`[ZimService] Could not delete old Wikipedia file: ${oldFile.name}`, error)
    }
  }
}
```

The trigger is in `downloadRemoteSuccessCallback`:

```ts
for (const url of urls) {
  if (url.includes('wikipedia_en_')) {
    await this.onWikipediaDownloadComplete(url, true)
  }
}
```

## Repro

On a fresh NOMAD with no Wikipedia installed:

1. `POST /api/zim/download-remote` for `wikipedia_en_simple_all_nopic_2026-02.zim` — completes, file is in `storage/zim/`.
2. `POST /api/zim/download-remote` for `wikipedia_en_medicine_nopic_2026-04.zim` — completes, file is in `storage/zim/`.
3. Observe `ls storage/zim/wikipedia_en_*.zim` — **only the medicine file remains**. simple has been silently deleted.

[...]

PR #1334, opened 2026-09-11
feat(content): content-language filter for curated collections, Wikipedia and Kiwix search

IsT3RiK:
Hi Chris, hi Jake 👋

This PR follows up directly on your answer in #1262:

> The content path isn't blocked on the framework, it's blocked on a language filter in the picker, and that's a smaller piece of work than full i18n.

It implements that language filter and nothing more. **This is not i18n:** no translation framework, no string extraction, and the UI stays in English. It only changes *which content is offered*. It also respects the constraint you raised on #1267 (manifest changes and app changes must not ship together): this PR is app code only, with no manifest changes.

I know CONTRIBUTING asks for an issue before a non-trivial PR. I treated #1262 and your note there as the design discussion, and it also overlaps #1323 (configurable catalog sources). If you'd rather have a fresh issue, I'll gladly open one and park this. And if it doesn't fit your plans, closing it is completely fine.

## What it does

A new **Content Languages** selector appears in **Easy Setup → Choose Content** and in **Settings → Content Explorer**. It saves a `content.languages` setting (ISO 639-1 codes, **default `en`**). The selected languages drive:

- **Curated collections**: English still comes from `collections/kiwix-categories.json`. Each other enabled language reads `collections/<code>/kiwix-categories.json`. Categories are merged and stamped with their language, and non-English slugs are namespaced (`fr:medicine`) so they can't collide with English ones.
- **Wikipedia picker**: same scheme for `wikipedia.json`. Option ids are namespaced (`fr:all-maxi`) and only English's "none" option is kept. The installed selection stays visible even if its language is deselected.
- **Kiwix library search**: `lang` becomes the selected languages (`eng,fra`) instead of the hardcoded `eng`.

Non-English categories and Wikipedia options get a small language badge. English-only users see no change.

## Why existing installs are safe

- With the default (`en` only), the manifest URLs, `collection_manifests` cache keys, category slugs and Wikipedia option ids are **exactly** what they are today. Tests pin this.
- Non-English manifests live in `collections/<code>/`, a path no released build requests. Publishing French content later can't reach an existing install; the safety comes from the file layout, not from a filter that old builds lack.
- There's **no DB migration**: non-English manifests are cached as `<type>:<code>` rows in the existing string-keyed table.
- The Wikipedia option schema gains an optional `language` field. VineJS strips unknown keys, so old builds simply ignore it.

## A related Wikipedia fix

The picker's Wikipedia used to be recognised by the `wikipedia_en_` prefix, and just extending that prefix to other languages would have spread an existing bug. Medicine's Comprehensive tier ships `wikipedia_en_medicine_maxi`, and when it finishes downloading it's treated as *the* Wikipedia:

- the selection's `filename` is overwritten with WikiMed;
- its `InstalledResource` row is skipped;
- on the next restart, `reconcileFromFilesystem` skips it as "the managed Wikipedia" and removes its row, so the tier silently shows as not installed.

I reproduced this end to end with a French WikiMed tier.

A new pure helper, `isSelectorManagedWikipedia()`, now treats a download as the picker's only when its stem matches the current selection (so auto-updates to a newer release still count) or a known option. None of the curated `wikipedia_*` resources share a stem with an option today (checked against `collections/`). Ordinary ZIMs are answered without any I/O, so an offline install never waits on GitHub to finish a download. The upload path now uses `findReplacedWikipediaFiles` like the download flow (#884), instead of deleting every other `wikipedia_en_*` file. Happy to split this into its own PR if you prefer.

## Security review

- `content.languages` is validated server-side against a fixed allow-list (23 languages). Empty, unknown and malformed values are rejected with a clear message.
- Only allow-listed codes can reach a manifest URL path or the Kiwix `lang` parameter. Values stored anyway (e.g. edited in the DB) are dropped when read, and `fetchAndCacheSpec` refuses unknown codes as defence in depth. Tests cover path traversal (`../`), encoded slashes, query/fragment injection, markup and NUL bytes.
- A non-English manifest can't shadow an English category or the "none" option (namespacing plus dedupe, tested).
- Non-English manifests go through the same VineJS schemas as English ones. There are no new routes and no dependency changes, and the UI renders language names as plain text.
- The Wikipedia fix *narrows* file deletion (only older releases of the same variant). Deletion still goes through `ZimService.delete()`, with its existing path-traversal guard.

## Testing

**Automated, run by Claude:**
- **Unit tests**: `content_languages.spec.ts` (24 new tests, 4 of them security) and `zim_filename.spec.ts` (+9).
- **Non-regression**: `npm run test:unit` gives 526/535. The 9 failing files (`app_auto_update`, `content_auto_update`, `content_auto_update_backoff`, `custom_app_guard`, `drug_ingest_status`, `drug_interactions`, `drug_labels`, `ollama_controller_chat`, `rag_retrieval_toggle`) fail the same way on `dev` without this change, in a plain `node:24` container. They look like they need services I didn't run.
- **Types and build**: `npm run typecheck` is clean and `npm run build` passes (what `build-admin-on-pr` runs). The inertia `tsc` error count is unchanged (35 before and after).
- **Lint**: no new ESLint/Prettier findings on the changed lines. Some touched files weren't Prettier-formatted before, and I matched their existing style rather than reformatting them.
- **End to end**, on a throwaway instance built from this branch with its own MySQL/Redis:
  - English-only is byte-identical;
  - EN+FR and FR-only filter collections, Wikipedia options and Kiwix search as expected;
  - invalid values are rejected;
  - a French Wikipedia package can be selected, cancelled (→ `failed`) and fully downloaded;
  - a French curated tier installs without touching the Wikipedia selection;
  - a Wikipedia-derived corpus (`wikipedia_en_knots_maxi`) and an ordinary ZIM are bookkept correctly;
  - everything survives a restart.

  French curated content was simulated with demo manifests injected into the cache, since `collections/fr/` doesn't exist yet.
- **UI** was checked in a real headless Chromium through Claude Code's browser skill: desktop, 390 px mobile and dark theme, with no console errors.

**Manual:** I tested it myself on that instance. I switched languages in Content Explorer and Easy Setup, downloaded French content (French Wikipedia, WikiMed FR, iFixit FR), and browsed it in Kiwix. Everything worked as expected.

## Noticed while testing (not addressed here)

- **Kiwix search on `dev`**: search queries `browse.library.kiwix.org`, which currently answers with an HTML "Access confirmation" page, so search is broken on `dev` regardless of this PR. `main` uses `opds.library.kiwix.org` (see the revert in 93be567). I patched it locally only to test.
- **Update checks**: `KiwixCatalogService` queries with `lang: 'eng'`, so sideloaded non-English ZIMs never resolve an update. `name=wikipedia_fr_all` returns 0 results with `lang=eng` and 3 without. I'm happy to open an issue or a one-line PR if useful.

## Next step (separate PR, once this ships)

A content-only PR adding `collections/fr/kiwix-categories.json` and `collections/fr/wikipedia.json` (Wikipédia, WikiMed FR, iFixit FR, Wiktionnaire, Wikilivres, Vikidia, Gutenberg FR, Les fondamentaux). I'd be glad to maintain those files. The same layout works for any language.

<details>
<summary>Suggested release note (optional)</summary>

> **Content languages**: Easy Setup and the Content Explorer now let you choose which languages of curated collections, Wikipedia packages and Kiwix library results are offered. English remains the default, and the interface itself is unchanged.
</details>

## Transparency

This change was written by Claude (Anthropic's Claude Opus 5, through Claude Code), working under my direction. It ran the automated, security and regression checks above. I reviewed the result and tested it by hand before opening this PR. Thanks a lot for NOMAD, and for taking the time to explain the constraints so clearly in #1262. It made this much easier to get right. Any feedback is very welcome.

### DIFF
--- a/admin/app/controllers/easy_setup_controller.ts
+++ b/admin/app/controllers/easy_setup_controller.ts
@@ -35,18 +35,19 @@ export default class EasySetupController {
 
   async refreshManifests({}: HttpContext) {
     const manifestService = new CollectionManifestService()
-    const [zimChanged, mapsChanged, wikiChanged] = await Promise.all([
+    const [zimChanged, mapsChanged, wikiChanged, otherLanguages] = await Promise.all([
       manifestService.fetchAndCacheSpec('zim_categories'),
       manifestService.fetchAndCacheSpec('maps'),
       manifestService.fetchAndCacheSpec('wikipedia'),
+      manifestService.refreshAdditionalLanguageManifests(),
     ])
 
     return {
       success: true,
       changed: {
-        zim_categories: zimChanged,
+        zim_categories: zimChanged || otherLanguages.zim_categories,
         maps: mapsChanged,
-        wikipedia: wikiChanged,
+        wikipedia: wikiChanged || otherLanguages.wikipedia,
       },
     }
   }
--- a/admin/app/models/collection_manifest.ts
+++ b/admin/app/models/collection_manifest.ts
@@ -1,12 +1,15 @@
 import { DateTime } from 'luxon'
 import { BaseModel, column, SnakeCaseNamingStrategy } from '@adonisjs/lucid/orm'
-import type { ManifestType } from '../../types/collections.js'
-
 export default class CollectionManifest extends BaseModel {
   static namingStrategy = new SnakeCaseNamingStrategy()
 
+  /**
+   * Cache key: the bare ManifestType for English ('zim_categories'), or
+   * `<type>:<language>` for another content language ('zim_categories:fr').
+   * See manifestCacheKey in utils/content_languages.
+   */
   @column({ isPrimary: true })
-  declare type: ManifestType
+  declare type: string
 
   @column()
   declare spec_version: string
--- a/admin/app/services/collection_manifest_service.ts
+++ b/admin/app/services/collection_manifest_service.ts
@@ -10,6 +10,16 @@ import { QueueService } from './queue_service.js'
 import { RunDownloadJob } from '#jobs/run_download_job'
 import { zimCategoriesSpecSchema, mapsSpecSchema, wikipediaSpecSchema, creatorPacksSpecSchema } from '#validators/curated_collections'
 import { isGatedResource } from '../utils/hosted_content.js'
+import KVStore from '#models/kv_store'
+import {
+  isSupportedContentLanguage,
+  manifestCacheKey,
+  manifestUrlFor,
+  mergeZimCategorySpecs,
+  parseContentLanguages,
+  type LanguageScopedManifestType,
+} from '../utils/content_languages.js'
+import { DEFAULT_CONTENT_LANGUAGE } from '../../constants/content_languages.js'
 import {
   ensureDirectoryExists,
   listDirectoryContents,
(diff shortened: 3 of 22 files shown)


IsT3RiK:
prmobile
prdark
prdesktop

chriscrosstalk:
Thanks for this, and for the care in the framing — you scoped it exactly to what I described in #1262, respected the manifest/app separation from #1267, and were upfront about the process gap. That's a model contribution even though I'm closing it, and I want to be straight about why.

The blocker isn't the code, it's the commitment on the other side of it: per-language manifests at `collections/<code>/` would mean us authoring and maintaining curated collections in languages we don't speak, indefinitely. We're not willing to sign up for that, so the official-catalog language route is off the table as designed.

What we are doing instead: building a community content distribution mechanism into a future version internally (details still in design). The client-side work you did here — the language setting, category merging, namespaced slugs — maps closely onto what that will need, so this wasn't wasted effort; it was early. Keep an eye out for when that lands.

PR #1347, opened 2026-09-14
fix(install): time out Redis readiness probes

### LINKED ISSUE #1329: fix: bound the Redis entrypoint readiness probe with a socket timeout
### Issue Category

Installation/Setup

### Bug Description

The Redis readiness loop in `install/entrypoint.sh` is intended to try once per
second for up to 60 attempts, but each Node `net.connect()` call has no socket
timeout. If packets are silently dropped rather than rejected, the first probe
can remain blocked for the kernel TCP timeout and the loop's 60-attempt bound is
never reached.

I observed this after starting an existing NOMAD stack while container bridge
traffic was temporarily unavailable. `nomad_admin` remained in `health:
starting`, port 8080 reset connections, and its Redis/BullMQ clients logged
repeated `connect ETIMEDOUT` errors. Redis and MySQL later became healthy and
were directly reachable from the Admin container, but Admin required a restart
to recover.

This is not WSL-specific: any supported Docker host that silently drops the
initial Redis SYN can trigger the same behavior.

### Expected Behavior

Each readiness attempt should have a short explicit timeout, close its socket,
and return failure so the shell loop can retry. If Redis never becomes reachable,
the existing 60-attempt limit should terminate the entrypoint with its actionable
error instead of allowing one connection attempt to block indefinitely.

### Proposed Implementation

Add a timeout to the Node socket probe in `install/entrypoint.sh`, for example by
calling `socket.setTimeout(...)`, destroying the socket on timeout, and using a
single guarded completion function so `connect`, `error`, and `timeout` cannot
race multiple exits.

### Test Approach

- Reachable Redis endpoint exits successfully.
- Refused connection fails promptly and advances the loop.
- Blackholed/dropped connection fails within the per-attempt timeout.
- Socket resources are closed on success, error, and timeout.
- The outer 60-attempt failure path remains reachable.

### Related Work

Commit `f4af81e` added the readiness loop and lazy ioredis connection handling to
avoid blocked startup. This follow-up preserves that design while making its
retry bound effective for silent packet drops.

dajiaohuang:
Closes #1329.

The Redis readiness probe now applies a one-second socket timeout and uses a guarded completion callback for connect, error, and timeout events. Timed-out sockets are destroyed so the existing 60-attempt retry bound remains effective.

Validation:
- Node probe reached a local listener successfully.
- Refused connection exited promptly.
- A blackholed test address exited after the one-second timeout.
- git diff --check passed.

POSIX shell syntax validation was not run because sh is unavailable on this Windows host; Docker validation was unavailable because the Docker daemon is not running.


chriscrosstalk (approved):
Reviewed the probe change and it's exactly right: the guarded completion callback covers the connect/error/timeout races and destroying the socket keeps the 60-attempt bound meaningful. Thanks for the standalone validation on the blackholed case, that's the scenario that mattered. LGTM for merge.

jakeaturner (approved):

Issue #1343, opened 2026-09-14
Link tile with a URL over 255 characters returns a 500 with raw SQL in the log


chriscrosstalk:
## Problem

The link tile URL validator allows 2048 characters but `services.custom_url` is `varchar(255)`, so any URL between 256 and 2048 characters fails at the database. The user gets a generic error page, and the server log records the raw MySQL error along with the full INSERT statement.

The URL input in the add-link-tile modal has no `maxLength`, so nothing catches it client-side either.

This is reachable by pasting an ordinary long URL — a Grafana dashboard link, a Nextcloud share link, any signed URL.

## Steps to reproduce

1. Dashboard → add a link tile.
2. Paste a URL longer than 255 characters.
3. Submit.

## Expected vs actual

**Expected:** 422 with a readable message, e.g. "URL must be 255 characters or fewer."

**Actual:** HTTP 500 and an Inertia error page. The log records:

```
ER_DATA_TOO_LONG (errno 1406): Data too long for column 'custom_url' at row 1
```

followed by the complete INSERT statement.

Boundary probe, identical on two servers:

| URL length | Result |
|---|---|
| 200 chars | 201 Created |
| 260 chars | **500** |
| 325 chars | **500** |
| 600 chars | **500** |

## Root cause

Validator/column mismatch in `admin/app/validators/system.ts`:

```ts
export const createLinkTileValidator = vine.compile(vine.object({
    friendly_name: vine.string().trim().minLength(1).maxLength(60),
    url: vine.string().trim().minLength(1).maxLength(2048),   // column is varchar(255)
    ...
}));
```

The same mismatch is present in `updateLinkTileValidator`.

## Suggested fix

Align the validator with the column (`maxLength(255)`) on both create and update, and add `maxLength` to the modal's URL input. Alternatively widen the column, though 255 seems reasonable for a dashboard tile.

Worth noting this release fixed the same class of bug one table over — map marker notes are now capped server-side at 500 characters and return a clean 422.

## Files involved

- `admin/app/validators/system.ts` — `createLinkTileValidator`, `updateLinkTileValidator`
- the add/edit link tile modal — URL input needs `maxLength`
- `services.custom_url` column definition

chriscrosstalk:
Confirmed in the browser on v1.35.0-rc.1, plus two details that change how this reads.

**1. The error message tells the user to do the one thing that cannot work.**

Pasting a 279-character URL and submitting produces this toast:

[...]

PR #1354, opened 2026-09-21
fix: reject oversized custom URLs

dajiaohuang:
## Summary

- Align custom URL validation with the services.custom_url VARCHAR(255) column.
- Reject over-limit normalized URLs before persistence and cap the UI input.
- Add unit coverage for exact-limit and over-limit normalized URLs.

Related to #1343, which documents the same validator/column mismatch in the link-tile flow. The current implementation uses the custom_url endpoint; this patch applies the same boundary there.

## Validation

- git diff --check
- Focused unit test, typecheck, and lint were attempted but the registered workspace has no admin/node_modules, so the commands cannot run until dependencies are installed in an authorized environment.


chriscrosstalk (approved):
Nice catch that normalizeCustomUrl is the chokepoint here. Fixing it in the shared function covers the link-tile create and update paths as well as the custom URL endpoint, which is a better shape than patching all four call sites, and the boundary tests with the exact scheme-prepend math are appreciated. Three small asks, none blocking: change "Related to #1343" in the description to "Fixes #1343" so the issue closes on merge; add the same maxLength={255} to LinkTileModal's URL input that you gave AppUrlModal; and since a valid-but-long URL now fails normalization, the 422 message "Enter a valid URL" is misleading for that case. Worth a distinct "URL must be 255 characters or fewer" message, and the link-tile validators' maxLength(2048) could come down to match while you are in there.

jakeaturner (approved):
Great fix, thank you!

Issue #1121, opened 2026-07-19
bug(maps): many regional PMTiles (~50) yields coast-only view; regional streets never paint


CAMOBAP:
### Issue Category

UI/Frontend Issue / Maps

### Bug Description

With many regional `.pmtiles` files installed at once (e.g. all 50 US states from project-nomad-maps **plus** a country extract like Poland), the offline map viewer shows only the low-zoom `world.pmtiles` overzoom — coastline / land silhouette — when zoomed into a city that **does** have a detailed regional archive on disk. Streets, buildings, and POIs are missing even though the regional file contains them.

Reducing active maps to `world.pmtiles` + one region (`pl_*_z15.pmtiles`) restores full city detail immediately.

This is distinct from:
- #997 (duplicate source/layer ids blanking the **entire** map)
- #838 (two neighboring countries blanking the **border** zone)

Here the map still renders, but only the z≤5 world basemap is effectively visible over the detailed region.

### Steps to Reproduce

1. Install a large set of regional maps (e.g. full US states collection) under `/opt/project-nomad/storage/maps/pmtiles/`.
2. Extract/install a non-US country at maxzoom 15 (e.g. Poland via Country Picker) so `pl_*_z15.pmtiles` is present (~2.8 GB).
3. Confirm `/api/maps/styles` lists ~52 sources and ~3600+ layers (70 style layers × N sources).
4. Open `/maps`, zoom to a city inside the country extract (e.g. Gdańsk ~54.35, 18.65 at z14).
5. Observe coast/land silhouette only — no streets/buildings/POIs.
6. Move the US state files **out** of the maps tree (so recursive listing no longer picks them up), restart admin, hard-refresh `/maps`.
7. Same Gdańsk view now shows full streets, buildings, labels, POIs.

### Expected Behavior

Regional extracts that cover the viewport should paint street/building/POI layers on top of the world basemap, regardless of how many *other* non-overlapping regions are installed.

Ideally the style should not multiply every basemap layer by every installed region (or should only attach sources whose bounds intersect the current viewport).

### Actual Behavior

- Style generation (`MapService.generateStylesFile`) clones the full layer template for **every** installed `.pmtiles` → with 50 US states + PL + world ≈ **52 sources / ~3641 layers**.
- In that configuration, zooming into Poland shows only world overzoom (coast outline).
[...]

PR #1355, opened 2026-09-21
fix(maps): add navigation for downloaded regions

sundruid:
## Problem

Downloaded regional PMTiles are difficult to discover from the default continental view. Users can successfully download large regional archives but still see only the low-detail world basemap, making the downloads appear empty.

## Changes

- Add a **Jump to downloaded region** selector to the map viewer.
- Read each selected PMTiles archive header in the browser and fit the map to its actual bounds.
- Support arbitrary future region and country downloads without hard-coded coordinates.
- Present friendly, sorted region names and loading/error states.
- Position the selector below the existing Pins control.
- Clarify in Maps Manager that the downloads are offline vector maps whose local detail appears after zooming into a region.

## Verification

- Production asset build and TypeScript compilation pass with `node ace build`.
- Verified the selector against the installed Arizona archive and confirmed bounds navigation.
- Verified Arizona and California archives with `pmtiles verify`.
- Existing map, marker, coordinate, scale, and saved-view behavior remains unchanged.


chriscrosstalk (approved):
This solves a confusion we've seen more than once, most recently in #1121 where regional streets "never painted" because the user was looking at the continental view. Reading the archive header for real bounds instead of hardcoding coordinates is the right call, and the semantic color tokens mean it themes correctly in dark mode.

Verified live on a v1.35.0-rc.1 test box with 14 installed region archives: the selector renders below the Pins control without collision, picking a state reads its header over a range request and fits the bounds exactly, and there are no console errors in either theme.

Two housekeeping asks before this can merge: please retarget the PR from main to dev (all changes land on dev first per CONTRIBUTING.md), and file a short issue describing the discoverability problem this fixes so we can link it. Heads up that the maps component has drifted between main and dev, so expect some conflicts to resolve on the rebase.

One small note for a possible follow-up: if an older duplicate archive for the same region is still on disk, the dropdown will list both entries even though the map only renders the newest one. Not a blocker.

chriscrosstalk:
Quick update before you sink time into that rebase. After approving I put this on a test box and actually used it for a while, and I ended up wanting a different home for the control: instead of a floating card on the map, fold it into the existing lat,lng box in the title bar as a single "Jump to: region or lat,lng" input backed by a datalist of the downloaded regions. Less chrome on the map for a control that gets used occasionally, and your core approach of reading the archive header for bounds carries over unchanged.

I prototyped that variant to make sure it holds up, and hit three things worth knowing no matter which UI wins:

1. Picking from a datalist fires onChange with no keypress, so the jump has to trigger on an exact label match, not on Enter. Wiring it to Enter alone means mouse selections silently do nothing.
2. Leftover text filters the datalist down to itself or to nothing, which reads as a broken dropdown. Clearing the box on focus and after a successful jump fixes it.
3. Alaska breaks fitBounds in both your version and my first attempt: the archive crosses the antimeridian, so the header bbox degenerates to [-180, 180] and the map zooms out to the whole world. I worked around it by detecting a longitude span over 350 degrees, probing which z3 tile columns actually have data inside the latitude band, and rebuilding wrap-aware bounds from the largest empty gap. No hardcoded coordinates, in the same spirit as your header read.

I realize this is the goalposts moving after an approval, so your call on how to proceed: if you want to rework the PR along these lines I will post my prototype diff here to start from (retargeted to dev), or if you would rather not re-cut it, I am happy to land the title-bar variant myself on top of your work with you credited as co-author. Either way, please do still file that issue for the discoverability problem so we can link whichever PR ships.

